PR #2, opened 2023-11-08
Initial owners file

bparees:


OWNERS
```diff
@@ -0,0 +1,6 @@
+reviewers:
```
bparees, line 1: note: i don't find prow's reviewer assignment to be useful so i am deliberately not populating this

vishnuchalla: late question: Does this change override all the OWNERS file that we have in PROW? 🤔

bparees: the owners file you put in prow is for who can approve merges to our CI job configs.

this owners file is for who can approve merges to this repo.

PR #5, opened 2023-11-13
Implementing kube-rbac-proxy authentication/authorization

vishnuchalla:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Implementing basic kube-rbac-proxy authentication.

- Related Issue # https://issues.redhat.com/browse/OLS-19?filter=-3

## Testing
Tested in my local environment with the below command
```
export OLS_IMAGE=quay.io/vchalla/lightspeed-service:latest
kustomize build . | envsubst | oc apply -f -
```
Verified valid responses from pod logs.
### ols container logs
```
INFO:     Started server process [1]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8080 (Press CTRL+C to quit)
INFO:     10.128.2.151:0 - "GET /docs HTTP/1.1" 200 OK
```

### ols client logs
```
*   Trying 172.30.61.91:8443...
* Connected to ols-proxy.ols.svc (172.30.61.91) port 8443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: none
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [122 bytes data]
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
{ [15 bytes data]
* TLSv1.3 (IN), TLS handshake, Request CERT (13):
{ [45 bytes data]
* TLSv1.3 (IN), TLS handshake, Certificate (11):
{ [1702 bytes data]
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
{ [264 bytes data]
* TLSv1.3 (IN), TLS handshake, Finished (20):
{ [36 bytes data]
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
} [1 bytes data]
* TLSv1.3 (OUT), TLS handshake, Certificate (11):
} [8 bytes data]
* TLSv1.3 (OUT), TLS handshake, Finished (20):
} [36 bytes data]
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=ols-proxy-77f77c9785-47gqd@1699908764
*  start date: Nov 13 19:52:44 2023 GMT
*  expire date: Nov 12 19:52:44 2024 GMT
*  issuer: CN=ols-proxy-77f77c9785-47gqd-ca@1699908764
*  SSL certificate verify result: self signed certificate in certificate chain (19), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
} [5 bytes data]
* Using Stream ID: 1 (easy handle 0x5635da475720)
} [5 bytes data]
> GET /docs HTTP/2
> Host: ols-proxy.ols.svc:8443
> user-agent: curl/7.76.1
> accept: */*
> authorization: Bearer <token>
> 
{ [5 bytes data]
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
{ [122 bytes data]
< HTTP/2 200 
< content-type: text/html; charset=utf-8
< date: Mon, 13 Nov 2023 20:52:59 GMT
< server: uvicorn
< content-length: 939
```

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -12,4 +12,4 @@ USER 1001
 RUN pip install -r requirements.txt 
 
 # Run the application
-CMD uvicorn ols:app --host 0.0.0.0
+CMD uvicorn ols:app --host 0.0.0.0 --port 8080
--- a/Containerfile
+++ b/Containerfile
@@ -12,4 +12,5 @@ USER 1001
 RUN pip install -r requirements.txt 
 
 # Run the application
+EXPOSE 8080
 CMD uvicorn ols:app --host 0.0.0.0 --port 8080
--- a/config/ols-client.yaml
+++ b/config/ols-client.yaml
@@ -1,0 +1,30 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: ols-client
+spec:
+  template:
+    metadata:
+      name: ols-client
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: ols-client
+        image: quay.io/brancz/krp-curl:v0.0.2
+        command:
+        - /bin/sh
+        - -c
+        - 'curl -v -s -k -H "Authorization: Bearer `cat /service-account/token`" https://ols-proxy.ols.svc:8443/docs'
+        volumeMounts:
+        - name: token-vol
+          mountPath: "/service-account"
+          readOnly: true
+      volumes:
+      - name: token-vol
+        projected:
+          sources:
+          - serviceAccountToken:
+              audience: ols-proxy.ols.svc
+              expirationSeconds: 3600
+              path: token
+  backoffLimit: 10
--- a/config/ols-rbac.yaml
+++ b/config/ols-rbac.yaml
@@ -1,0 +1,63 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ols-proxy
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: ols-proxy
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ols-proxy
+subjects:
+- kind: ServiceAccount
+  name: ols-proxy
+  namespace: ols
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ols-proxy
+rules:
+- apiGroups: ["authentication.k8s.io"]
+  resources:
+  - tokenreviews
+  verbs: ["create"]
+- apiGroups: ["authorization.k8s.io"]
+  resources:
+  - subjectaccessreviews
+  verbs: ["create"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ols-proxy-backend-register
+rules:
+- nonResourceURLs:
+    - "/docs"
+    - "/ols"
+    - "/base_llm_completion"
+    - "/feedback"
+  verbs:
+    - get
+    - create
+    - delete
+    - list
+    - update
+    - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: ols-proxy-backend-register
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ols-proxy-backend-register
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: ols
--- a/config/ols-rbac.yaml
+++ b/config/ols-rbac.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: ols-proxy
+  namespace: ols
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -30,34 +31,3 @@ rules:
   resources:
   - subjectaccessreviews
   verbs: ["create"]
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: ols-proxy-backend-register
-rules:
-- nonResourceURLs:
-    - "/docs"
-    - "/ols"
-    - "/base_llm_completion"
-    - "/feedback"
-  verbs:
-    - get
-    - create
-    - delete
-    - list
-    - update
-    - patch
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: ols-proxy-backend-register
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: ols-proxy-backend-register
-subjects:
-- kind: ServiceAccount
-  name: default
-  namespace: ols
--- a/config/ols.yaml
+++ b/config/ols.yaml
@@ -1,0 +1,49 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: ols-proxy
+  name: ols-proxy
+spec:
+  ports:
+  - name: https
+    port: 8443
+    targetPort: https
+  selector:
+    app: ols-proxy
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ols-proxy
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: ols-proxy
+  template:
+    metadata:
+      labels:
+        app: ols-proxy
+    spec:
+      securityContext:
+        runAsUser: 65532
+      serviceAccountName: ols-proxy
+      containers:
+      - name: ols-proxy
+        image: quay.io/brancz/kube-rbac-proxy:v0.15.0
+        args:
+        - "--secure-listen-address=0.0.0.0:8443"
+        - "--upstream=http://0.0.0.0:8080"
+        - "--ignore-paths=/healthz"
+        - "--auth-token-audiences=ols-proxy.ols.svc"
+        - "--logtostderr=true"
+        - "--v=10"
+        ports:
+        - containerPort: 8443
+          name: https
+        securityContext:
+          allowPrivilegeEscalation: false
+      - name: ols
+        image: quay.io/vchalla/lightspeed-service:latest
--- a/config/ols.yaml
+++ b/config/ols.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     app: ols-proxy
   name: ols-proxy
+  namespace: ols
 spec:
   ports:
   - name: https
@@ -17,6 +18,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: ols-proxy
+  namespace: ols
 spec:
   replicas: 1
   selector:
@@ -32,10 +34,10 @@ spec:
       serviceAccountName: ols-proxy
       containers:
       - name: ols-proxy
-        image: quay.io/brancz/kube-rbac-proxy:v0.15.0
+        image: quay.io/openshift/origin-kube-rbac-proxy:4.14.0
         args:
         - "--secure-listen-address=0.0.0.0:8443"
-        - "--upstream=http://0.0.0.0:8080"
+        - "--upstream=http://0.0.0.0:8080/"
         - "--ignore-paths=/healthz"
         - "--auth-token-audiences=ols-proxy.ols.svc"
         - "--logtostderr=true"
@@ -46,4 +48,4 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
       - name: ols
-        image: quay.io/vchalla/lightspeed-service:latest
+        image: ${OLS_IMAGE}
--- a/README.md
+++ b/README.md
@@ -45,19 +45,37 @@ _Below is an example of how you can instruct your audience on installing and set
 <!-- USAGE EXAMPLES -->
 ## Usage
 
-### Run the server
+### Local Deployment
+
+#### Run the server
 in order to run the API service  
    ```sh
         uvicorn ols:app --reload
    ```
 
-### Query the server
+#### Query the server
 
 To send a request to the server you can use the following curl command:
    ```sh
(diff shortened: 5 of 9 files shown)


config/ols.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - "--auth-token-audiences=ols-proxy.ols.svc"
+        - "--logtostderr=true"
+        - "--v=10"
+        ports:
+        - containerPort: 8443
+          name: https
+        securityContext:
+          allowPrivilegeEscalation: false
+      - name: ols
+        image: quay.io/vchalla/lightspeed-service:latest
```
vishnuchalla, line 49: We will need to replace this with our official image repository if any.

config/ols-rbac.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - "/ols"
+    - "/base_llm_completion"
+    - "/feedback"
+  verbs:
+    - get
+    - create
+    - delete
+    - list
+    - update
+    - patch
```
bparees, line 50: how do these verbs relate to the URLs?  What does it mean for a user to have "create" vs "get" permission on "/ols"?

config/ols.yaml
```diff
@@ -0,0 +1,49 @@
...
+    metadata:
+      labels:
+        app: ols-proxy
+    spec:
+      securityContext:
+        runAsUser: 65532
+      serviceAccountName: ols-proxy
+      containers:
+      - name: ols-proxy
+        image: quay.io/brancz/kube-rbac-proxy:v0.15.0
```
bparees, line 35: we should pull the openshift version of this image.  `quay.io/openshift/origin-kube-rbac-proxy:4.14.0` should work.

config/ols.yaml
bparees: since you appear to be using kustomize, can you make this is parameter for now?

config/ols-client.yaml
```diff
@@ -0,0 +1,30 @@
...
+      name: ols-client
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: ols-client
+        image: quay.io/brancz/krp-curl:v0.0.2
+        command:
+        - /bin/sh
+        - -c
+        - 'curl -v -s -k -H "Authorization: Bearer `cat /service-account/token`" https://ols-proxy.ols.svc:8443/docs'
```
bparees, line 17: it's not obvious to me what this is doing or why we need it

config/ols-rbac.yaml
```diff
@@ -0,0 +1,63 @@
+---
```
bparees, line 1: split up the roles/rbac that the kube-rbac proxy itself needs (to do subject access reviews) from the role/rbac we'd grant a user to let them invoke OLS

config/ols-rbac.yaml
```diff
@@ -0,0 +1,63 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: ols-proxy-backend-register
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ols-proxy-backend-register
+subjects:
+- kind: ServiceAccount
+  name: default
```
bparees, line 62: we should probably never use the default SA for anything, create a new one.

but also i'm not clear why the OLS SA needs this rolebinding?

bparees (commented):
Can we get an update to the readme or a new file that explains these resources and how they come together to define the rbac/what someone has to do to use them(including granting permission to users to use OLS)?

config/ols-rbac.yaml
williamcaban: It might be good to start with a an SA with read-only role (e.g. cluster-reader) for future querying of the system during  contextualization.

bparees: cluster context retrieval is not something we are looking at yet, if/when we get there it will probably need to be its own SA entirely so we can separate permissions needed to interrogate the cluster(it may even be an entirely separate process), from permissions OLS needs for running the client->OLS->LLM flows and validating requests.

So let's keep the SA permission focused on that latter bit for now.

config/ols-client.yaml
vishnuchalla: This is just to demonstrate an example on how some client can make requests.

config/ols.yaml
vishnuchalla: Changed as suggested.

config/ols.yaml
vishnuchalla: Yes parameterized this image param, using envs

config/ols-rbac.yaml
vishnuchalla: Makes sense. Changed as suggested.

config/ols-client.yaml
bparees: cool, got it.  maybe name it "ols-client-test" or "ols-client-sample" just to make it clear it's not really part of the OLS stack

config/ols-rbac.yaml
vishnuchalla: Actually those verbs don't add any value when we have `nonResourceURLs` defined. Its just path based action. Since the verb attribute is mandatory i am only keep `get` over there.

bparees: > Actually those verbs don't add any value when we have nonResourceURLs defined. Its just path based access.

that was my guess

> Since the verb attribute is mandatory I am only keeping get, even a dummy works TBH.

weird that it's required, but given that it is i'd say it's best to use a value that's clearly a noop, rather than a standard verb like `get`.  Some suggestesion: `na', `noop`, `none`

also maybe see what other people have used in this situation.

config/ols-rbac.yaml
vishnuchalla: OLS SA doesn't require the above rolebinding, its just an SA from client side. Create a new one for demonstrating the `client-test` example.

config/ols-rbac.yaml
vishnuchalla: Sorry, I just slipped on my previous comment. Looks like those verbs apply. To cross check this behavior I quickly cloned this [example](https://github.com/brancz/kube-rbac-proxy/tree/master/examples/non-resource-url-token-request) and tried to run it with `noop` verb, it failed with `403 Forbidden`. Below is the mapping for those verbs.

```
get: HTTP GET
create: HTTP POST
delete: HTTP DELETE
list: HTTP GET
update: HTTP PUT
patch: HTTP PATCH
```
I wasn't able to catch this earlier because I was looking at `/docs` and `/healthz` endpoints. They seem to allow all kinds of requests. So to check it properly, I took `/ols` endpoint, enabled `verb:get` and did a POST request to it and ran into `403 Forbidden Error` and the request was not even forwarded to server which is quite expected. 
```
INFO:     Started server process [1]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8080 (Press CTRL+C to quit)
```
But when I added `verb:create` I was able to make requests and get legit responses. So keeping only `get` and `create` verbs as they are sufficient for all the endpoints that we have as of now.

bparees: cool, so basically get==you can do http get on this url, and create==you can do a post on this url.

vishnuchalla: Yes correct.

vishnuchalla:
> Can we get an update to the readme or a new file that explains these resources and how they come together to define the rbac/what someone has to do to use them(including granting permission to users to use OLS)?

Can I do this documentation part in a different PR? This PR already went so long 😄

bparees:
>> Can we get an update to the readme or a new file that explains these resources and how they come together to define the rbac/what someone has to do to use them(including granting permission to users to use OLS)?
> Can I do this documentation part in a different PR? This PR already went so long 😄

You can defer the explanation part, but I think we should at least have an update to the readme about how to deploy these resources to a cluster.  It can be a new section under "Usage".  Given we're bootstrapping a team here one of the most important things we can do right now is make sure the people who join us in the future can get started w/o needing us to hold their hand.  (But there is certainly also a tradeoff, we don't want to waste time documenting things that are likely to change/become invalid)

vishnuchalla:
> > > Can we get an update to the readme or a new file that explains these resources and how they come together to define the rbac/what someone has to do to use them(including granting permission to users to use OLS)?
> 
> > Can I do this documentation part in a different PR? This PR already went so long 😄
> 
> You can defer the explanation part, but I think we should at least have an update to the readme about how to deploy these resources to a cluster. It can be a new section under "Usage". Given we're bootstrapping a team here one of the most important things we can do right now is make sure the people who join us in the future can get started w/o needing us to hold their hand. (But there is certainly also a tradeoff, we don't want to waste time documenting things that are likely to change/become invalid)

Okay, Let me get this done in this PR itself. Just one more experiment left with service account, will provide an update soon.

config/ols-rbac.yaml
vishnuchalla: OLS SA doesn't require this role binding. This is for client and can be in the client's test namespace, but looks like it has to use only `default` service account. This seems to be mandatory on `kube-rbac-proxy` end. I did an experiment by deploying client in a different namespace and created a different service account, but kube-proxy automatically looks for `default` service account no matter what SA we create in client's namespace for authorization. Relevant failure logs here:

```
I1115 02:18:47.899587       1 request.go:1172] Request Body: {"kind":"SubjectAccessReview","apiVersion":"authorization.k8s.io/v1","metadata":{"creationTimestamp":null},"spec":{"nonResourceAttributes":{"path":"/docs","verb":"get"},"user":"system:serviceaccount:ols-test:default","groups":["system:serviceaccounts","system:serviceaccounts:ols-test","system:authenticated"],"extra":{"authentication.kubernetes.io/pod-name":["ols-client-7cbl5"],"authentication.kubernetes.io/pod-uid":["b1227ff6-0b69-4a43-87f1-678e405ad560"]},"uid":"7a2f0494-47cf-429c-baf0-2b954daba0e0"},"status":{"allowed":false}}


I1115 02:18:47.901768       1 request.go:1172] Response Body: {"kind":"SubjectAccessReview","apiVersion":"authorization.k8s.io/v1","metadata":{"creationTimestamp":null,"managedFields":[{"manager":"kube-rbac-proxy","operation":"Update","apiVersion":"authorization.k8s.io/v1","time":"2023-11-15T02:18:47Z","fieldsType":"FieldsV1","fieldsV1":{"f:spec":{"f:extra":{".":{},"f:authentication.kubernetes.io/pod-name":{},"f:authentication.kubernetes.io/pod-uid":{}},"f:groups":{},"f:nonResourceAttributes":{".":{},"f:path":{},"f:verb":{}},"f:uid":{},"f:user":{}}}}]},"spec":{"nonResourceAttributes":{"path":"/docs","verb":"get"},"user":"system:serviceaccount:ols-test:default","groups":["system:serviceaccounts","system:serviceaccounts:ols-test","system:authenticated"],"extra":{"authentication.kubernetes.io/pod-name":["ols-client-7cbl5"],"authentication.kubernetes.io/pod-uid":["b1227ff6-0b69-4a43-87f1-678e405ad560"]},"uid":"7a2f0494-47cf-429c-baf0-2b954daba0e0"},"status":{"allowed":false}}
I1115 02:18:47.901872       1 auth.go:97] Forbidden (user=system:serviceaccount:ols-test:default, verb=get, resource=, subresource=). Reason: "".
```
The same request works fine when we have `default` SA bound. It always looks for `default` SA and fails no matter what service account we have on the client side. So, only default SA should be bound with client-side role bindings.

I see others using different SA [here](https://github.com/redhat-appstudio/sprayproxy/blob/main/config/sprayproxy-rbac.yaml#L61) and they do use a different [image](https://github.com/redhat-appstudio/sprayproxy/blob/main/config/sprayproxy.yaml#L48). I am not sure if they are able to make requests using a token or not, but I tried it in my local and it failed for me with the same error above.

Let me know if this is something not acceptable and if we need to raise a issue with `kube-rbac-proxy` team on this.

vishnuchalla:
> > > > Can we get an update to the readme or a new file that explains these resources and how they come together to define the rbac/what someone has to do to use them(including granting permission to users to use OLS)?
> > 
> > 
> > > Can I do this documentation part in a different PR? This PR already went so long 😄
> > 
> > 
> > You can defer the explanation part, but I think we should at least have an update to the readme about how to deploy these resources to a cluster. It can be a new section under "Usage". Given we're bootstrapping a team here one of the most important things we can do right now is make sure the people who join us in the future can get started w/o needing us to hold their hand. (But there is certainly also a tradeoff, we don't want to waste time documenting things that are likely to change/become invalid)
> 
> Okay, Let me get this done in this PR itself. Just one more experiment left with service account, will provide an update soon.

Added instructions in README.md

config/ols-rbac.yaml
bparees: If i'm understanding what you're saying correctly, regardless of what SA token the client provides, the rbac proxy is attempting to check if the default SA from the client's NS has the right permissions?


Keep in mind the end-goal here is going to be that a user can make requests to OLS using their own creds/token.  So the idea that kube-rbac is falling back to trying to use the default SA (from what NS?  it seems to have picked the client NS somehow, but how did it do that?) is going to be a problem for how we intend to use this. 

It must be a configuration issue.  Try reaching out to Jeff Cantrill.  Tag me in the conversation

Containerfile
```diff
@@ -12,4 +12,4 @@ USER 1001
 RUN pip install -r requirements.txt 
 
 # Run the application
-CMD uvicorn ols:app --host 0.0.0.0
\ No newline at end of file
+CMD uvicorn ols:app --host 0.0.0.0 --port 8080
```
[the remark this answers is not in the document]
vishnuchalla: Changed as suggested.

config/ols-rbac.yaml
vishnuchalla: Fixed it as per our discussion offline: https://redhat-internal.slack.com/archives/C05EG043LER/p1700070319423039

bparees:
lgtm for now.  Suggested followup refactoring:

separate the test-client resources from the core resources, so you can deploy just the core (ols, rbac proxy, service) w/o the test-client job/NS/etc.  (Since you're already using kustomize this should naturally work as different "environments" (`test` vs `run` or something)

I also have some nitpicks on names, but i'll submit a follow up proposal for those.

/lgtm

PR #6, opened 2023-11-14
add an extremely primitive conversation caching implementation

bparees:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/modules/conversation_cache.py
+++ b/modules/conversation_cache.py
@@ -1,0 +1,38 @@
+from collections import OrderedDict
+
+
+# TODO
+# this is not thread safe
+# this does not scale
+# this does not work in a distributed system
+# this is a placeholder to be replaced with something like memcached
+class LRUCache:
+    def __init__(self, max_size):
+        self.cache = OrderedDict()
+        self.max_size = max_size
+
+
+    # Get the value from the cache.
+    # Returns the value if the key exists, empty string if not
+    def get(self, key):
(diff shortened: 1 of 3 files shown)


vishnuchalla (approved):
approving this PR to help with merge conflicts problem.

modules/conversation_cache.py
```diff
@@ -0,0 +1,38 @@
...
+            return self.cache[key]
+        else:
+            return ""
+
+    # Adds a new key-value pair to the cache.  If the key already exists the new value will be appended to the existing value
+    def upsert(self, key, value):
+        oldVal=self.get(key)
+        if oldVal:
+            # TODO limit length of cached value
+            self.cache[key]=oldVal+"\n"+value
```
vishnuchalla, line 30: This operation looks a bit strange to when a key is already present. I hope it will be refined later.

bparees: the primary client usage currently is to append the new LLM response to the previous conversation history, hence why it behaves this way.

there's certainly no reason it couldn't be a more traditional cache in which the client is responsible for defining the new value and updating the key explicitly.

thoraxe:
Good enough?

thoraxe (approved):

PR #7, opened 2023-11-14
Adding Contributing doc

williamcaban:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

Adding CONTRIBUTING.md

### DIFF
--- a/modules/conversation_cache.py
+++ b/modules/conversation_cache.py
@@ -1,0 +1,38 @@
+from collections import OrderedDict
+
+
+# TODO
+# this is not thread safe
+# this does not scale
+# this does not work in a distributed system
+# this is a placeholder to be replaced with something like memcached
+class LRUCache:
+    def __init__(self, max_size):
+        self.cache = OrderedDict()
+        self.max_size = max_size
+
+
+    # Get the value from the cache.
+    # Returns the value if the key exists, empty string if not
+    def get(self, key):
+        if key in self.cache:
+            # Move the accessed key to the end to mark it as most recently used
+            self.cache.move_to_end(key)
+            return self.cache[key]
+        else:
+            return ""
+
+    # Adds a new key-value pair to the cache.  If the key already exists the new value will be appended to the existing value
+    def upsert(self, key, value):
+        oldVal=self.get(key)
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,9 +1,14 @@
+# ignore python and editors cache
 __pycache__
+venv
+.env*
+.vscode
+.DS_Store
+
+# ignore locally significat directories
```
bparees, line 8:
```suggestion
# ignore locally significant directories
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Install dependencies
+pip install -r requirements.txt
+```
+
+Happy hacking!
+
+## Before a PR
+
+If updating `requirements.txt` identify if doing so for the `main` branch or a release branch and follow the corresponding guidance. If a dependency is no longer required, remove it from the list.
```
bparees, line 38: i don't understand this part.  If you're updating requirements.txt for main branch, the PR will be against the main branch.

if you're doing it for a release branch (note: we're probably about to delete our release branches for now as they don't make sense right now) you'd submit the PR against the release branch.

so you don't need to "identity" which you're doing it for

bparees (commented):
couple comments and also if you can squash this down to a single commit to keep our history a bit cleaner it'd be appreciated (we can also squash on merge but then we get all the concatenated commit messages)

src/logger.py
```diff
@@ -0,0 +1,66 @@
+import logging, sys, os
```
bparees, line 1: pretty sure you didn't mean to include this file in this PR

williamcaban: yeah, this is not part of it. When I try to create the other PR it updated this one

CONTRIBUTING.md
williamcaban: The intent is to have the guidance for whoever cut the releases.

bparees: ok then maybe put this in a separate section about cutting releases

PR #11, opened 2023-11-15
Updating Containerfile to use UBI-based Python 3.11

williamcaban:
Updating Containerfile to use UBI-based Python 3.11. 

## Type of change

- [X] Refactor
- [X] New feature
- [ ] Bug fix
- [X] Optimization
- [ ] Documentation Update

## Description

- Updating Containerfile to use UBI-based Python 3.11. 
- Adding environment variables to optimize Python-based images and improve stdout logging behavior in Kubernetes environments. 
- Add explicit files and directories to the image to avoid accidental information leaking during the dev process. 
- Adding script wrapper to help developers build and test the container image locally

- Related Issue #
- Closes #

## Testing

To test the changes, execute the `./build-container.sh` and run the resulting container image.

Run image locally and successfully interacted with the API

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,15 +1,56 @@
-FROM registry.access.redhat.com/ubi9/python-39:1-143.1697647134
+FROM registry.access.redhat.com/ubi9/ubi-minimal
 
-# Add application sources with correct permissions for OpenShift
-USER 0
+ARG VERSION
+ARG APP_ROOT=/app-root
 
-WORKDIR /opt/app-root/src
-ADD . /opt/app-root/src/
-RUN chown -R 1001:0 /opt/app-root/src/
+RUN microdnf install -y --nodocs --setopt=keepcache=0 --setopt=tsflags=nodocs \
+    python3.11 python3.11-devel python3.11-pip jq shadow-utils \
+    && microdnf clean all --enablerepo='*' \
+    && useradd -r -u 1001 -g 0 -m -c "Default Application User" -d ${APP_ROOT} -s /bin/bash default
+
+# PYTHONDONTWRITEBYTECODE 1 : disable the generation of .pyc
+# PYTHONUNBUFFERED 1 : force the stadout and stderr streams to be unbufferred
+# PYTHONCOERCECLOCALE 0, PYTHONUTF8 1 : skip lgeacy locales and use UTF-8 mode
+ENV PYTHONDONTWRITEBYTECODE=1 \ 
+    PYTHONUNBUFFERED=1 \
+    PYTHONCOERCECLOCALE=0 \
+    PYTHONUTF8=1 \
+    PYTHONIOENCODING=UTF-8 \
+    LANG=en_US.UTF-8 \
+    PIP_NO_CACHE_DIR=off
+
+WORKDIR ${APP_ROOT}
+# Add explicit files and directories
+# (avoid accidental inclusion of local directories or env files or credentials)
+COPY modules ./modules
+COPY tools ./tools
+COPY logs ./logs
+COPY ./ols.py ./requirements.txt ./
+RUN chown -R 1001:0 ${APP_ROOT}
+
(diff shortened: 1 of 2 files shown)


Containerfile
```diff
@@ -1,15 +1,39 @@
...
+FROM registry.redhat.io/rhel9/python-311
+
+ARG VERSION
+ENV OLS_VERSION $VERSION
+
+# disable the generation of .pyc
+ENV PYTHONDONTWRITEBYTECODE 1
+# force the stadout and stderr streams to be unbufferred
+ENV PYTHONUNBUFFERED 1
+# skip lgeacy locales and use UTF-8 mode
```
bparees, line 10:
```suggestion
# skip legacy locales and use UTF-8 mode
```

Containerfile
```diff
@@ -1,15 +1,39 @@
-FROM registry.access.redhat.com/ubi9/python-39:1-143.1697647134
+FROM registry.redhat.io/rhel9/python-311
```
bparees, line 1: i think we're safer sticking w/ the ubi images, as they are not licensed/restricted compared to the rhel9 ones.

can we just use a ubi based image and then install/upgrade python within that image?  (yes i realize the build will take longer)

williamcaban: I thought rhel9 and ubi9 were the same. The intent is default to UBI9 but pointing to the Python 3.11 image general name not a subversion of it

bparees: my understanding is ubi is unencumbered, i don't think that's the case for rhel.

build-container.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
bparees, line 1: 1) if we are going to have developer scripts, we should put them in a scripts directory
2) i think this would be better encoded into a makefile with targets like "make images" that do this sort of thing (even if the make target then calls "scripts/build-container.sh")

Containerfile
```diff
@@ -1,16 +1,58 @@
...
-RUN pip install -r requirements.txt 
+# activate virtualenv with workaround RHEL/CentOS 8+
+ENV BASH_ENV="${APP_ROOT}/venv/bin/activate" \
+    ENV="${APP_ROOT}/venv/bin/activate" \
+    PROMPT_COMMAND=". ${APP_ROOT}/venv/bin/activate" \
+    PATH="${APP_ROOT}/venv/bin:${PATH}"
 
 # Run the application
-EXPOSE 8080
-CMD uvicorn ols:app --host 0.0.0.0 --port 8080
```
vishnuchalla, line 16: We need these two lines.

build-container.sh
thoraxe: so `makefile` in the project root, and then a `scripts` folder that houses things that aren't directly used by the app, and then all app stuff under `src`?

bparees: yeah that seems like a good fundamental structure

Containerfile
thoraxe: AFAIU the `EXPOSE` stanza is a courtesy and helps other things figure out what to do in the case of automation, but it is not required.
uvicorn default port is `8000` which seems fine to me. I'm not sure we need to force `8080`.

bparees: i will submit a followup to restore this

bparees: https://github.com/openshift/lightspeed-service/pull/13

ptal

bparees: > AFAIU the EXPOSE stanza is a courtesy and helps other things figure out what to do in the case of automation, but it is not required.

that is correct, it's not required, but it is a good thing to do

> uvicorn default port is 8000 which seems fine to me. I'm not sure we need to force 8080.

yeah i don't have a strong opinion either way on that one, but 8080 is a little more "standard" for an http app imho, regardless of why uvicorn picked 8000 as a default.  It's all somewhat moot since at some point this needs to move to TLS.

PR #18, opened 2023-11-17
Adding class to handle configuration and first interconnection of logging class

williamcaban:
Updating logging class.
Wrapped gradio interface into a class to remove gradio import from ols.py 
Created env variable to enable/disable the loading of the local UI interface 
Connecting ols.py to the logging class as a single instance on run. 
Minor updates to script and env file to reflect changes.

## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

See intro.

- Related Issue #
- Closes #

## Testing

To test this change I generated the container locally and made sure it is still operational interacting with the LLM.

### DIFF
--- a/modules/gradio_ui.py
+++ b/modules/gradio_ui.py
@@ -2,61 +2,73 @@ import gradio as gr
 import requests
 import json
 
-conversation_id=None
-
-def yaml_gen(prompt,history,use_history):
-    global conversation_id
-    # URL of the HTTP endpoint
-    url = "http://127.0.0.1:8080/ols"
-
-    # Headers for the HTTP request
-    headers = {
-        "Accept": "application/json",
-        "Content-Type": "application/json"
-    }
-
-    print("Using history: "+str(use_history))
-    # Body of the request (a JSON object with a "query" field)
+class gradioUI:
+    def __init__(self,
+                 ols_url="http://127.0.0.1:8080/ols",
+                 mount_path="/ui",
+                 conversation_id=None) -> None:
+        # class variable
+        self.ols_url=ols_url
+        self.mount_path=mount_path
+        self.conversation_id = conversation_id
     
-    data = {
-        "query": prompt
-    }
+        # ui specific
+        use_history=gr.Checkbox(value=True, label="Use history")
+        self.ui = gr.ChatInterface(self.yaml_gen,
+                                    additional_inputs=[
+                                        use_history
+                                    ])
 
-    if not use_history:
-        print("Ignoring conversation history")
-    elif use_history and conversation_id!=None:
-        data["conversation_id"]=conversation_id
-        print("Using conversation ID: "+conversation_id)
+    def yaml_gen(self, prompt, history, use_history=None):
+        # URL of the HTTP endpoint
+        url = self.ols_url
 
-    # Convert the data dictionary to a JSON string
-    json_data = json.dumps(data)
+        # Headers for the HTTP request
+        headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/json"
+        }
 
-    try:
-        # Make the HTTP POST request
-        response = requests.post(url, headers=headers, data=json_data)
+        print("Using history: "+str(use_history))
+        # Body of the request (a JSON object with a "query" field)
+        
+        data = {
+            "query": prompt
+        }
 
-        # Check if the request was successful (status code 200)
-        if response.status_code == 200:
-            print("Response JSON:", response.json())
-            conversation_id=response.json().get("conversation_id")
-            return response.json().get("response")
-        else:
-            print(f"Request failed with status code {response.status_code}")
-            print("Response text:", response.text)
-            return "Sorry, an error occurred: "+ response.text
-    except requests.RequestException as e:
-        # Handle any exceptions that may occur during the request
-        return f"An error occurred: {e}"
+        if not use_history:
+            print("Ignoring conversation history")
+        elif use_history and self.conversation_id!=None:
+            data["conversation_id"]=self.conversation_id
+            print(f"Using conversation ID: {self.conversation_id}")
 
+        # Convert the data dictionary to a JSON string
+        json_data = json.dumps(data)
 
-use_history=gr.Checkbox(value=True, label="Use history")
-ui = gr.ChatInterface(yaml_gen,
-                        additional_inputs=[
-                            use_history
-                        ])
+        try:
+            # Make the HTTP POST request
+            response = requests.post(url, headers=headers, data=json_data)
 
+            # Check if the request was successful (status code 200)
+            if response.status_code == 200:
+                print("Response JSON:", response.json())
+                self.conversation_id=response.json().get("conversation_id")
+                return response.json().get("response")
+            else:
+                print(f"Request failed with status code {response.status_code}")
+                print(f"Response text: {response.text}")
+                return "Sorry, an error occurred: "+ response.text
+            
+        except requests.RequestException as e:
+            # Handle any exceptions that may occur during the request
+            return f"An error occurred: {e}"
+
+    def mount_ui(self, fast_api_instance):
+        return gr.mount_gradio_app(fast_api_instance, 
+                                   self.ui, 
+                                   path=self.mount_path)
 
 if __name__ == "__main__":
-    ui.launch(show_api=False)   
+    gradioUI.ui.launch(show_api=False)   
 
 
--- a/modules/gradio_ui.py
+++ b/modules/gradio_ui.py
@@ -14,15 +14,12 @@ class gradioUI:
     
         # ui specific
         use_history=gr.Checkbox(value=True, label="Use history")
-        self.ui = gr.ChatInterface(self.yaml_gen,
+        self.ui = gr.ChatInterface(self.chat_ui,
                                     additional_inputs=[
                                         use_history
                                     ])
 
-    def yaml_gen(self, prompt, history, use_history=None):
-        # URL of the HTTP endpoint
-        url = self.ols_url
-
+    def chat_ui(self, prompt, history, use_history=None):
         # Headers for the HTTP request
         headers = {
             "Accept": "application/json",
@@ -47,7 +44,8 @@ class gradioUI:
 
         try:
             # Make the HTTP POST request
-            response = requests.post(url, headers=headers, data=json_data)
+            response = requests.post(
+                self.ols_url, headers=headers, data=json_data)
 
             # Check if the request was successful (status code 200)
             if response.status_code == 200:
--- a/ols.py
+++ b/ols.py
@@ -38,7 +38,7 @@ logger = config.logger
 if config.enable_ui:
     app=gradioUI().mount_ui(app)
 else:
-    logger.info("Local UI is disabled. To enable set OLS_ENABLE_UI=True")
+    logger.info("Embedded Gradio UI is disabled. To enable set OLS_ENABLE_UI=True")
 
 # application global variables
 base_completion_model = config.base_completion_model
@@ -56,7 +56,7 @@ def read_root():
 @app.get("/status")
 def root(request: Request):
     """
-    TODO: In the future should resond
+    TODO: In the future should respond
     """
     return {
         "message" : "This is the default endpoint for OLS",
--- a/ols.py
+++ b/ols.py
@@ -1,33 +1,24 @@
 # base python things
-from typing import Union
-from fastapi import FastAPI, HTTPException
-import gradio as gr
-from dotenv import load_dotenv
-from pydantic import BaseModel
-import os
 import uuid
+from typing import Union
+from pydantic import BaseModel
+from fastapi import FastAPI, HTTPException, Request
+
+# auxiliary functionalities
+from src.config import Config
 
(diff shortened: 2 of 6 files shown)


thoraxe:
I'm no python expert but I think this is alright

modules/gradio_ui.py
```diff
@@ -2,61 +2,73 @@
...
-        # Handle any exceptions that may occur during the request
-        return f"An error occurred: {e}"
-
-
-use_history=gr.Checkbox(value=True, label="Use history")
-ui = gr.ChatInterface(yaml_gen,
-                        additional_inputs=[
-                            use_history
-                        ])
+        print("Using history: "+str(use_history))
```
vishnuchalla, line 32: Do we want to replace them with log? Also lets use format specifiers instead of `+` concatenation operators.

modules/gradio_ui.py
```diff
@@ -2,61 +2,73 @@
...
+        # Body of the request (a JSON object with a "query" field)
+        
+        data = {
+            "query": prompt
+        }
+
+        if not use_history:
+            print("Ignoring conversation history")
```
vishnuchalla, line 40: Same as above, applies everywhere where we have print statements.

modules/gradio_ui.py
```diff
@@ -2,61 +2,73 @@
...
+        except requests.RequestException as e:
+            # Handle any exceptions that may occur during the request
+            return f"An error occurred: {e}"
 
+    def mount_ui(self, fast_api_instance):
+        return gr.mount_gradio_app(fast_api_instance, 
+                                   self.ui, 
+                                   path=self.mount_path)
 
 if __name__ == "__main__":
```
vishnuchalla, line 71: I think we can remove this if block as it is just for testing purposes.

modules/gradio_ui.py
```diff
@@ -2,61 +2,73 @@
...
+                                        use_history
+                                    ])
 
-def yaml_gen(prompt,history,use_history):
-    global conversation_id
-    # URL of the HTTP endpoint
-    url = "http://127.0.0.1:8080/ols"
+    def yaml_gen(self, prompt, history, use_history=None):
+        # URL of the HTTP endpoint
+        url = self.ols_url
```
bparees, line 24: what's the point of making a copy of this value?

modules/gradio_ui.py
```diff
@@ -2,61 +2,73 @@
...
-conversation_id=None
+class gradioUI:
+    def __init__(self,
+                 ols_url="http://127.0.0.1:8080/ols",
+                 mount_path="/ui",
+                 conversation_id=None) -> None:
+        # class variable
+        self.ols_url=ols_url
+        self.mount_path=mount_path
+        self.conversation_id = conversation_id
```
bparees, line 13: orthogonal to your PR but something I was wondering about....  there are multiple users hitting the gradio UI, do they each get their own instance of this class?  If not, the "conversation_id" would be shared between the two users which would be bad.

Not super critical for a debug ui, but something i'd like to understand about gradio in general if you happen to know. (And if nothing else, if it is indeed going to be shared then we should at least understand that constraint).

Same thing for the "use_history" config value/checkbox.

ols.py
```diff
@@ -1,68 +1,78 @@
...
-gr.mount_gradio_app(app,ui,path="/ui")
 
+config = Config()
+logger = config.logger
+
+# determine if need to enable local UI
+if config.enable_ui:
+    app=gradioUI().mount_ui(app)
+else:
+    logger.info("Local UI is disabled. To enable set OLS_ENABLE_UI=True")
```
bparees, line 41:
```suggestion
    logger.info("Embedded Gradio UI is disabled. To enable set OLS_ENABLE_UI=True")
```

it's not really a local UI, it's accessible on the same ip/port as the OLS apis themselves

ols.py
```diff
@@ -1,68 +1,78 @@
...
 @app.get("/healthz")
+@app.get("/readyz")
 def read_root():
     return {"status": "1"}
 
+@app.get("/")
+@app.get("/status")
+def root(request: Request):
+    """
+    TODO: In the future should resond
```
bparees, line 59:
```suggestion
    TODO: In the future should respond
```

ols.py
```diff
@@ -1,68 +1,78 @@
...
+base_completion_model = config.base_completion_model
+conversation_cache = LRUCache(100)
 
 def get_suid():
     return str(uuid.uuid4().hex)
 
-
 @app.get("/healthz")
+@app.get("/readyz")
 def read_root():
```
bparees, line 52: "read_root"?  can we give this a better name, if it's going to be the healthz/readyz endpoint?

(also those paths/endpoints may not always share the same implementation, but for now i guess it's fine)

scripts/build-container.sh
```diff
@@ -10,4 +10,4 @@ podman build --no-cache --build-arg=VERSION=${OLS_VERSION} -t ols:latest -f Cont
 # podman tag ols:latest quay.io/<org>/ols:$OLS_VERSION
 
 # To test-run for local development
-# podman run --rm -ti -p 8080:8080 localhost/ols:latest
\ No newline at end of file
+# podman run --rm -ti --env-file=.env -p 8080:8080 localhost/ols:latest
```
bparees, line 13: you want to encode this as a new Makefile target, like `run-image` ?

src/config.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def set_defaults(self):
+        """
+        Set default global required parameters if none was found
+        """
+        # set logs file
+        self.logfile = os.getenv("OLS_LOGFILE", "logs/ols.logs")
+
+        # enable local ui?
+        self.enable_ui = True if os.getenv("OLS_ENABLE_UI", True) in [True,"True"] else False
```
bparees, line 23: how about a case-insensitive string comparison to `true` instead?

src/config.py
```diff
@@ -0,0 +1,27 @@
...
+
+        # set default LLM model
+        self.base_completion_model = os.getenv("BASE_COMPLETION_MODEL", 
+                                               "ibm/granite-20b-instruct-v1")
```
bparees, line 27: I struggle w/ why this particular model value is set in config/defaulted, but none of the others are?  Should we be doing this for the "yaml gen" model, "happy response model", and "classification model" here too?  Is it because this config object is not going to be shared to those other modules?

(note that those might all be the same model, but having them able to be configured separately seems logical)

src/config.py
```diff
@@ -0,0 +1,27 @@
...
+        # Parse arguments & set logger
+        self.logger = logger if logger else Logger(
+            logger_name="default", logfile=self.logfile).logger
+
+    def set_defaults(self):
+        """
+        Set default global required parameters if none was found
+        """
+        # set logs file
+        self.logfile = os.getenv("OLS_LOGFILE", "logs/ols.logs")
```
bparees, line 20: i guess if i don't want file logging i need to set OLS_LOGFILE=/dev/null?

because if i don't set it at all, this default will take effect on me.  And i can't set it to "empty string" (or at least i'm not sure what would happen if i did....)

src/logger.py
```diff
@@ -26,18 +30,26 @@ class Logger:
...
-        Set LOG_LEVEL environment variable (e.g., INFO, DEBUG) to cocntrol verbosity
+        Set LOG_LEVEL or LOG_LEVEL_CONSOLE environment variable (e.g., INFO, DEBUG) 
+        to control general logging verbosity or console specific logging level
         ############################################################################
         """
         if show_message: print(msg)
 
+        self.logger_name=logger_name
         self.log_level=os.getenv("LOG_LEVEL", log_level)
+        self.log_level_console=os.getenv("LOG_LEVEL_CONSOLE", self.log_level)
```
bparees, line 48: does this need to explicitly invoke load_dotenv()? Otherwise i'm not sure you can be guaranteed load_dotenv() was called before you get here.  (you can't assume config was init'd before someone instantiates a logger, right?)

src/logger.py
```diff
@@ -26,18 +30,26 @@ class Logger:
...
         self.logfile=logfile
+        self.logfile_maxSize=1048576*100
+        self.logfile_backupCount=3
```
bparees, line 51: why don't we want to expose these as parameters?

bparees (commented):
bunch of nits/questions but I like the direction this takes us in

modules/gradio_ui.py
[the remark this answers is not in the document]
williamcaban: Agree 100%. I didn't wanted to change all of them at once but if that is the preference, sure. Let me know.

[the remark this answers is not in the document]
bparees: i have no preference, if it's on your list to come back to, that's fine with me

modules/gradio_ui.py
williamcaban: They should, but it is actually a little bit more complicated than that. The uvicorn execute a thread per session per endpoint. (each request is handle on a separate thread). In that case each thread accessing the UI expoing will have its own settings. Each one will get its own copy of everything in main app (ols.py) that it is needed for that "def" function on FastAPI and these are processed async. When aggregating endpoints in other ways (e.g. like route aggregation), the behavior can be different.

modules/gradio_ui.py
williamcaban: moving the function into a class without modifying the function

modules/gradio_ui.py
williamcaban: They should be move to use logs, right now I'm just moving it "as-is" as possible into a class.

modules/gradio_ui.py
williamcaban: yes. same answer. The class should eventually move to receive a logging instance to move all this print to a logging info or similar. right now, I'm just moving the code as-is as possible into a class

modules/gradio_ui.py
williamcaban: For dev & test purposes, having a "main" block to test the functionality as stand-alone it is easier for the developer. It is ignored during execution when imported from another file.

ols.py
williamcaban: Yes, I would rather map all these to a status function. For now, moving "as is"

scripts/build-container.sh
williamcaban: No particular preference. This is the exact command I use to test this during dev

src/config.py
williamcaban: Yes, we can add that. Right now it checks for boolean True or string "True" but it can be anything. When it comes from ENV it comes as string.

src/config.py
williamcaban: Actually, if OLS_LOGFILE is None the logger will not log to file, but this one is forcing it to its current behavior. Ideally this should just disable it by default unless explicitly enabled (because it is targeting a container)

src/config.py
williamcaban: I'm maintaining current behavior with these changes, but I agree, I rather have this only come from the environment definitions and not having one hardcoded anywhere.

src/logger.py
williamcaban: In ols.py Config is called and that is the one that create the logger instance. So, that way we know dotenv was invoked before creating the logger instance

src/logger.py
williamcaban: We should

modules/gradio_ui.py
bparees: is there a reason not to modify the function though?  let's not carry cruft in that we don't need

modules/gradio_ui.py
bparees: uvicorn having a thread per session doesn't tell me whether the gradio instance is shared between those threads, though...

i haven't tested w/ your refactor, but i just tested the code from main and indeed the conversation id is shared between two browser sessions (I started one in chrome and one in FF).  So at least in the implementation in main, there is only a single "Gradio" object instance, with its associated "conversation id" field, and once it's set by one interaction in one browser, it gets reused by interactions in the other browser.

again, not critical for a debugging tool right now, but important to understand (and probably has implications for other spots where we might be using class instance variables and the class is shared between requests/sessions)

src/config.py
bparees: > Actually, if OLS_LOGFILE is None the logger will not log to file

yeah, but as a user who's just invoking the python app, i cannot set it to "None", because all i can do is set the env variable, or not set the env variable.  That's my point.

bparees: (and if i don't set the env variable, then this code is going to default a value)

src/logger.py
bparees: seems unsafe to assume no future refactor is going to invalidate the assumption about that ordering.

At a minimum some sort of comment/documentation is needed about the assumptions this code is making/how it can be invoked.

modules/gradio_ui.py
williamcaban: A quick test using two browser sessions seems they are independent but this needs better testing

src/config.py
bparees: i think you've updated this to default to None now, so people who don't want file logging can just not set the env var.  resolving.

PR #20, opened 2023-11-21
Add env vars for controlling image build behavior

bparees:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/scripts/build-container.sh
+++ b/scripts/build-container.sh
@@ -4,10 +4,11 @@
 OLS_VERSION=v0.0.0
 
 # To build container for local use
-podman build --no-cache --build-arg=VERSION=${OLS_VERSION} -t ols:latest -f Containerfile
-
-# To publish to a registry tag registry name
-# podman tag ols:latest quay.io/<org>/ols:$OLS_VERSION
+if [ -z $OLS_NO_IMAGE_CACHE ]; then
+  podman build --no-cache --build-arg=VERSION=${OLS_VERSION} -t ${OLS_API_IMAGE:-quay.io/openshift/lightspeed-service-api:latest} -f Containerfile
+else
+  podman build --build-arg=VERSION=${OLS_VERSION} -t ${OLS_API_IMAGE:-quay.io/openshift/lightspeed-service-api:latest} -f Containerfile
+fi
 
 # To test-run for local development
-# podman run --rm -ti -p 8080:8080 localhost/ols:latest
+# podman run --rm -ti -p 8080:8080 ${OLS_API_IMAGE:-quay.io/openshift/lightspeed-service-api:latest}


bparees:
/assign @williamcaban 

because it's his script and he's got opinions about the use of image caches :)

i'm ok defaulting this either way (use the cache by default, or don't use the cache by default), but making it something we can control seems useful/important.

scripts/build-container.sh
```diff
@@ -4,10 +4,14 @@
...
 # To build container for local use
-podman build --no-cache --build-arg=VERSION=${OLS_VERSION} -t ols:latest -f Containerfile
+if [ -z $OLS_NO_IMAGE_CACHE ]; then
+  podman build --no-cache --build-arg=VERSION=${OLS_VERSION} -t ${OLS_API_IMAGE:-quay.io/openshift/lightspeed-service-api:latest} -f Containerfile
+else
+  podman build --build-arg=VERSION=${OLS_VERSION} -t ${OLS_API_IMAGE:-quay.io/openshift/lightspeed-service-api:latest} -f Containerfile
+fi
 
 # To publish to a registry tag registry name
 # podman tag ols:latest quay.io/<org>/ols:$OLS_VERSION
```
bparees, line 14: not sure what we want to do w/ this, in theory having OLS_API_IMAGE as a configurable env var should make this step unnecessary

scripts/build-container.sh
```diff
@@ -4,10 +4,11 @@
 OLS_VERSION=v0.0.0
 
 # To build container for local use
-podman build --no-cache --build-arg=VERSION=${OLS_VERSION} -t ols:latest -f Containerfile
-
-# To publish to a registry tag registry name
-# podman tag ols:latest quay.io/<org>/ols:$OLS_VERSION
```
bparees, line 10: tagging step not needed any more because you can just use OLS_API_IMAGE to build/tag the image with the appropriate name in one shot.

scripts/build-container.sh
williamcaban: Correct, the version is not necessary. Is there a good way to encode the version in the container so that in troubleshooting we can check the version is the correct one and it is not the system using an older version of the image? (or should this be done by the Pod.spec definition and not on build)?

scripts/build-container.sh
williamcaban: Okay. It is commented out so it can be removed.

/lgtm

scripts/build-container.sh
bparees: OCP encodes git commits in the container labels, but i believe that's done as part of the build process, not something coded in each repo's dockerfile/build script.

take a look at

`docker inspect quay.io/openshift/lightspeed-service-api:latest`

has
```
                "io.k8s.description": "AI-powered OpenShift Assistant Service.",
                "io.k8s.display-name": "OpenShift LightSpeed Service",
                "io.openshift.build.commit.author": "",
                "io.openshift.build.commit.date": "",
                "io.openshift.build.commit.id": "29c08c736fea065f0e6faff557a326e49533716b",
                "io.openshift.build.commit.message": "",
                "io.openshift.build.commit.ref": "main",
                "io.openshift.build.name": "",
                "io.openshift.build.namespace": "",
                "io.openshift.build.source-context-dir": "",
                "io.openshift.build.source-location": "https://github.com/openshift/lightspeed-service",
```

that's probably good enough for now.

PR #21, opened 2023-11-22
Refactoring/Restructuring the project

vishnuchalla:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing


CONTRIBUTING.md
```diff
@@ -29,6 +29,9 @@ pip install --upgrade pip
 
 # Install dependencies
 pip install -r requirements.txt
+
+# Code formatting (Run this as a pre-commit step for your code changes)
+black .
```
bparees, line 34: we should add this as a part of the validation presubmit job

bparees: pursuing adding this test job here: https://github.com/openshift/release/pull/45980

bparees: and as a "make verify" target in this repo.

requirements.txt
```diff
@@ -8,3 +8,4 @@ llama_index
 torch
 transformers
 uvicorn
+black
```
bparees, line 11: do we actually need to add this as a project requirement?  i'd prefer not to since we don't actually need it at runtime, so we don't need/want to be adding it to our published images.

Seems like something developers can just install globally on their machine

bparees:
my #1 comment on this PR would be to please try to redo your changes using "git mv" so that git can track the file moves/renames.  It'll make it a lot easier to see what the actual delta is beyond directory restructuring.  (usually git is pretty good at tracking file moves, and it caught at least a few of yours, but for whatever reason most of them it is treating as deletes+adds)

CONTRIBUTING.md
vishnuchalla: Acknowledged. Adding it as `make verify` target in this repo.

requirements.txt
vishnuchalla: Yes, sounds good to me.

vishnuchalla:
> my #1 comment on this PR would be to please try to redo your changes using "git mv" so that git can track the file moves/renames. It'll make it a lot easier to see what the actual delta is beyond directory restructuring. (usually git is pretty good at tracking file moves, and it caught at least a few of yours, but for whatever reason most of them it is treating as deletes+adds)

I already did `git mv`. But looks like git treats them as `deletes+adds` if we make significant changes in a file that has been moved. For example, when I `git mv` a file, I see
```
renamed:    modules/conversation_cache.py -> src/cache/conversation_cache.py
```
and once I make decent amount of changes in that file content and do a `git add -A`, I see it being treated as `deletes+adds`
```
deleted:    modules/conversation_cache.py
new file:   src/cache/conversation_cache.py
```
I am not sure on what we can do here. Looks like git's default behavior.

bparees:
> I am not sure on what we can do here. Looks like git's default behavior.

weird, ok, thanks for looking into it

bparees:
@williamcaban you're probably the only other person qualified to have strong opinions about project/code structure at this point, do you have any concerns/feedback on this?

otherwise i'm inclined to merge it and get back to iterating on the code functionality itself (config, logging additions we discussed, as well as the obvious core funcionality)

Makefile
```diff
@@ -12,4 +12,7 @@ images:
 	scripts/build-container.sh
 
 run:
-	uvicorn ols:app --reload --port 8080
+	uvicorn app.main:app --reload --port 8080
+
+verify:
+	pip install black && black . --check
```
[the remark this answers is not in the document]
vishnuchalla: Acknowledged with the future idea.

thoraxe:
* I don't think YAML generator belongs inside the `validation` folder
* I don't think yes_no_classifier belongs in the `response` folder.

I think this is a little over-zealous optimization/segmentation on the modules. I don't think we need this much division.

bparees:
/lgtm

thanks for giving us a solid starting point for a repo structure

PR #22, opened 2023-11-22
Use python3.11 base image, save time on python install

bparees:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,10 +1,10 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal
+FROM registry.access.redhat.com/ubi9/python-311:latest
 
 ARG VERSION
 ARG APP_ROOT=/app-root
 
 RUN microdnf install -y --nodocs --setopt=keepcache=0 --setopt=tsflags=nodocs \
-    python3.11 python3.11-devel python3.11-pip jq shadow-utils \
+    jq shadow-utils \
     && microdnf clean all --enablerepo='*' \
     && useradd -r -u 1001 -g 0 -m -c "Default Application User" -d ${APP_ROOT} -s /bin/bash default
 


bparees:
/assign @williamcaban 

@vishnuchalla heads up

williamcaban:
The Python3.11 contains a considerable number of pre installed unnecessary dependencies which will make the final image significantly larger.

bparees:
> The Python3.11 contains a considerable number of pre installed unnecessary dependencies which will make the final image significantly larger.

do you have a comparison of the two final image sizes w/ this base vs ubi-minimal?

i'm not entirely surprised, but given how much stuff our own deps add anyway, i'd like to see that the delta is significant enough to justify the optimization.

williamcaban:
Here is a more detailed comparison:
- Changing to the python image removes the option to use microdnf and instead dnf should be used (so this would need to be updated as well)
- When using the ubi9/python-311 image, it is already unprivileged so no additional dnf install can be executed and it will fail with `Error: This command has to be run with superuser privileges (under the root user on most systems).`. This means the Containerfile must not have any RUN statement for this 
- The same applies for the statement `RUN chown -R 1001:0 ${APP_ROOT}`, it failed to execute and must be remove when using the python image
- The resulting image when using the Python 3.11 UBI image is of 2.2GB (compressed) vs 1.2GB (compressed) when using the ubi-minimal
- The resulting image from the Python 3.11 UBI has 475 rpm packages installed, versus 123 packages the resulting image from uni-minimal (including the installed ones during build)
- In both cases the same python libraries dependencies need to be pip installed

bparees:
> Changing to the python image removes the option to use microdnf and instead dnf should be used (so this would need to be updated as well)

setting aside the size addition that may come w/ dnf vs microdnf, this is a trivial change and not really an obstacle (but i should have caught it before creating the PR...I assumed since both claim to be "ubi" images, they'd both use microdnf.  Interestingly the python 3.11 image is also complaining about missing subscriptions(but this does not cause a failure), unlike the ubi-minimal based image. Perhaps due to microdnf vs dnf. 

> When using the ubi9/python-311 image, it is already unprivileged so no additional dnf install can be executed and it will fail with Error: This command has to be run with superuser privileges (under the root user on most systems).. This means the Containerfile must not have any RUN statement for this

We can just set "USER 0" to work around that issue and then set it back to "USER 1001", so this is not an obstacle.

> The same applies for the statement RUN chown -R 1001:0 ${APP_ROOT}, it failed to execute and must be remove when using the python image

same thing here


> The resulting image when using the Python 3.11 UBI image is of 2.2GB (compressed) vs 1.2GB (compressed) when using the ubi-minimal

strange, when i built it after fixing up the issues w/ setting USER 0 and microdnf vs dnf, i got a size of 6.6gigs(!!!) for the python 3.11 based image, and 5.6gigs for the ubi-minimal based one.  So delta of 1 gig either way.

Given the 1 gig delta, i'll close this for now.  But we can make use of this image in CI if we need to run a python 3.11 client, so i'm still pursuing adding it here: https://github.com/openshift/release/pull/45981

PR #23, opened 2023-11-27
Implementing cache factory class

vishnuchalla:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Implemented a factory class to toogle between caches based on the ENV value. Default would be in-memory lru.
Implemented a thread safe redis cache using singleton design pattern. It also ships with a route that help us to look at some of the stats on the cache and its usage.

Related Issue: https://issues.redhat.com/browse/OLS-31

## Testing
Tested building my clone image in local setting. We can toogle between caches using `OLS_CONVERSATION_CACHE` env variable. Default would be LRU.
## local testing (By default uses LRU)
Run the below command to spin up the server
```
uvicorn app.main:ols --reload --port 8080
```
Make a request using this command
```
curl -X 'POST' 'http://127.0.0.1:8080/ols' -H2 'accept: application/json' -H 'Content-Type: application/json' -d '{"query": "write a deployment yaml for the mongodb image"}'
```

## Redis (set OLS_CONVERSATION_CACHE="redis" in your .env)
Deploy in cluster manifests using below commands
```
export OLS_IMAGE=`quay.io/vchalla/lightspeed-service:latest`
export OLS_REDIS_IMAGE=`quay.io/vchalla/redis-stack:latest`
kustomize build . | envsubst | oc apply -f -
```
### Cache settings
```
# redis-cli CONFIG GET maxmemory
1) "maxmemory"
2) "524288000"
# redis-cli CONFIG GET maxmemory-policy
1) "maxmemory-policy"
2) "allkeys-lru"
```
### Cache insights
Screenshot from 2023-11-26 22-13-21

Screenshot from 2023-11-26 22-14-20

Screenshot from 2023-11-26 22-12-38
### Results
```
Answer:
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mongodb
  labels:
    app: mongodb
spec:
  replicas: 1
  selector:
    matchLabels:
      app: mongodb
  template:
    metadata:
      labels:
        app: mongodb
    spec:
      containers:
      - name: mongodb
        image: mongo:4.2.3
        ports:
        - containerPort: 27017

2023-11-27 03:20:11,709 [conversation_cache.py:59] INFO: Cache Hits: 0
2023-11-27 03:20:11,709 [conversation_cache.py:60] INFO: Cache Misses: 1
INFO:     10.128.2.206:0 - "POST /ols HTTP/1.1" 200 OK
```


config/ols-redis.yaml
```diff
@@ -0,0 +1,52 @@
...
+    matchLabels:
+      app: redis
+  template:
+    metadata:
+      labels:
+        app: redis
+    spec:
+      containers:
+      - name: redis
+        image: ${OLS_REDIS_IMAGE}
```
bparees, line 49: w/ the reliance on "envsubst" to populate these things, we need to provide defaults somewhere.

vishnuchalla:
> > @vishnuchalla it looks like this completely ditches the LRU cache and "forces" Redis.
> 
> +1 on keeping the in-process cache available as an option for local dev mode. Especially since this not only requires you spin up redis, but really requires you have an entire cluster to deploy to (or spin up redis yourself and fake the hostname)

Okay, let me go ahead and make this change backward compatible. Meaning, our logic will check for redis cache first and if its unreachable it should fallback to our default LFU cache.

bparees:
> Meaning, our logic will check for redis cache first and if its unreachable it should fallback to our default LFU cache.

probably easier/better to just make it configured via an env variable.  Because we want to fail for real if we can't hit redis in production, not fallback to the local cache.

vishnuchalla:
> > Meaning, our logic will check for redis cache first and if its unreachable it should fallback to our default LFU cache.
> 
> probably easier/better to just make it configured via an env variable. Because we want to fail for real if we can't hit redis in production, not fallback to the local cache.

Sounds good to me. Implemented a factory class based on this idea.

config/ols-redis.yaml
vishnuchalla: kustomize doesn't provide a way to achieve this. For now we have to rely on "evnsubst". And in future we can make our python code to deploy these resources by populating those values in the runtime.

src/constants.py
williamcaban, line 1: Can we integrate these into the config class instead of having another file with config params?

src/cache/in_memory_cache.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self, key, value):
+        """Initialize a ListNode with a key, value, and frequency."""
+        self.key = key
+        self.val = value
+        self.freq = 1
+
+
+class InMemoryCache:
+    """An in-memory LFU cache implementation using a combination of a dictionary and an OrderedDict."""
```
bparees, line 18: you've converted this from an LRU to an LFU cache, but that is not what we want for this use case.

If a user iterates on a single prompt 100 times, and then goes away, and then another 50 users each iterate twice, the first cache entry we want to throw away is the old user who iterated 100 times but hasn't been back recently, not the 50 users who are currently active.

What matters is how recently the cache entry was referenced, not how many times.

src/constants.py
```diff
@@ -167,3 +167,12 @@
 PRODUCT_DOCS_PERSIST_DIR = "./vector-db/ocp-product-docs"
 SUMMARY_INDEX = "summary"
 SUMMARY_DOCS_PERSIST_DIR = "./vector-db/summary-docs"
+
+# cache constants
+IN_MEMORY_CACHE = "in-memory"
+IN_MEMORY_CACHE_MAX_SIZE = 1000
```
bparees, line 173:
```suggestion
IN_MEMORY_CACHE_MAX_ENTRIES = 1000
```

makes it clearer what the units are

src/constants.py
```diff
@@ -167,3 +167,12 @@
...
+REDIS_CACHE = "redis"
+REMOTE_CACHE_HOST = "redis-stack.ols.svc"
+REMOTE_CACHE_PORT = 6379
+REMOTE_CACHE_MAX_MEMORY = "500mb"
+REMOTE_CACHE_MAX_MEMORY_POLICY = "allkeys-lfu"
```
bparees, line 178: per above, can we make this an lru policy instead?

src/constants.py
vishnuchalla: Changed as suggested.

src/cache/cache_factory.py
```diff
@@ -0,0 +1,55 @@
...
+
+        Returns:
+            None
+        """
+        pass
+
+
+class CacheFactory:
+    @staticmethod
+    def conversational_cache() -> Cache:
```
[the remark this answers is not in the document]
vishnuchalla: Done

src/cache/in_memory_cache.py
vishnuchalla: Changed it to LRU.

src/constants.py
vishnuchalla: Attaching offline discussion for more details: https://redhat-internal.slack.com/archives/C05EG043LER/p1701352818900439

src/constants.py
```diff
@@ -167,3 +167,12 @@
 PRODUCT_DOCS_PERSIST_DIR = "./vector-db/ocp-product-docs"
 SUMMARY_INDEX = "summary"
 SUMMARY_DOCS_PERSIST_DIR = "./vector-db/summary-docs"
+
+# cache constants
+IN_MEMORY_CACHE = "in-memory"
+IN_MEMORY_CACHE_MAX_ENTRIES = 1000
+REDIS_CACHE = "redis"
+REMOTE_CACHE_HOST = "redis-stack.ols.svc"
```
bparees, line 175: not critical at the moment but for the future this should probably be configurable so devs can stand up a redis service on their laptop and point to it.

of course they can use /etc/hosts to make it work, but it would be nicer if they could just set it to localhost via an env.

bparees (commented):
i'm going to /lgtm this as is, but i'm going to try to follow it up w/ adding some unit tests for the memory cache, just to further build out our test framework muscles.

i might also take on the follow up of making the redis host configurable, unless you explicitly want to handle that, @vishnuchalla

src/constants.py
vishnuchalla: Yes I had this as a #TODO in my mind while I was writing the code. Should have dropped a comment over there, my bad 🙁

PR #26, opened 2023-11-28
Pytest skel

thoraxe:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Implements the basics of pytest testing.

N/A

## Testing
`make test`!!! 🥳


Makefile
```diff
@@ -1,12 +1,16 @@
-.PHONY: unit e2e-test
+.PHONY: test
```
bparees, line 1: not sure what you were trying to do here, but i don't think this is right:
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

if you want to add a new test target that invokes other targets, it would look like:

```
test:  test-unit test-e2e
```

bparees:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_lightspeed-service/26/pull-ci-openshift-lightspeed-service-main-unit/1729607918023610368

failed because our "root" image is actually golang based.

need to fix up the unit test job like https://github.com/openshift/release/pull/45980 does

i'll submit a PR and rehearse it.

bparees:
/hold

need to fix the makefile targets

bparees:
@thoraxe well we've fixed your pip problem, but there's still a failure in the job you'll need to look at:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_lightspeed-service/26/pull-ci-openshift-lightspeed-service-main-unit/1729616953988878336

```
ImportError while importing test module '/home/prow/go/src/github.com/openshift/lightspeed-service/test_ols.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test_ols.py:1: in <module>
    from fastapi.testclient import TestClient
E   ModuleNotFoundError: No module named 'fastapi'
```

bparees:
guessing you need to pip install requirements.txt *and* requirements-test.txt

Makefile
```diff
@@ -1,12 +1,16 @@
...
 
 test-unit:
+	@echo "Installing test deps..."
+	pip install -r requirements-test.txt
```
bparees, line 5: i think you need to pip install requirements.txt too

Makefile
thoraxe: i did a quick read on `.PHONY` and forgot to change this back, but I think it was wrong to begin with anyway.

tl;dr
`.PHONY` needs to include any targets that might also be files. I don't think we would have a file called `e2e-test` or `test-unit`, so this `.PHONY` is probably irrelevant.

Makefile
thoraxe: Fixed.

bparees: thinking about this more, i don't think we want this make target to install any deps, similar to my reasoning here:  https://github.com/openshift/lightspeed-service/pull/21/files#r1409664767

(it also takes additional time, even when the deps are already installed, which will be annoying to local developers).

i think we should have a couple more make targets:
`make install-deps`
`make install-test-deps`

and then either another target invokes the install-deps/test-deps targets and the test-unit target, or something like that.

but we can follow up w/ that i guess.

test_ols.py
```diff
@@ -0,0 +1,69 @@
+from fastapi.testclient import TestClient
+
+from unittest.mock import MagicMock
+
+from ols import app
+
+client = TestClient(app)
```
bparees, line 7: you were wondering what to classify this test as before..... since it is standing up the server and invoking it, but not using a real backend LLM, i'd say it's technically an "integration test", so we should probably call it that in the make target and CI job.

but again we can follow up on that after we land this.

test_ols.py
```diff
@@ -0,0 +1,69 @@
...
+        testing extra ugly
+        """
+
+        def __init__(self, *args, **kwargs):
+            pass
+
+        def __call__(self, *args, **kwargs):
+            return "test response"
+
+    monkeypatch.setattr(modules.model_context, "LangChainInterface", MockChainInterface)
```
bparees, line 61: i think the need to do this in this way speaks to the fact that `model_context` is insufficiently configurable, currently.

but again, a follow up.

bparees (commented):
several follow ups, but good enough for now to get a stake in the ground

test_ols.py
thoraxe: Is your comment about the use of the monkeypatch, or the use of `test response`? The former is due to the way Python works -- configuration won't change this. The latter is because we need to simulate some kind of expected response and test that the endpoint actually does the thing.

test_ols.py
thoraxe: That sounds good to me. However, with that reasoning, every endpoint test on FastAPI will then be an "integration" test. I don't know if this is good or bad.

test_ols.py
bparees: it's about the monkeypatch and i'm not sure why configuration can't change it.  Model_context is choosing which interface to instantiate, you could make that configurable so that model_context can be configured to use the mockinterface instead of the ibm one.  (or use an openai one, if we wanted to make model_context be completely model agnostic)

bparees: i'll agree that it's possible that monkeypatch is the "preferred" way to do this in python, at least in some cases, but if we want OLS to be pluggable in terms of what model providers it can talk to, we won't do that via monkeypatch and once we can do that, "mock provider" becomes one of the providers we should be able to configure.

(and i think even if we don't want to allow users to swap in alternate providers, we're likely to want to be able to swap in providers in the short term for testing/evaluation)

thoraxe: I think we may be saying the same thing actually, sort of.

I'm not sure what you would tell `get_watsonx_predictor` to use that isn't returning a `LangChainInterface`, because that's what we're using to call the LLM -- a `LangChainInterface`.

In other words, `model_context` (the file) includes a helper function, `get_watsonx_predictor` which takes some (limited) args and returns a `LangChainInterface`.

Are you suggesting that `get_watsonx_predictor` should somehow _not_ return a `LangChainInterface`, based on some configuration? Because, if that was the case, how would we be calling the LLM? `LangChainInterface` is how (internally) we call the LLM.

thoraxe:
@vishnuchalla 100% agree. Once your refactor lands I will refactor the tests appropriately.

PR #28, opened 2023-11-30
minor refactor of tests location, adds junit output

thoraxe:
## Type of change

- [X] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

minor refactor of tests location, adds junit output

N/A

## Testing
Tests still run!


Makefile
```diff
@@ -1,18 +1,20 @@
-.PHONY: test-unit test-e2e
+.PHONY: test-unit test-e2e pytest
```
bparees, line 1: don't need this.  you only need a phony if you're trying to ensure there's no collision between a make target and a file.

pytest isn't a target

Makefile
```diff
@@ -1,18 +1,20 @@
...
+	# Command to run unit tests goes here
+	python -m pytest tests/ --junit-xml=tests/test_results/results.xml
 
 test-e2e:
 	@echo "Installing test deps..."
 	pip install -r requirements.txt
 	pip install -r requirements-test.txt
 	@echo "Running end-to-end tests..."
 	# Command to run e2e tests goes here
+	python -m pytest tests/ --junit-xml=tests/test_results/results.xml
```
bparees, line 17: doesn't make sense to run the same set of tests on both targets, let's leave this out for now.

Makefile
thoraxe: that's an accident. will fix. I was going to make a `pytest` target but then thought better of it, exactly related to your other comment.

Makefile
thoraxe: we're only even running `test-unit` right now, so I'm not sure we even need the `test-e2e` target -- would it make more sense to zero-out the `test-e2e` target or remove it entirely?

bparees: zero it out but leave the target.  it reflects our aspirations/intent :)

bparees:
junit.xml output location is definitely not right since it didn't even get collected by prow, let alone visualized:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_lightspeed-service/28/pull-ci-openshift-lightspeed-service-main-unit/1730328210928832512/

but we will follow up on that

/lgtm
/unhold

PR #29, opened 2023-12-01
Add unit tests for in memory cache

bparees:
also refactors our Makefile to separate dep installation from test execution,
adds a new target for formatting our code

## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing


bparees:
@vishnuchalla ptal.

note that the corresponding updates to our CI jobs are being made here:
https://github.com/openshift/release/pull/46324

so this PR won't pass CI until that PR is merged (and that PR can't pass rehearsals because it needs this PR to merge.  it's not worth the effort of doing this is a backwards compatible way, so i'm just gonna merge the release PR and then run tests here)

tests/unit/test_cache.py
```diff
@@ -0,0 +1,43 @@
...
+    assert cache.get("key1") == "value1"
+
+
+def test_insert_or_append_existing_key(cache):
+    cache.insert_or_append("key1", "value1")
+    cache.insert_or_append("key1", "value2")
+    assert cache.get("key1") == "value1\nvalue2"
+
+
+def test_insert_or_append_overflow(cache):
```
vishnuchalla, line 23: I don't think this is a unit test. Because, We should never have a for loop logic to iterate over same piece of code in a unit-test (like in source code). Also We should just be testing the pass or fail scenarios of the function/method under test covering all its lines of code.

bparees: > We should never have a for loop logic to iterate over same piece of code in a unit-test (like in source code).

i think you're taking an overly narrow approach to how a "unit test" is defined.

"A unit test is a type of software test that focuses on components of a software product. The purpose is to ensure that each unit of software code works as expected. A unit can be a function, method, module, object, or other entity in an application's source code."

https://brightsec.com/blog/unit-testing/

> Also We should just be testing the pass or fail scenarios of the function/method under test covering all its lines of code.

i'm not sure what you mean by this

bparees: we can certainly change the capacity and i considered doing so, but the test currently runs incredibly quick already so it didn't seem worth it.  The entire unit test suite runs in 0.11s on my laptop.

i'll update it, but that is why i didn't originally.

Makefile
```diff
@@ -1,21 +1,37 @@
...
+run:
+	uvicorn app.main:app --reload --port 8080
+
+test: test-unit test-integration test-e2e
+
+test-unit:
 	@echo "Running unit tests..."
 	# Command to run unit tests goes here
-	python -m pytest tests/ --junit-xml=tests/test_results/results.xml
+	python -m pytest tests/unit --junit-xml=tests/test_results/unit/results.xml
```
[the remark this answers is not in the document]
bparees: done

PR #30, opened 2023-12-04
Output junit.xml to a spot where CI jobs will pick it up

bparees:
- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@
 # See: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
 .PHONY: test test-unit test-e2e images run format verify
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,6 +3,9 @@
 # See: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
 .PHONY: test test-unit test-e2e images run format verify
 
+ARTIFACT_DIR := $(if $(ARTIFACT_DIR),$(ARTIFACT_DIR),tests/test_results)
```
bparees, line 6: i debated defaulting this to `/dev/null` so you don't get these files unless you explicitly set an env var indicting where you want them to go.

@vishnuchalla @thoraxe  any strong opinions?

otherwise i think this is good to go

PR #31, opened 2023-12-05
Unit test skeleton for our internal modules

thoraxe:
## Type of change

- [X] Refactor
- [X] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Creates a first example of how to test our internal modules.

* yes_no_classifier test, an extremely basic one
* mock class for LLMChain which is used by all our modules - takes an argument to simulate whatever response you want to guarantee so that you can base your tests off of it
* fixes a bug in the Logging class
* fixes a bug in the yes_no_classifier

N/A

- [X] If it is a core feature, I have added thorough tests.

## Testing
`make test` succeeded.

### DIFF
--- a/src/query_helpers/yes_no_classifier.py
+++ b/src/query_helpers/yes_no_classifier.py
@@ -56,10 +56,10 @@ class YesNoClassifier:
 
         response = llm_chain(inputs={"statement": statement})
 
-        # strip  from the response
-        clean_response = response["text"].split("")[0]
+        self.logger.info(f"{conversation} bare response: {response}")
+        self.logger.info(f"{conversation} yes/no response: {response['text']}")
 
-        self.logger.info(f"{conversation} yes/no response: {clean_response}")
+        if response["text"] not in ["0", "1", "9"]:
+            raise ValueError("Returned response not 0, 1, or 9")
 
-        # TODO: handle when this doesn't end up with an integer
-        return int(clean_response)
+        return int(response["text"])
--- a/tests/mock_classes/llm_chain.py
+++ b/tests/mock_classes/llm_chain.py
@@ -1,0 +1,20 @@
+def mock_llm_chain(retval):
+    class MockLLMChain:
+        """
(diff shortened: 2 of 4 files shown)


src/query_helpers/yes_no_classifier.py
```diff
@@ -56,10 +56,11 @@ def classify(self, conversation, statement, **kwargs):
...
-        # strip  from the response
-        clean_response = response["text"].split("")[0]
+        self.logger.info(f"{conversation} bare response: {response}")
+        self.logger.info(f"{conversation} yes/no response: {response['text']}")
 
-        self.logger.info(f"{conversation} yes/no response: {clean_response}")
+        if response["text"] not in ["0", "1", "9"]:
+            raise ValueError("Returned response not 0, 1, or 9")
 
         # TODO: handle when this doesn't end up with an integer
```
bparees, line 65: this todo appears to be addressed above

tests/mock_classes/llm_chain.py
```diff
@@ -0,0 +1,18 @@
+def mock_llm_chain(retval):
+    class MockLLMChain:
+        """
+        Mock LLMChain class for testing
+
+        Example usage in a test:
+
+        monkeypatch.setattr(langchain.chains, "LLMChain",
+            MockLLMChain(desired_response='9'))
```
bparees, line 9: desired_response isn't an arg name, the argname is retval, so should this be 

```suggestion
            MockLLMChain(retval='9'))
```

instead?

bparees (commented):
couple nits, otherwise lgtm

utils/logger.py
```diff
@@ -1,4 +1,4 @@
-import logging, sys, os
+import logging, logging.handlers, sys, os
```
williamcaban, line 1: FYI. PR#36 also updates this line in addition to line 97 to fix a compatibility bug with recent versions of python and the way to invoke the file handler.

PR #36, opened 2023-12-11
(Part 1) Introduces LLMLoader class

williamcaban:
Part 1 introduces the LLMLoader class. Part 2 will connect the class to existing query_helpers and modules.

- Fixes import in logger class to fix the error with Python 3.11.6+ 
- Added requirements files for backend LLMs and used the `-r src/llms/requirements.txt` in the primary requirements file
- Updated `default.env` to reflect new backend LLM options Adding `LLMLoader` class for pluggable backend LLMs `['ollama','tgi','openai','bam','watson']`

## Type of change

- [ ] Refactor
- [X] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Create LLMLoader class without breaking changes. Will use Part 2 to connect class to existing modules and helpers to avoid breaking changes.

- Related Issue #
- Closes #

## Testing

Stand-alone test documented in Test run validation:
https://asciinema.org/a/a9wOSkb3MzePAQqoke1CPSkom?t=20

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,5 @@
 fastapi
 gradio
-ibm-generative-ai
-ibm_watson_machine_learning
 kubernetes
 langchain
 llama_index
@@ -9,3 +7,4 @@ torch
 transformers
 uvicorn
 redis
+-r src/llms/requirements.txt
--- a/src/llms/llm_loader.py
+++ b/src/llms/llm_loader.py
@@ -1,0 +1,252 @@
+# workaround to disable UserWarning
+import warnings
+warnings.simplefilter("ignore", UserWarning)
+
+from langchain.callbacks.streaming_stdout import StreamingStdOutCallbackHandler
+from langchain.callbacks.manager import CallbackManager
+
+import os, inspect
+
+## append path of current caller when __main__
+if __name__ == '__main__':
+    import sys 
+    sys.path.insert(0,'')
+
+from utils.logger import Logger
+
+class LLMLoader:
+    """
+    Note: This class loads the LLM backend librearies if the specific LLM is loaded.
+    Known caveats: Currently supports a single instance/model per backend
+    """
+    def __init__(self, llm_backends=set(),
+                 inference_url=None,
+                 prompt_type=None,
+                 api_key=None,
+                 model=None, logger=None) -> None:
+        self.logger = logger if logger is not None else Logger("llm_loader").logger
+        # a set of backend LLMs to activate
+        self.llm_backends = set(os.environ.get('LLM_BACKENDS', {'ollama'})) if len(llm_backends) == 0 else llm_backends
+        # defalt LLM backend to use or use the first index from loaded ones
+        self.llm_default = os.environ.get('LLM_DEFAULT', list(self.llm_backends)[0])
+        self.llm = {}
+        self._set_llm_instance()
+
+    def _set_llm_instance(self, inference_url=None, api_key=None, model=None):
+        self.logger.debug(f"[{inspect.stack()[0][3]}] Loading LLM instances with default {self.llm_default}")
+
+        for backend in self.llm_backends:
+            self.logger.debug(f"Loading backend LLM {backend}")
+            match backend:
+                case 'openai':
+                    self._openai_llm_instance()
+                case 'ollama':
+                    self._ollama_llm_instance()
+                case 'tgi':
+                    self._tgi_llm_instance()
+                case 'watson':
+                    self._watson_llm_instance()
+                case 'bam':
+                    self._bam_llm_instance()
+                case _:
+                    self.logger.warning(f"WARNING: Unsupported LLM {backend}")
+
+    def _openai_llm_instance(self):
+        self.logger.debug(
+            f"[{inspect.stack()[0][3]}] Creating OpenAI LLM instance")
+        try:
+            import openai
+            from langchain.llms import OpenAI 
+        except e:
+            self.logger.error(f"ERROR: Missing openai libraries. Skipping loading backend LLM.")
+            return
+        params = {
+            'base_url': os.environ.get('OPENAI_API_URL', 'https://api.openai.com/v1'),
+            'api_key': os.environ.get('OPENAI_API_KEY', None),
+            'model': os.environ.get('OPENAI_MODEL', None),
+            'model_kwargs': {}, # TODO: add model args
+            'organization': os.environ.get('OPENAI_ORGANIZATION', None),
+            'timeout': os.environ.get('OPENAI_TIMEOUT', None),
+            'cache': None,
+            'streaming': True,            
+            'temperature': 0.01,
+            'max_tokens': 512,
+            'top_p': 0.95,
+            'frequency_penalty': 1.03,
+            'verbose': False
+        }
+        self.llm['openai']=OpenAI(**params)
+        self.logger.debug(
+            f"[{inspect.stack()[0][3]}] OpenAI LLM instance {self.llm['openai']}")
+
+    def _ollama_llm_instance(self):
+        self.logger.debug(
+            f"[{inspect.stack()[0][3]}] Creating Ollama LLM instance")
+        try:
+            from langchain.llms import Ollama 
+        except e:
+            self.logger.error(f"ERROR: Missing ollama libraries. Skipping loading backend LLM.")
+            return
+        params = {
+            'base_url': os.environ.get('OLLAMA_API_URL', "http://127.0.0.1:11434"),
+            'model': os.environ.get('OLLAMA_MODEL', 'Mistral'),
+            'cache': None,
+            'temperature': 0.01,
+            'top_k': 10,
+            'top_p': 0.95,
+            'repeat_penalty': 1.03,
+            'verbose': False,
+            'callback_manager': CallbackManager([StreamingStdOutCallbackHandler()])
+        }
+        self.llm['ollama'] = Ollama(**params)
+        self.logger.debug(
+            f"[{inspect.stack()[0][3]}] Ollama LLM instance {self.llm}")
+
+    def _tgi_llm_instance(self):
+        """
+        Note: TGI does not support specifying the model, it is an instance per model.
+        """
+        self.logger.debug(
+            f"[{inspect.stack()[0][3]}] Creating Hugging Face TGI LLM instance")
+        try:
+            from langchain.llms import HuggingFaceTextGenInference
+        except e:
+            self.logger.error(f"ERROR: Missing HuggingFaceTextGenInference libraries. Skipping loading backend LLM.")
+            return
+        params = {
+            'inference_server_url': os.environ.get('TGI_API_URL', None),
+            'model_kwargs': {}, # TODO: add model args
+            'max_new_tokens': 512,
+            'cache': None,
+            'temperature': 0.01,
+            'top_k': 10,
+            'top_p': 0.95,
+            'repetition_penalty': 1.03,
+            'streaming': True,
+            'verbose': False,
+            'callback_manager': CallbackManager([StreamingStdOutCallbackHandler()])
+        }        
+        self.llm['tgi'] = HuggingFaceTextGenInference(**params)
+        self.logger.debug(
+            f"[{inspect.stack()[0][3]}] Hugging Face TGI LLM instance {self.llm}")
+
+    def _bam_llm_instance(self):
+        """BAM Research Lab"""
+        self.logger.debug(f"[{inspect.stack()[0][3]}] BAM LLM instance")
+        try:
+            # BAM Research lab
+            from genai.extensions.langchain import LangChainInterface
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,11 +1,10 @@
...
 kubernetes
 langchain
 llama_index
 torch
 transformers
 uvicorn
-redis
\ No newline at end of file
+redis
+-r src/llms/requirements.txt
```
bparees, line 10: is this the preferred approach vs just breaking up the deps into sections/groups w/ comments?

src/llms/llm_loader.py
```diff
@@ -0,0 +1,252 @@
...
+## append path of current caller when __main__
+if __name__ == '__main__':
+    import sys 
+    sys.path.insert(0,'')
+
+from utils.logger import Logger
+
+class LLMLoader:
+    """
+    Note: This class loads the LLM backend librearies if the specific LLM is loaded.
```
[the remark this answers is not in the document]
williamcaban: Fixed in update

requirements.txt
williamcaban: This maintains a dependency list of libraries only used by the class with the class for easy removal if a specific backend library is not compatible with future Python versions or is not included with the final build. Once we introduce a dependency management tool, these should move to it as well.

src/llms/llm_loader.py
```diff
@@ -0,0 +1,252 @@
...
+        except e:
+            self.logger.error(f"ERROR: Missing openai libraries. Skipping loading backend LLM.")
+            return
+        params = {
+            'base_url': os.environ.get('OPENAI_API_URL', 'https://api.openai.com/v1'),
+            'api_key': os.environ.get('OPENAI_API_KEY', None),
+            'model': os.environ.get('OPENAI_MODEL', None),
+            'model_kwargs': {}, # TODO: add model args
+            'organization': os.environ.get('OPENAI_ORGANIZATION', None),
+            'timeout': os.environ.get('OPENAI_TIMEOUT', None),
```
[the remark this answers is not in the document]
williamcaban: See if the update provides a better alignment

src/llms/llm_loader.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # using default parameters
+    llm_backend = {'backend': 'tgi'}
+
+    # Using the class
+    llm_config = LLMLoader(llm_backend=llm_backend)
+    llm_chain = LLMChain(llm=llm_config.llm, prompt=prompt)
+
+    """
+    def __init__(self, llm_backend={}, logger=None) -> None:
```
onmete, line 39: Don't set mutables (dict, list, ...) as defaults in init. https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/llms/llm_loader.py
```diff
@@ -0,0 +1,233 @@
...
+        self.logger = logger if logger is not None else Logger("llm_loader").logger
+        # LLM backend to use or fallback to default one
+        self.llm_backend = llm_backend.get('backend') if llm_backend.get('backend') is not None else os.environ.get('LLM_DEFAULT', {'backend':'ollama'})
```
onmete, line 42: Consider this setting.

```python
def __init__(self, llm_backend=None, ...):
        self.llm_backend = llm_backend or os.environ.get('LLM_DEFAULT', 'ollama')
        ....
```

If `llm_backend` is provided to constructor, its used. Otherwise its picked from env (and if env is not set, ollama is used as default). You also make a clear dependency injection available in tests by this.

src/llms/llm_loader.py
```diff
@@ -0,0 +1,233 @@
...
+        self.llm_params = llm_backend.setdefault('params',{}) # return empty dictionary if not present
+        self.llm = None
+        self._set_llm_instance()
+
+    def _set_llm_instance(self):
+        self.logger.debug(f"[{inspect.stack()[0][3]}] Loading LLM instances with default {self.llm_backend}")
+        match self.llm_backend:
+            case 'openai':
```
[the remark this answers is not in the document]
williamcaban: The purpose is to support multiple keys to avoid chaining the class to support other tunings in the future. The backend and params keys are the first two relevant today. Because of the difference in how params are split in libraries from IBM, vs other LLMs, it is the simplest way I found to account for variability and difference in attributes. The minimum required key is 'backend'.

src/llms/llm_loader.py
williamcaban: I actually had it as `llm_backend = None : dict` but reverted back last minute. Would `None` or `None : dict` preferred?

src/llms/llm_loader.py
williamcaban: ollama does not have additional dependencies (it is built in into langchain), all the other do have external libraries. Would you prefer none?

onmete: You mean `os.environ.get('LLM_DEFAULT', None)`?

src/llms/llm_loader.py
vishnuchalla: I personally prefer `None: dict` which specifies both value and type (i.e more readability). But in case we are doing it lets do it for every function.

onmete: `llm_backend: dict = None` should be fine

src/llms/llm_loader.py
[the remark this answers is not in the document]
onmete: I understand the usecase. If "backend" key is a common param key to all llms, consider adding it further down the road. Eg. where you create params. I've raised this because it seems the current code can't work as you are matching dict, but casing actual value. Maybe I'm missing something, but it should be `match self.llm_backend["backend"]:`.

src/llms/llm_loader.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    def __init__(self, llm_backend={}, logger=None) -> None:
+        self.logger = logger if logger is not None else Logger("llm_loader").logger
+        # LLM backend to use or fallback to default one
+        self.llm_backend = llm_backend.get('backend') if llm_backend.get('backend') is not None else os.environ.get('LLM_DEFAULT', {'backend':'ollama'})
+        self.llm_params = llm_backend.setdefault('params',{}) # return empty dictionary if not present
+        self.llm = None
+        self._set_llm_instance()
+
+    def _set_llm_instance(self):
```
vishnuchalla, line 47:
```suggestion
    def _set_llm_instance(self):
      self.logger.debug(f"[{inspect.stack()[0][3]}] Loading LLM instances with default {self.llm_backend}")
  
      llm_instances = {
          'openai': self._openai_llm_instance,
          'ollama': self._ollama_llm_instance,
          'tgi': self._tgi_llm_instance,
          'watson': self._watson_llm_instance,
          'bam': self._bam_llm_instance,
      }
  
      llm_instance_method = llm_instances.get(self.llm_backend)
      if llm_instance_method:
          llm_instance_method()
      else:
          self.logger.error(f"ERROR: Unsupported LLM {self.llm_backend}")
```
Can we just rewrite this logic like this? I am not much inclined towards growing ladders.

bparees: > Can we just rewrite this logic like this? I am not much inclined towards growing ladders.

this seems like a bit of a personal preference.  Is there more motivation behind why your approach is preferred?  If there are real maintainability, complexity, or bug/side-effect risks that's a different situation.

vishnuchalla: just my .02 cents on readability. Also if someone wants to add/remove an instance in future, they can just edit the`llm_instances` map

onmete: +1 for mapping instead switches

src/llms/llm_loader.py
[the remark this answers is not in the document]
bparees: self.llm_backend is not a dict, at least not alway.  The init arg is a dict, but the self.llm_backend is set here:

```
self.llm_backend = llm_backend.get('backend') if llm_backend.get('backend') is not None else os.environ.get('LLM_DEFAULT', {'backend':'ollama'})
```

which means it's either the value of the "backend" key in the `llm_backend` dict parameter argument, or it's `{'backend':'ollama'}` or whatever value the env var had.  And whether that's going to get treated by python as a dict or a string I do not know.

Clearly not good, though :)

src/llms/llm_loader.py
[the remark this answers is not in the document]
bparees: it's better in that now a caller can get an instance of a particular LLM.  I still think the configuration/parameterization leaves something to be desired, but since i'm in the middle of working on the next iteration of how we configure this thing, we can leave it be for now.

src/llms/llm_loader.py
```diff
@@ -0,0 +1,233 @@
...
+        'backend' : 'ollama',
+        'params': {'temperature': 0.02, 'top_p': 0.95},
+    }
+
+    # using default parameters
+    llm_backend = {'backend': 'tgi'}
+
+    # Using the class
+    llm_config = LLMLoader(llm_backend=llm_backend)
+    llm_chain = LLMChain(llm=llm_config.llm, prompt=prompt)
```
bparees, line 36: maybe this is a python idioms thing, but this flow still looks weird to me.

Why have LLMLoader return an llm_config and not just return an llm?  (don't make LLMLoader a class, it can just be a function that instantiates the desired llm implementation, using the passed in or loaded configuration/parameters/arguments)

The only reason i can see to return a wrapper object is if you can use that wrapper object to instantiate more llms (either of the same provider implementation but different args, or a different provider impl entirely).

as it is you seem to be returning an LLMLoader wrapper class just so someone can use the `.llm` instance it contains.  The wrapper doesn't seem to serve a purpose (maybe you have a future purpose in mind?)

src/llms/llm_loader.py
williamcaban: This is a matter of preference. While I understand the request and the increase in memory utilization by using the dictionary will be about ~1520 Bytes (with those 5 and will continue increasing as entries are added) versus a constant of ~192 Bytes when using the `match` statement, it also introduces a limitation in which, if an LLM instance requires additional parameters during invocation, it will go back to requiring `if-elseif-elseif` or going back to `match`.   Because of these reasons, I want to avoid arrays to map functions at this time.

src/llms/llm_loader.py
williamcaban: I'm okay with fallback to None if that is the preference.

src/llms/llm_loader.py
williamcaban: Okay, I'll update with this

src/llms/llm_loader.py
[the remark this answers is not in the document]
williamcaban: I will split the backend (string) and params (dict) for clarity and if no LLM definition is found, will fallback to None instead of ollama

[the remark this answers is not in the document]
williamcaban: Updated

src/llms/llm_loader.py
williamcaban: updated

src/llms/llm_loader.py
williamcaban: Updated

src/llms/llm_loader.py
williamcaban: We can return the LLM class or just do `LLMLoader(llm_backend=...).llm` but you won't be able to introspect anything about that instance of the class. For example, lets say that we want to add the capability of modifying parameters or reload the LLM object or to have a proper status function that validates the LLM backend is still operational. It all depends on preferences and of any additional methods we add to the class.

bparees: Fair enough.  I struggle a bit w/ calling it "LLMLoader" when it sounds like your future usage is more like "LLMWrapper" (wraps the native langchain llm instance w/ something that abstracts the implementation and provides a common interface for checking status or modifying config), but that's a bit of a bikeshed.

bparees (commented):
couple of nits and then lgtm.

i still have bigger picture concerns about the structure of some of this stuff, but it moves us in the right direction by enabling pluggable llms and we can work on the structure from there.

src/llms/llm_loader.py
```diff
@@ -0,0 +1,226 @@
...
+            "llm_loader").logger
+        self.llm_backend = llm_backend if llm_backend else os.environ.get(
+            'LLM_DEFAULT', None)
+        # return empty dictionary if not defined
+        self.llm_params = params if params else {}
+        self.llm = None
+        self._set_llm_instance()
+
+    def _set_llm_instance(self):
+        self.logger.debug(f"[{inspect.stack()[0][3]}] Loading LLM instances with default {str(self.llm_backend)}")
```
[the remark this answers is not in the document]
williamcaban: yeah, I need to fix that. It is a leftover from the original PR that had multiple LLMs in the same instance

bparees:
replacing with https://github.com/openshift/lightspeed-service/pull/56
/close

PR #37, opened 2023-12-12
chore: update deployment manifests

vbelouso:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

- Added use of SCC to run OLS in OpenShift
- Container image names have been replaced with direct link instead of variables

This is a preparatory work for completing the task https://issues.redhat.com/browse/OLS-50

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/manifests/ols-rbac.yaml
+++ b/manifests/ols-rbac.yaml
@@ -31,17 +31,3 @@ rules:
   resources:
   - subjectaccessreviews
   verbs: ["create"]
----
-kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: 'system:openshift:scc:anyuid'
-  namespace: ols
-subjects:
-  - kind: ServiceAccount
-    name: ols-proxy
-    namespace: ols
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: 'system:openshift:scc:anyuid'
--- a/manifests/ols-rbac.yaml
+++ b/manifests/ols-rbac.yaml
@@ -31,3 +31,17 @@ rules:
   resources:
   - subjectaccessreviews
   verbs: ["create"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: 'system:openshift:scc:anyuid'
+  namespace: ols
+subjects:
+  - kind: ServiceAccount
+    name: ols-proxy
+    namespace: ols
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: 'system:openshift:scc:anyuid'
--- a/manifests/ols-redis.yaml
+++ b/manifests/ols-redis.yaml
@@ -46,7 +46,7 @@ spec:
     spec:
       containers:
       - name: redis
-        image: ${OLS_REDIS_IMAGE}
+        image: docker.io/redis/redis-stack:latest
         ports:
         - containerPort: 6379
         - containerPort: 8001
--- a/manifests/ols-redis.yaml
+++ b/manifests/ols-redis.yaml
@@ -46,7 +46,7 @@ spec:
     spec:
       containers:
       - name: redis
-        image: docker.io/redis/redis-stack:latest
+        image: ${OLS_REDIS_IMAGE}
         ports:
         - containerPort: 6379
         - containerPort: 8001
--- a/manifests/ols.yaml
+++ b/manifests/ols.yaml
@@ -30,8 +30,6 @@ spec:
         app: ols-proxy
     spec:
       serviceAccountName: ols-proxy
-      securityContext:
-        allowPrivilegeEscalation: false
       containers:
       - name: ols-proxy
         image: quay.io/openshift/origin-kube-rbac-proxy:4.14.0
@@ -45,5 +43,9 @@ spec:
         ports:
         - containerPort: 8443
           name: https
+        securityContext:
+          allowPrivilegeEscalation: false
       - name: ols
-        image: quay.io/openshift/lightspeed-service-api:latest
+        image: ${OLS_IMAGE}
+        securityContext:
+          allowPrivilegeEscalation: false
--- a/manifests/ols.yaml
+++ b/manifests/ols.yaml
@@ -29,9 +29,9 @@ spec:
       labels:
         app: ols-proxy
     spec:
-      securityContext:
-        runAsUser: 65532
       serviceAccountName: ols-proxy
+      securityContext:
+        allowPrivilegeEscalation: false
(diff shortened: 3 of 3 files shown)


manifests/ols-redis.yaml
```diff
@@ -46,7 +46,7 @@ spec:
     spec:
       containers:
       - name: redis
-        image: ${OLS_REDIS_IMAGE}
+        image: docker.io/redis/redis-stack:latest
```
vishnuchalla, line 49: I am not sure if its safe to keep docker here. I often run into `download limit exceeded` error while pulling images from a docker account.

vbelouso: As an alternative, I can only suggest cloning Redis image somewhere to quay.io

manifests/ols-rbac.yaml
```diff
@@ -30,4 +30,18 @@ rules:
 - apiGroups: ["authorization.k8s.io"]
   resources:
   - subjectaccessreviews
-  verbs: ["create"]
\ No newline at end of file
+  verbs: ["create"]
+---
+kind: RoleBinding
```
vishnuchalla, line 35: Just curious, what is the significance of running components with `scc:anyuid`? Why can't we just leave it as before?

vbelouso: It doesn't work without SCC

```text
pods "ols-proxy-644444fb76-" is forbidden: 
unable to validate against any security context constraint: 
[provider "anyuid": Forbidden: not  usable by user or serviceaccount, 
provider "pipelines-scc": Forbidden: not usable by user or serviceaccount, 
provider restricted-v2:  .containers[0].runAsUser: Invalid value: 65532: must be in the ranges: [1000840000, 1000849999]
 ---OMITTED---
 ```

File: manifests/ols-redis.yaml (hunk 1 above)
Comment (bparees) [reply manifests/ols-redis.yaml:49]: can we use https://quay.io/repository/sclorg/redis-6-c9s ?

that's the redhat centos streams distribution of redis.

File: manifests/ols.yaml [hunk 3]
```diff
@@ -45,7 +45,5 @@ spec:
         ports:
         - containerPort: 8443
           name: https
-        securityContext:
-          allowPrivilegeEscalation: false
       - name: ols
-        image: ${OLS_IMAGE}
\ No newline at end of file
+        image: quay.io/openshift/lightspeed-service-api:latest
```
bparees, line 49: i'm not clear on your motivation to drop the use of env vars in these files.  You indicate it's preparatory work for creating the helm chart but i'm not sure why you can't make this change when you introduce the helm chart structure.

If you are determined to drop these env vars now (i don't overly care either way since i doubt anyone is really using this much today), then the readme should also be updated since it mentions using `envsubst` to deploy these resources, which has no purpose if there are no env vars to be substituted.

manifests/ols-rbac.yaml
vbelouso: @bparees Yes, I tried it on `restricted-v2`

But then the container is not created with an error
```text
Error: container create failed: time="2023-12-12T15:32:03Z" level=error msg="runc create failed: unable to start container process: exec: \"uvicorn\": executable file not found in $PATH "
```

To fully support arbitrary users, we will probably have to rewrite the Dockerfile

manifests/ols.yaml
vbelouso: Yes, I agree that it will be necessary to update Readme.
The motivation was simple - to start the service, I had to either create additional environment variables and use Kustomize or modify manifests directly.

I have the following idea -

- make working manifests so that they can be used without Helm (if someone really want to)
- make Helm and there we can configure the service as user wants

manifests/ols-rbac.yaml
bparees: that looks like an error launching the ols container, not the kube-rbac-proxy container.  We 100% control that container and can fix that issue on our side.

manifests/ols-redis.yaml
vishnuchalla: This redhat centos image has just `redis` but not other things like insights etc that we have redis-stack image. Can't we have a dedicated repo for OLS project that we can to store our images?

bparees: > Can't we have a dedicated repo for OLS project that we can to store our images in quay.io?

yes, we can, but then we need to setup something to push images to it, like we did for the OLS image.  Whether that is just mirroring an image, or it's us having our own Containerfile that extends from the base image we want and possibly does other configuration of the image.  The latter is probably the best approach because ultimately when we productize this, we're going to need our own Containerfile that lets us build the image.

So the easiest thing to do is introduce a Containerfile.redis to this repo.  Please create a "images/redis" folder structure to contain it.  Once we've landed the containerfile we can update openshift/release to build+publish the image to quay.io like we are doing with OLS.

(and then we're going to want to move the existing Containerfile under that images folder structure too, but that'll take a little more coordination to update openshift/release)

manifests/ols-rbac.yaml
bparees: @vbelouso https://github.com/openshift/lightspeed-service/pull/38 should resolve the issue with running under restricted-v2.  Let me know if you are still having issues.

manifests/ols-redis.yaml
vishnuchalla: > So the easiest thing to do is introduce a Containerfile.redis to this repo. Please create a "images/redis" folder structure to contain it. Once we've landed the containerfile we can update openshift/release to build+publish the image to quay.io like we are doing with OLS.

According to this plan, @vbelouso  can you create a "images/redis" folder and write up a Containerfile to publish our own redis image to quay.io to start with? 
```
docker.io/redis/redis-stack:latest
EXPOSE 6379
EXPOSE 8001
```

bparees: I didn't intend to make that a prereq to this PR landing.  We can live w/ the docker.io location temporarily.

manifests/ols.yaml
vbelouso: What to do in the end in this PR?
Leave variables as, Kustomize and remove SCC?

bparees: > What to do in the end in this PR?
> Leave variables as is, Kustomize and remove SCC?

yes i think that is correct (assuming that my updates to our image have fixed the issue you were hitting).

manifests/ols.yaml
```diff
@@ -29,8 +29,6 @@ spec:
       labels:
         app: ols-proxy
     spec:
-      securityContext:
-        runAsUser: 65532
```
bparees, line 33: @vbelouso so this is working now?

vbelouso: yes

PR #39, opened 2023-12-12
OLS-38: Complete work for OLS-38

williamcaban:
Case insensitive string for OLS_ENABLE_UI
https://github.com/openshift/lightspeed-service/pull/18#discussion_r1400985521

Making logfile name and size configurable
https://github.com/openshift/lightspeed-service/pull/18#discussion_r1400993505

Use logger in gradio class
https://github.com/openshift/lightspeed-service/pull/18#discussion_r1400969721

Remove hardcoded default fallback for BASE_COMPLETION_MODEL (return None if not defined) https://github.com/openshift/lightspeed-service/pull/18#discussion_r1400987617

## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [X] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/utils/logger.py
+++ b/utils/logger.py
@@ -1,4 +1,6 @@
-import logging, logging.handlers, sys, os, dotenv
+import logging, sys, os, dotenv
+from logging.handlers import RotatingFileHandler
+
 
 class Logger:
     """
@@ -61,7 +63,7 @@ class Logger:
         """
         if show_message:
             print(msg)
-        
+
         # Load the dotenv configuration in case config class has not been used
         dotenv.load_dotenv()
 
@@ -72,9 +74,9 @@ class Logger:
         if _logfile is not None and len(str(_logfile)) >= 1:
             # enforce the use of logs/ directory
             # use last token if user attemps to provide full paths
-            _logfile = "logs/"+str(_logfile).split('/')[-1]
+            _logfile = "logs/" + str(_logfile).split("/")[-1]
         self.logfile = _logfile if _logfile else logfile
-        self.logfile_maxSize = os.getenv("LOG_FILE_SIZE", (1048576 * 100)) 
+        self.logfile_maxSize = os.getenv("LOG_FILE_SIZE", (1048576 * 100))
         self.logfile_backupCount = 3
 
         self.set_handlers()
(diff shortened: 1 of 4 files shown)


utils/logger.py
```diff
@@ -60,12 +61,20 @@ def __init__(
...
         self.log_level = os.getenv("LOG_LEVEL", log_level)
         self.log_level_console = os.getenv("LOG_LEVEL_CONSOLE", self.log_level)
-        self.logfile = logfile
-        self.logfile_maxSize = 1048576 * 100
+        _logfile = os.getenv("LOG_FILE_NAME")
+        if _logfile is not None and len(str(_logfile)) >= 1:
+            # enforce the use of logs/ directory
+            # use last token if user attemps to provide full paths
+            _logfile = "logs/"+str(_logfile).split('/')[-1]
+        self.logfile = _logfile if _logfile else logfile
```
bparees, line 76: what's the actual behavior if `logfile` ends up `None`?

williamcaban: It disables logging to file (see line 105)
```python
        if self.logfile is not None:
```

bparees: Since the "logfile" init arg has a default value, the only way to actually disable logging is to pass `None` to init (and leave the env var unset)?

I think the logfile init arg should default to `None`.  Then you can enable logging by either setting the env var, or passing an arg to init.

also while you don't need to change this here, as i work on config I am going to centralize the config loading.  Having different components reading the env vars directly to determine their own config is not the best structure imho.  Loading the config in a central location makes for a better abstraction as the config format changes, and allows us to do config validation, overriding, etc, in a single spot

williamcaban: Okay. I'll address this.

williamcaban: Done.

bparees:
@williamcaban please confirm you have tested this locally and it logs properly and i will merge it.

williamcaban:
I have tested this with and without env var definition. Without env var it does not log and with `LOG_FILE_NAME=somefile.txt` and it does logs to `logs/somefile.txt`

bparees:
looks like you need to run a `make format` and repush

PR #40, opened 2023-12-13
chore: add Helm chart

vbelouso:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Added Helm chart to simplify OLS installation

https://issues.redhat.com/browse/OLS-50

## Testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@
 
 
 ### Built With
-this project is built using IBM watson machine learning 
+
+This project is built using IBM Watson machine learning.
 
 
 
@@ -22,7 +23,8 @@ To get a local copy up and running follow these simple example steps.
 ### Prerequisites
 
 
-* python 
+* Python 3.11 or newer
+* Git and pip installed
 * BAM account 
     - BAM_API_KEY
     - BAM_URL
@@ -66,12 +68,12 @@ curl -X 'POST' 'http://127.0.0.1:8080/ols' -H2 'accept: application/json' -H 'Co
 
 ### Gradio UI
 
-There is a minimal Gradio UI you can use when running the OLS server locally.  To use it, first start the OLS server per [Run the server](#run-the-server) and then browse to the built in gradio interface at http://localhost:8080/ui
+There is a minimal Gradio UI you can use when running the OLS server locally.  To use it, first start the OLS server per [Run the server](#run-the-server) and then browse to the built in Gradio interface at http://localhost:8080/ui
 
 By default this interface will ask the OLS server to retain and use your conversation history for subsequent interactions.  To disable this behavior, expand the `Additional Inputs` configuration at the bottom of the page and uncheck the `Use history` checkbox.  When not using history each message you submit to OLS will be treated independently with no context of previous interactions.
 
 ### In Cluster Deployment
-Deploying OLS on an openshift cluster is fairly easy with the configuration files we have in [manifests](./manifests) folder.
+Deploying OLS on an OpenShift cluster is fairly easy with the configuration files we have in [manifests](./manifests) folder.
 
 You can use the existing image built from the latest code via this image pullspec: quay.io/openshift/lightspeed-service-api:latest
 
@@ -87,7 +89,7 @@ Once we have our image ready, export it as an ENV and use the below [kustomize](
 export OLS_IMAGE=<image-pullspec>
 kustomize build . | envsubst | oc apply -f -
 ``` 
-This should deploy ols fronting with a [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) along with a sample [client](./config/ols-client-test.yaml) that makes requests to one of the ols endpoints demonstrating client usage of our service.
+This should deploy OLS fronting with a [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) along with a sample [client](./config/ols-client-test.yaml) that makes requests to one of the OLS endpoints demonstrating client usage of our service.
 
 <!-- ROADMAP -->
 ## Roadmap
--- a/README.md
+++ b/README.md
@@ -72,24 +72,29 @@ There is a minimal Gradio UI you can use when running the OLS server locally.  T
 
 By default this interface will ask the OLS server to retain and use your conversation history for subsequent interactions.  To disable this behavior, expand the `Additional Inputs` configuration at the bottom of the page and uncheck the `Use history` checkbox.  When not using history each message you submit to OLS will be treated independently with no context of previous interactions.
 
-### In Cluster Deployment
-Deploying OLS on an OpenShift cluster is fairly easy with the configuration files we have in [manifests](./manifests) folder.
+### Deploying OLS on OpenShift
 
-You can use the existing image built from the latest code via this image pullspec: quay.io/openshift/lightspeed-service-api:latest
+A Helm chart is available for installing the service in OpenShift.
 
-If you need to build your own image, you can use the following commands:
+Before installing the chart, you must configure the `auth.key` parameter in the [Values](helm/values.yaml) file
 
+To install the chart with the release name `ols-release` in the namespace `openshift-lightspeed`:
+
+```shell
+helm upgrade --install ols-release helm/ --create-namespace --namespace openshift-lightspeed
 ```
-podman build -f Containerfile -t=<your-image-pullspec> .
-podman push <your-image-pullspec>
+
+The command deploys the service in the default configuration.
+
+The default configuration contains OLS fronting with a [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy).
+
+To uninstall/delete the chart with the release name `ols-release`:
+
+```shell
+helm delete ols-release --namespace openshift-lightspeed
 ```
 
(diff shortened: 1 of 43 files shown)


README.md
```diff
@@ -66,24 +66,27 @@ There is a minimal Gradio UI you can use when running the OLS server locally.  T
...
+### Deploying OLS using OpenShift
 
-You can use the existing image built from the latest code via this image pullspec: quay.io/openshift/lightspeed-service-api:latest
+A Helm chart is available for installing the service in OpenShift.
 
-If you need to build your own image, you can use the following commands:
+To install the chart with the release name `ols-release` in the namespace `ols`:
 
+```shell
+helm upgrade --install ols-release helm/ --create-namespace --namespace ols
```
vishnuchalla, line 76: Do we want to specify the `ols` namespace as a resource manifest file too? just like we did with `ols-client-test` namespace.

vishnuchalla:
Tested in local. looks good to me

README.md
vbelouso: I thought about this, but probably we should give the user the opportunity to choose their own namespace.

vbelouso:
> Tested in local and I see this error in client. Looks like clusterrole and its binding in `ols` namespace are missing.
> 
> ```
> { [5 bytes data]
> * TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
> { [130 bytes data]
> * We are completely uploaded and fine
> { [5 bytes data]
> < HTTP/2 401 
> < content-type: text/plain; charset=utf-8
> < x-content-type-options: nosniff
> < content-length: 13
> < date: Thu, 14 Dec 2023 08:40:24 GMT
> < 
> { [5 bytes data]
> Unauthorized
> * Connection #0 to host ols-proxy.ols.svc left intact
> ```

Nice catch, I forgot to add role/rolebinding to ols
Thanks!

vbelouso:
@vishnuchalla regarding OLS itself
For the service to work correctly, we need access to BAM (KEY and URL)

We can keep sensitive data in secret and then mount it to OLS.
Or (at the first stage) add information to the README, how to do this manually
WDYT?

vishnuchalla:
> @vishnuchalla regarding OLS itself For the service to work correctly, we need access to BAM (KEY and URL)
> 
> We can keep sensitive data in secret and then mount it to OLS. Or (at the first stage) add information to the README, how to do this manually WDYT?

I think we need to add instructions in `README.md` to do it manually for now. I will need @bparees to share his thoughts in terms of on-premise offering.

vbelouso:
> > We can keep sensitive data in secret and then mount it to OLS.
> > Or (at the first stage) add information to the README, how to do this manually
> 
> right now OLS consumes the key from an env var. So the ideal flow would be:
> 
> 1. helm chart takes the key as a parameter
> 2. helm chart creates a secret from the parameter
> 3. ols deployment injects the env var from the secret into the ols pod
> 
> long term i am anticipating a flow more like:
> 
> 1. there exists a OLS config CRD that includes object references to LLM cred secrets
> 2. there exists an OLS config controller that watches the CRD and creates/updates the OLS deployment to mount the referenced secrets
> 3. the OLS config controller also creates an ols.config.yaml which includes the filesystem paths of the mounted LLM creds.  The ols.config.yaml gets injected into the OLS deployment either as an env var, or as a mounted configmap.
> 4. OLS app uses the ols.config.yaml to configure itself on startup
> 
> does that make sense?

I added the secret manifest.
In terms of `long-term` support - also understandable

helm/templates/client/clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ols-proxy-backend-register
```
bparees, line 4: this is the role that, when granted to a user/service account, allows that account to make OLS api requests, right?

let's give it a better name like `ols-user` or `ols-requester`

helm/templates/client/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: ols-proxy-backend-register
```
bparees, line 4: i think we should drop this binding or move it to some sort of examples folder.  it's not part of the core things that are actually needed when installing OLS

helm/templates/client/job.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: batch/v1
+kind: Job
```
bparees, line 2: i don't think it makes sense to create this as part of the helm chart.

if you want to include this in some sort of "usage examples" folder elsewhere, that's fine

helm/templates/client/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ols-client-test
```
bparees, line 4: move to examples folder or even delete entirely.

helm/templates/client/sa.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ols-client-test
+  namespace: ols-client-test
```
bparees, line 5: move to examples folder or delete entirely

helm/templates/ols/clusterrole.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: ols-proxy
```
bparees, line 4:
```suggestion
  name: ols-rbac-proxy
```

helm/templates/ols/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: ols-proxy
```
bparees, line 4:
```suggestion
  name: ols-rbac-proxy
```

helm/templates/ols/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ols-proxy
+subjects:
+- kind: ServiceAccount
+  name: ols-proxy
+  namespace: ols
```
bparees, line 12:
```suggestion
  namespace: openshift-lightspeed
```

helm/templates/ols/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+    matchLabels:
+      app: ols-proxy
+  template:
+    metadata:
+      labels:
+        app: ols-proxy
+    spec:
+      serviceAccountName: ols-proxy
+      containers:
+      - name: ols-proxy
```
bparees, line 20:
```suggestion
      - name: ols-rbac-proxy
```

helm/templates/ols/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ols-secret
```
bparees, line 4:
```suggestion
  name: ols-llm-credentials
```

helm/values.yaml
```diff
@@ -0,0 +1,36 @@
+ols:
+  replicaCount: 1
+  proxy:
```
bparees, line 3:
```suggestion
  rbac-proxy:
```

bparees (commented):
some general naming suggestions:

`proxy`/`ols-proxy` -> `rbac-proxy` when referring to the kube rbac proxy bits
`ols` -> `openshift-lightspeed` (for namespaces)
`ols`/`ols-proxy` -> `ols-api-server` (when referring to the core ols server bits)

and let's remove the test/example stuff from the chart, move it over to some tutorial/setup/example-usage folder (can be a separate PR, just take it out of the helm chart in this PR)

helm/templates/_helpers.tpl
```diff
@@ -0,0 +1,88 @@
...
+*/}}
+{{- define "ols.image" -}}
+{{- printf "%s/%s:%s" .Values.ols.api.image.registry .Values.ols.api.image.repository .Values.ols.api.image.tag | quote -}}
+{{- end -}}
+
+{{/*
+Return the proper kube-rbac-proxy image name
+*/}}
+{{- define "rbac-proxy.image" -}}
+{{- printf "%s/%s:%s" (index .Values.ols "rbac-proxy" "image" "registry") (index .Values.ols "rbac-proxy" "image" "repository") (index .Values.ols "rbac-proxy" "image" "tag") | quote -}}
```
bparees, line 87: why is the format of this construct different from the ones used for redis.image and ols.image?

vbelouso: Because of the dash character in 'rbac-proxy' in Values file. 
Because of this we get the error "bad character U+002D '-'"

bparees:
needs a rebase, otherwise lgtm

PR #52, opened 2023-12-15
Fixed multiple imports

tisnik:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

According to PEP-8, multiple imports should usually be on separate lines:
https://peps.python.org/pep-0008/#imports

## Testing
- N/A

### DIFF
--- a/src/ui/gradio_ui.py
+++ b/src/ui/gradio_ui.py
@@ -1,5 +1,6 @@
 import gradio as gr
-import requests, json
+import requests
+import json
 from utils.logger import Logger
 
 
--- a/utils/config.py
+++ b/utils/config.py
@@ -1,4 +1,5 @@
-import os, dotenv
+import os
+import dotenv
 from .logger import Logger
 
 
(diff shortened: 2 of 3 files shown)


bparees:
@tisnik @vishnuchalla   is this something we can get `black` or another formatter to format+enforce for us through our `make format` and `make verify` targets?  otherwise we'll be fighting it forever.

i'll merge this one, but if it's something we think is important we're going to want automatic formatting+enforcement or we'll never manage to stick to it.

/approve
/lgtm

tisnik:
> @tisnik @vishnuchalla is this something we can get `black` or another formatter to format+enforce for us through our `make format` and `make verify` targets? otherwise we'll be fighting it forever.
> 
@bparees IMHO it can be tested by `Ruff` at least. I can update Makefile's targets, not a problem, but not sure where to put the dependency. Now it seems that these "devel dependencies" are not specified, right?

PR #55, opened 2023-12-18
More Ruff checks

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- Added more Ruff checks
- Few fixes in the code
- Makefile updated - we accept `assert` statement in tests (but check for it in the real code)

Please note that "opinionated" rules/checks (like "D") are not yet added. Rules added are all well known and accepted in Python community.

## Testing
- `make verify`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,5 +40,5 @@ format:
 
 verify:
 	black . --check
-	ruff .
+	ruff . --per-file-ignores=tests/*:S101
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,7 @@ select = ["E", "F", "W", "C", "S"]
 
 # E501 rule checks for maximum line length
 # S113 rule checks for HTTP request being made without specifying timeout
+# TODO: verify which timeout value for HTTP request should be added (src/ui/gradio_ui.py:40), then S113 can be enabled
 ignore = ["E501", "S113"]
 
 line-length = 100
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [tool.ruff]
-select = ["E"]
+select = ["E", "F"]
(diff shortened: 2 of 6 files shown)


tisnik:
@bparees 
@vishnuchalla

Makefile
```diff
@@ -40,5 +40,5 @@ format:
 
 verify:
 	black . --check
-	ruff .
+	ruff . --per-file-ignores=tests/*:S101
```
vishnuchalla, line 43: Do we want to ignore tests files from linting?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.ruff]
-select = ["E", "F", "W", "C"]
-ignore = ["E501"]
+select = ["E", "F", "W", "C", "S"]
+ignore = ["E501", "S113"]
```
bparees, line 3: it would be nice to have some comments telling us what rules are being selected/ignored and why(so the next person who comes in and wants to change the set has some context for why we have the existing set).  or at least link to the docs.

bparees: > re E501 - imho we need to decide about the "right" max line length. I saw other PR that format for 80 columns which is (IMHO!) too small and the code is a bit ugly. OTOH now there are lines > 300 chars ;)

i would vote for at least 120.  Right now my IDE is showing about 140 chars wide.

> re S113 - we can add timeout, but I'm not sure about the right value (how fast/slow the server might be)

Note it as a TODO/followup in a comment in the file, that way at least we know why it's being ignored for now

tisnik: added, and I'll find the author (@vishnuchalla probably) to talk about the timeout

PR #57, opened 2023-12-19
Use timeout for HTTP POST method to avoid infinite waits

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Use timeout for HTTP POST method to avoid infinite waits

## Testing
- Right now it is not covered by integration tests, unfortunately

### DIFF
--- a/src/ui/gradio_ui.py
+++ b/src/ui/gradio_ui.py
@@ -37,7 +37,9 @@ class gradioUI:
 
         try:
             # Make the HTTP POST request, wait for response with 10 seconds timeout
-            response = requests.post(self.ols_url, headers=headers, data=json_data, timeout=10)
+            response = requests.post(
+                self.ols_url, headers=headers, data=json_data, timeout=10
+            )
 
             # Check if the request was successful (status code 200)
             if response.status_code == 200:
--- a/src/ui/gradio_ui.py
+++ b/src/ui/gradio_ui.py
@@ -36,9 +36,9 @@ class gradioUI:
         json_data = json.dumps(data)
 
         try:
-            # Make the HTTP POST request, wait for response with 10 seconds timeout
+            # Make the HTTP POST request, wait for response with 30 seconds timeout
             response = requests.post(
-                self.ols_url, headers=headers, data=json_data, timeout=10
+                self.ols_url, headers=headers, data=json_data, timeout=30
             )
 
             # Check if the request was successful (status code 200)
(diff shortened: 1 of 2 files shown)


tisnik:
@bparees

tisnik:
@vishnuchalla

src/ui/gradio_ui.py
```diff
@@ -49,7 +51,7 @@ def chat_ui(self, prompt, history, use_history=None):
                     f"Request failed with status code {response.status_code}"
                 )
                 self.logger.info(f"Response text: {response.text}")
-                return "Sorry, an error occurred: " + response.text
+                return f"Sorry, an error occurred: {response.text}"
```
bparees, line 54: I don't particularly care either way, but is there a particular reason to prefer this style of string construction?

src/ui/gradio_ui.py
```diff
@@ -36,8 +36,10 @@ def chat_ui(self, prompt, history, use_history=None):
         json_data = json.dumps(data)
 
         try:
-            # Make the HTTP POST request
-            response = requests.post(self.ols_url, headers=headers, data=json_data)
+            # Make the HTTP POST request, wait for response with 10 seconds timeout
+            response = requests.post(
+                self.ols_url, headers=headers, data=json_data, timeout=10
```
bparees, line 41: removing my lgtm.  This needs to be higher.  The LLM generation calls can be pretty slow.  Make it 30s.

src/ui/gradio_ui.py
tisnik: well Python people find it more idiomatic. TBH I don't care either (returning to Python from Go world), but IMHO we should be consistent

bparees: > IMHO we should be consistent

yeah, i'm also coming from a golang world so i miss the consistent structures the language itself enforces.

I was just curious if there was an actual performance difference or anything.  I'm fine w/ us trying to standardize on use of f-string.

tisnik: no, no performance improvements - Python is still slow, as usual :-)  [and Python 3.11/3.12 is a just bit better: some benchmarks I did https://www.root.cz/obrazek/1110447/]

src/ui/gradio_ui.py
tisnik: okay

tisnik: updated

PR #59, opened 2023-12-19
Catch JSON deserialization error for payload reading

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

During response payload retrieval and deserialization, any JSON deserialization exception can be thrown. We don't know/care which JSON library is used (it can be `simplejson` or `json`), but luckily for us, the ancestor of such exceptions is `ValueError`. So all we have to, is to catch this one.

## Testing
- Integration tests are not available right now


tisnik:
@vishnuchalla

tisnik:
@bparees

PR #61, opened 2023-12-19
Initial pass at abstracting the configuration into a central location and class.

bparees:
This will allow us to replace the current use of env vars to configure OLS, with
a more formal config yaml.

It also gives us a central location to worry about loading the config and validating
it, as well as defining defaults.

Some notes to reviewers:
The overall structure of the config models is not really the focus here, we will continue iterating on that as we evolve our configuration.  The focus should be on abstracting the config loading process, and making the various endpoints dependent on that central config.  Also the leveraging of the LLMLoader abstraction so that all points where we call an LLM can now use any LLM that the config defines.

## Type of change

- [x ] Refactor
- [x ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing


bparees:
Self merging based on Erik's slack feedback and desire not to have people blocked behind this pr or conflicting with it.

We'll keep iterating on it but I believe it moves us in a good and necessary direction

PR #65, opened 2023-12-20
Use sorted imports + enable more linters

tisnik:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Description

- Use sorted imports everywhere
- More linters can be enabled right now

## Testing
- See CI jobs

### DIFF
--- a/app/models/models.py
+++ b/app/models/models.py
@@ -1,5 +1,6 @@
-from pydantic import BaseModel
 from typing import Union
+
+from pydantic import BaseModel
 
 
 class LLMRequest(BaseModel):
--- a/src/docs/docs_summarizer.py
+++ b/src/docs/docs_summarizer.py
@@ -1,11 +1,13 @@
 import os
+
 import llama_index
-from src import constants
 from dotenv import load_dotenv
 from llama_index import StorageContext, load_index_from_storage
 from llama_index.prompts import PromptTemplate
-from utils.model_context import get_watsonx_context
+
+from src import constants
 from utils.logger import Logger
+from utils.model_context import get_watsonx_context
 
 load_dotenv()
 
--- a/app/endpoints/feedback.py
+++ b/app/endpoints/feedback.py
@@ -1,4 +1,5 @@
 from fastapi import APIRouter
+
 from app.models.models import FeedbackRequest
 from utils.logger import Logger
 
--- a/app/endpoints/ols.py
+++ b/app/endpoints/ols.py
@@ -1,15 +1,17 @@
 import os
+
 from fastapi import APIRouter, HTTPException
+
 from app import constants
 from app.models.models import LLMRequest
 from app.utils import Utils
-from utils.model_context import get_watsonx_predictor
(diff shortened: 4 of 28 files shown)


app/models/models.py
```diff
@@ -1,6 +1,7 @@
-from pydantic import BaseModel
 from typing import Union
 
+from pydantic import BaseModel
```
thoraxe, line 3: t comes after p -- this isn't sorted?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.ruff]
 
-select = ["E", "F", "W", "C", "S"]
+select = ["E", "F", "W", "C", "Q", "SLF", "I"]
```
thoraxe, line 3: What does this line do?

src/docs/docs_summarizer.py
```diff
@@ -1,11 +1,13 @@
 import os
+
 import llama_index
```
thoraxe, line 3: why is this still here (l) if we are alpha sorting, and then you left the other llama-related a few lines down?

thoraxe (changes requested):
I don't fully understand this PR. In some cases, it appears that you are not alpha sorting, but rather some other ordering comes first and then, within that ordering, you are alpha sorting.

Is there some PEP that describes the correct way to sort and group imports?

app/models/models.py
tisnik: it sorts, but after empty line, basically the new "import group" is sorted on its own. It's a bit similar like in Go

pyproject.toml
tisnik: It enables more linters. All of them can be listed by using `ruff linter`. Q - check string quotes, SLF - self. param check, I - import sorting

tisnik:
> I don't fully understand this PR. In some cases, it appears that you are not alpha sorting, but rather some other ordering comes first and then, within that ordering, you are alpha sorting.
> 
> Is there some PEP that describes the correct way to sort and group imports?

Ideal behavior is:

1) sort default packages
2) sort third party packages
3) sort in-project packages

src/docs/docs_summarizer.py
tisnik: please look below:
Ideal behavior is:

    "future" imports first
    sort default packages
    sort third party packages
    sort in-project packages

pyproject.toml
thoraxe: Can you make the linter change a separate PR? I want Ben and others to discuss that.

src/docs/docs_summarizer.py
thoraxe: :+1:

pyproject.toml
tisnik: yes sure, I'll delete this commit ASAP

tisnik: done

PR #66, opened 2023-12-20
[WIP] implements SSAR-based authentication check

thoraxe:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This PR is currently a science experiment that implements an in-cluster authentication check using SelfSubjectAccessReview and FastAPI middleware.

* It assumes that a `nonResourceURL` `clusterRole` has been created outside of OLS, a'la:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: ols-clusterrole
rules:
- nonResourceURLs: ["/ols"]
  verbs: ["get"]
  ```
* it assumes that a user or group has been given this cluster role
* it adds a requirement that a request hitting FastAPI/OLS must include an authorization header that is their Kubernetes API token
* it whitelists any of our default routes, like status checks, and only performs auth checks for everything else
* it supports globally disabling auth checks

This probably needs a lot of work before it could be merged, but it potentially replaces what we wanted to do with kube-rbac-proxy.

N/A

## Testing
No tests, yet.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,4 +1,7 @@
-from fastapi import FastAPI, Request
+import json
+import requests
+
+from fastapi import FastAPI, Request, JSONResponse
 
 from app.endpoints import feedback, ols
 from src.ui.gradio_ui import gradioUI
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/main.py [hunk 1]
```diff
@@ -15,6 +18,103 @@
...
+    configuration.verify_ssl = False
+
+    # create a k8s client using the supplied authentication information
+    k8s_client = kubernetes.client.ApiClient(configuration)
+
+    # create a self subject access review to determine if the user can access the lightspeed service
+    # TODO: this is currently checking a non-resource attribute path of `/ols` which is totally generic.
+    #       in the future we might want to restrict users to specific ols endpoints even and make this much
+    #       fancier
+    jsonstr = '{"apiVersion":"authorization.k8s.io/v1","kind":"SelfSubjectAccessReview","spec":{"nonResourceAttributes":{"path":"/ols","verb":"get"}}}'
```
bparees, line 101: rather than us passing the user's token on to k8s, i think we should be determining the user's identity (from the token) and then doing a subject access review in which we ask k8s "can user X do Y"?

but this is an area where we probably need to get some advice from the auth team.

thoraxe: My question is simply: Why?

The user passed us their token. Why would we first determine who they are, and then ask if they can do a thing? We already are them via their key. That's why this is a SelfSubjectAccessReview -- because we are that user (via their key) asking if *I* can do the thing.

Reversing the discussion - once we determine the user can do the thing, then what?  The purpose of using their key was so that later/in the future, we can interact with the API at their permission level to do things. For example, we use their API key to decide if they can use OLS, and then, if they can, we use their API key to maybe modify an object in the cluster. 

The user had access to OLS, but did not have write/edit access on the resource in question, so they were blocked.

If we didn't use their key, we'd end up having to do accessreview requests for pretty much everything, which seems wasteful/inefficient.

:shrug:

bparees: > My question is simply: Why?

Passing the user token around feels like an unnecessary risk.  The token is very powerful.

The ideal flow in my mind would be more like a traditional oauth flow in which OLS just gets a token that only has enough scope/permission to verify the user's identity, but can't do anything else.

but more fundamentally us defining auth flows w/ talking to the auth experts worries me.

thoraxe: I agree that we should be concerned about what we could do with the user's token.

I agree that we should talk to someone about how we could get more granular, but understanding how granular we need to get requires us to have a clear picture of what we want to be able to do on behalf of the user. Right now the scope is "whatever the user could do, OLS should be able to do on their behalf."

Remember, at some point we want the user to be able to do things like "can you tell me which pods are crash looping?"

This requires a lookup. It sounds like you are saying you don't want OLS to do that. Or, if you want OLS to do that, you want a system that _temporarily_ gets enough permissions to do _only_ that.

?

bparees: > This requires a lookup. It sounds like you are saying you don't want OLS to do that. Or, if you want OLS to do that, you want a system that temporarily gets enough permissions to do only that.

I definitely don't want to give it permission to do that until we are committed to it doing that and have fully thought through the approach.  So i'm focused on "what's the right way to do the thing we know we need to do now" and trying to avoid  "let's open a potential sudo hole that will allow us to do things we might want to do in the future"

And if someone can come back and say "yeah, using the user token for this is fine", then i'm ok with it.  I believe it is what the console does, so it's not unprecedented.  I just want us to do the exercise of confirming the approach.

thoraxe: ACK

PR #69, opened 2024-01-02
Verify project sources with Python 3.11 expected idioms

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Description

Linters are able to detect some old and now non-standard patterns in the code. In order to do this its need to know the target Python version. So let's start with Python 3.11 as documented.

## Testing
- Will be done on CI
- You can use `make verify` as usual


onmete:
LGTM

PR #70, opened 2024-01-02
Help in Makefile shown using `make help` phony target

tisnik:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

## Description

Help in Makefile shown using `make help` phony target:

```$ make help
Usage: make <OPTIONS> ... <TARGETS>

Available targets are:

images                    Build container images
install-deps              Install all required dependencies needed to run the service
install-deps-test         Install all required dependencies needed to test the service
run                       Run the service locally
test                      Run all tests
test-unit                 Run the unit tests
test-integration          Run integration tests tests
format                    Format the code into unified format
verify                    Verify the code using various linters
help                      Show this help screen
```

## Testing
- Just use `make help` from CLI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,39 +6,39 @@
 ARTIFACT_DIR := $(if $(ARTIFACT_DIR),$(ARTIFACT_DIR),tests/test_results)
 
 
-images:
+images: ## Build container images
 	scripts/build-container.sh
 
-install-deps: 
+install-deps: ## Install all required dependencies needed to run the service
 	pip install -r requirements.txt
 
-install-deps-test:
+install-deps-test: ## Install all required dependencies needed to test the service
 	pip install -r requirements-test.txt
 
-run:
+run: ## Run the service locally
 	uvicorn app.main:app --reload --port 8080
 
-test: test-unit test-integration test-e2e
+test: test-unit test-integration test-e2e ## Run all tests
 
-test-unit:
+test-unit: ## Run the unit tests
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,44 +1,53 @@
 # Put targets here if there is a risk that a target name might conflict with a filename.
 # this list is probably overkill right now.
 # See: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
-.PHONY: test test-unit test-e2e images run format verify
+.PHONY: test test-unit test-e2e images run format verify help
```
thoraxe, line 4: do we think we'll have a file called `help` ?

tisnik: very probably not :)

thoraxe: so then I think you should remove help from `.PHONY` since that is the purpose (filename conflict)... no?

thoraxe (changes requested):

Makefile
tisnik: fixed!

thoraxe:
Is it normal/best-practice to use double hashes and include the description of the make target "inline"?

tisnik:
> Is it normal/best-practice to use double hashes and include the description of the make target "inline"?

It is commonly used idiom described in several articles https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html

PR #72, opened 2024-01-02
Update contrib install instructions

onmete:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [X] Documentation Update

## Description

Updated contributing documentation and pre-commit dependencies.

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,9 @@ pip install --upgrade pip
 # install project dependencies
 make install-deps
 
+# install dev/tests dependencies
+make install-deps-test
+
 # code formatting (run this as a pre-commit step for your code changes)
 make format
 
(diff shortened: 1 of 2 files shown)


tisnik (approved):
LGTM

thoraxe (changes requested):
As this is a markdown file, things with `#` are header sections. I'm not sure why you would uncapitalize titles (Clone -> clone).

I can see un-capitalizing "virtual environment" but "Python" is a proper noun.

But, in spite of all that, as these are section headings, I wonder if they shouldn't appropriately be title case and not sentence case:
https://titlecaseconverter.com/blog/sentence-case-vs-title-case/

But "no capitals-case" is probably not what we want here.

@bparees WDYT?

onmete:
@thoraxe Those are inline code comments, not markdown titles.

bparees:
/ok-to-test
/lgtm
/approve

> But "no capitals-case" is probably not what we want here.

since they are just incline script comments, all lower case is fine.    As it was would have been fine too imho, so i hope not to see too much of people making personal preference style edits(we need to focus on more material concerns), but since this PR also updates some steps that were legitimately wrong/out of date, i'm ok with it in this case.

@onmete if you join the openshift github org, we will no longer have to explicilt approve testing for your PRs

PR #76, opened 2024-01-02
Prevent double call of `file.close()`

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Description

Prevent double call of `file.close()`
Actually Python ignores this call for already closed file, i.e. exception is not raised. But it is not correct to call close two times.

## Testing
- N/A, not covered by tests

### DIFF
--- a/scripts/transform_coverage_report.py
+++ b/scripts/transform_coverage_report.py
@@ -64,7 +64,6 @@ if __name__ == "__main__":
     try:
         with open(output_file_path, "w") as output_file:
             output_file.write("mode: set\n")
-            pass
 
         with open(json_file_path, "r") as json_file:
             json_content = json_file.read()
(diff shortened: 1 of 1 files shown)


thoraxe:
@tisnik are you saying we never need to use `.close`? Because now there are _no_ `.close` statements.

tisnik:
> @tisnik are you saying we never need to use `.close`? Because now there are _no_ `.close` statements.

all operations on files are performed within the `with` block. And when the control escapes this block, the `file.close()` is called automatically - that's basically the purpose of using `with` block with files.

thoraxe:
OK, makes sense to me that we don't need to call close then if the `with open` behavior in Python does that automagically.

/approve

PR #78, opened 2024-01-03
Imports statements are sorted + added new linter rule

tisnik:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Description

- Imports statements are sorted according to common Python standards (see below)
- Added new linter rule, so `make verify` will check for this problem

### Details

This rule is derived from the **isort** linter.

## What it does
De-duplicates, groups, and sorts imports based on the provided `isort` settings.

## Why is this bad?
Consistency is good. Use a common convention for imports to make your code
more readable and idiomatic.

## Example
```python
import pandas
import numpy as np
```

Use instead:
```python
import numpy as np
import pandas
```

## Testing
- To be performed on CI

### DIFF
--- a/app/endpoints/ols.py
+++ b/app/endpoints/ols.py
@@ -1,14 +1,14 @@
 from fastapi import APIRouter, HTTPException
 
+from app import constants
 from app.models.models import LLMRequest
 from app.utils import Utils
 from src.docs.docs_summarizer import DocsSummarizer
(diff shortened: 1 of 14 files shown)


thoraxe:
The linter change should probably be in its own PR.

That being said -
/approve

PR #81, opened 2024-01-03
Added more types for functions called from unit tests

tisnik:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Description

Added more types for functions called from unit tests (and thus tested properly)

## Testing
- To be checked on CI

### DIFF
--- a/src/query_helpers/question_validator.py
+++ b/src/query_helpers/question_validator.py
@@ -18,7 +18,9 @@ class QuestionValidator:
         """
         self.logger = Logger("question_validator").logger
 
-    def validate_question(self, conversation: str, query: str, verbose: bool = False) -> list[str]:
+    def validate_question(
+        self, conversation: str, query: str, verbose: bool = False
+    ) -> list[str]:
         """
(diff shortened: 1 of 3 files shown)


src/query_helpers/question_validator.py
```diff
@@ -18,7 +18,7 @@ def __init__(self):
         """
         self.logger = Logger("question_validator").logger
 
-    def validate_question(self, conversation, query, **kwargs):
+    def validate_question(self, conversation: str, query: str, **kwargs) -> list[str]:
```
thoraxe, line 21: Does this return a list of strs or just a single str?

bparees: list/array: https://github.com/openshift/lightspeed-service/pull/81/files#diff-9ca3ebc421243362c6b59edda5939229ef8cd3c3cd4c83753986de3b315d3092L66

PR #82, opened 2024-01-04
Ability to export coverage report into interactive HTML

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Description

- New Makefile target to export code coverage report into interactive HTML
- It can then be viewed locally
- No need to export the resulting HTML

## Testing
- Run `make coverage-report` on CLI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ test-integration: ## Run integration tests tests
 test-e2e: ## Run e2e tests
 	# Command to run e2e tests goes here
 
+coverage-report:	test-unit ## Export unit test coverage report into interactive HTML
+	coverage html
+
 format: ## Format the code into unified format
 	black .
 
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ logs/*
 
 # ignore test results
 tests/test_results/*
+htmlcov/
 .coverage
 
 # ignore cache directories


Makefile
```diff
@@ -35,6 +35,9 @@ test-integration:
 test-e2e:
 	# Command to run e2e tests goes here
 
+coverage-report:	test-unit
+	coverage html
```
bparees, line 39: where does this expect to pick up the coverage report from and does it align w/ where test-unit will put them(especially since test-unit picks the output dir from an env variable)?  I'd prefer to see this be explicit.

tisnik: I really think that coverage database is still created in project root directory. It does not depend on `ARTIFACT_DIR` AFAIK, but let me check

tisnik: yes, the `ARTIFACT_DIR` does not change where the coverage database is stored.

bparees: looks like `.coverage` is already in our .gitignore, but can you add whatever file(s) this generates?

tisnik: yes, sure, nice idea

tisnik: @bparees updated

PR #83, opened 2024-01-04
Ignore cache directories - these should not be pushed into the repo

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Description

Ignore cache directories - these should not be pushed into the repo

## Testing
- Run any linter or unit test and the call `git status`. It should not list those "dot directories" even if "dot files" are included in search path in Git settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,8 @@ logs/*
 # ignore test results
 tests/test_results/*
 .coverage
+
+# ignore cache directories
+.mypy_cache
+.pytest_cache
+.ruff_cache


onmete:
Speaking about pushing to repo, what about the `/logs` dir? Does it make sense to push an empty folder?

bparees:
/approve

> Speaking about pushing to repo, what about the /logs dir? Does it make sense to push an empty folder?

probably not, but it's slightly convenient since by default that's where OLS logs to when you run it locally (at least last i recall).  If/when we change that default behavior, it might make sense to drop the dir, but until then it's a convenience that it already exists.

Not sure if @thoraxe had something else in mind when he pushed the dir:
https://github.com/openshift/lightspeed-service/commit/454a168635e1647f3efbd08dfc4bd9284c3fc8da#diff-dc9938ca8f545b1e38dc16f33a4fa29042b3fe7256e990461814ac72190b13f8

PR #86, opened 2024-01-05
Fixed type related errors

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Description

There are some errors found during type checking (NPE can occurs etc.). So let's fix them.

- Related Issue [OLS-83](https://issues.redhat.com//browse/OLS-83)

## Testing
- N/A right now

### DIFF
--- a/app/endpoints/ols.py
+++ b/app/endpoints/ols.py
@@ -98,9 +98,12 @@ def ols_request(llm_request: LLMRequest):
             # Further processing of YAML response (filtering, cleaning, linting, RAG, etc.)
 
             llm_response.response = wrapper + "\n" + generated_yaml
-            config.conversation_cache.insert_or_append(
-                conversation, llm_request.query + "\n\n" + llm_response.response
-            )
+
+            if config.conversation_cache is not None:
+                config.conversation_cache.insert_or_append(
+                    conversation,
+                    llm_request.query + "\n\n" + str(llm_response.response or ""),
+                )
             return llm_response
 
         else:
--- a/src/indexer/indexer.py
+++ b/src/indexer/indexer.py
@@ -13,6 +13,8 @@ llama_index.set_global_handler("simple")
 load_dotenv()
 
 # Select Model
+embed_model: TextEmbeddingsInference | str
+
 ## check if we are using remote embeddings via env
 url = os.getenv("TEI_SERVER_URL", "local")
 if url != "local":
--- a/src/llms/llm_loader.py
+++ b/src/llms/llm_loader.py
@@ -1,6 +1,7 @@
 import inspect
 import os
 import warnings
+from typing import Optional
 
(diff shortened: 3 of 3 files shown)


tests/unit/utils/test_json_tools.py
```diff
@@ -1,8 +1,10 @@
+from typing import Any
+
 import pytest
 
 from utils.json_tools import parse_and_check_json_markdown, parse_json_markdown
 
-correct_inputs = (
+correct_inputs: tuple[tuple[str, dict[str, Any]], ...] = (
```
onmete, line 7: Is it necessary to have typing in tests?

app/endpoints/ols.py
```diff
@@ -98,9 +98,12 @@ def ols_request(llm_request: LLMRequest):
...
             llm_response.response = wrapper + "\n" + generated_yaml
-            config.conversation_cache.insert_or_append(
-                conversation, llm_request.query + "\n\n" + llm_response.response
-            )
+
+            if config.conversation_cache is not None:
+                config.conversation_cache.insert_or_append(
+                    conversation,
+                    llm_request.query + "\n\n" + str(llm_response.response or ""),
+                )
```
onmete, line 106: I might be missing some code context, but if this was changed by the tool, I'm not entirely sure if it's a good move. If `config.conversation_cache` is `None`, and it shouldn't be, we are missing a raise telling us that there is something wrong.

tests/unit/utils/test_json_tools.py
tisnik: not strictly necessary

app/endpoints/ols.py
tisnik: it is done by hand. Tooling can't do these types of changes right now.

tisnik: I inferred, that cache might not be configured and the service would still work (just w/o having conversation history).

tests/unit/utils/test_json_tools.py
onmete: so if you remove it, would linting fail?

app/endpoints/ols.py
onmete: makes sense

tests/unit/utils/test_json_tools.py
tisnik: no, if we setup it correctly - can be done after we change project structure

tisnik: removed as agreed upon with @onmete

thoraxe (changes requested):
Please squash and force push

PR #87, opened 2024-01-05
Integer values should be really converted into integer

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Description

- Integer values should be really converted into integers
- It will make the failure on the right place (during init+setup), not later

## Testing
- Try to set incorrect env. variable value and see...

### DIFF
--- a/utils/config.py
+++ b/utils/config.py
@@ -31,7 +31,7 @@ def load_config_from_env() -> None:
     logger_config = config_model.LoggerConfig()
     ols_config.logger_config = logger_config
 
-    logger_config.default_level = int(os.getenv("LOG_LEVEL", logging.INFO))
+    logger_config.default_level = os.getenv("LOG_LEVEL", logging.INFO)
     logger_config.default_filename = os.getenv("LOG_FILE_NAME", None)
     logger_config.default_size = int(os.getenv("LOG_FILE_SIZE", (1048576 * 100)))
     default_logger = Logger(
--- a/utils/config.py
+++ b/utils/config.py
@@ -31,9 +31,9 @@ def load_config_from_env() -> None:
     logger_config = config_model.LoggerConfig()
     ols_config.logger_config = logger_config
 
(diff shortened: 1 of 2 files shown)


utils/config.py
```diff
@@ -31,9 +31,9 @@ def load_config_from_env() -> None:
     logger_config = config_model.LoggerConfig()
     ols_config.logger_config = logger_config
 
-    logger_config.default_level = os.getenv("LOG_LEVEL", logging.INFO)
+    logger_config.default_level = int(os.getenv("LOG_LEVEL", logging.INFO))
```
onmete, line 34: There is `LOG_LEVEL=INFO` in `default.env`. That wouldn't work with this. You can set the logging level as string or int, it can work with both, so make sure we choose just one way of setting it. For me, it's a bit more visible to have debug/info/,... instead of numeric values.

tisnik: ok in this case we need to update it somewhere else, yes. Lemme do it

tisnik: fixed @onmete

thoraxe (commented):
This changes several things from strings to ints, but doesn't change anywhere these values are used. Does the thing that is consuming the redis port expect a string or an int? If it works before this PR, wouldn't this PR break it?

tisnik:
> This changes several things from strings to ints, but doesn't change anywhere these values are used. Does the thing that is consuming the redis port expect a string or an int? If it works before this PR, wouldn't this PR break it?

Default redis port is an integer. And it should be an integer.

thoraxe:
I dug into the code -- you're right in that `StrictRedis` takes an integer for the port, but we are still pulling from environ in our cache itself:
https://github.com/openshift/lightspeed-service/blob/main/src/cache/redis_cache.py#L43-L47

I will open an issue for that, since it would likely blow up if someone set the env wrong.

PR #96, opened 2024-01-08
Line length for Ruff

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Description

It was decided to use max 79 characters on line
https://docs.google.com/document/d/1M5PfCFV6TehAgVklUhAEdB7pGMXbI_2c1DFAYoj2bfo

So it changes the Ruff settings.

## Testing
- N/A right now


pyproject.toml
```diff
@@ -6,5 +6,5 @@ select = ["E", "F", "W", "C", "S", "I", "TCH"]
 ignore = ["E501"]
 
 target-version = "py311"
-line-length = 100
+line-length = 79
```
bparees, line 9: This seems excessively short.  I don't even have a widescreen monitor and i get 140 chars on my editor of choice(vscode), with a navigation tab on the left.  And even when lines wrap, i have never found that to be a huge problem.

What is the motivation to set this so low?

tisnik: Well we had a mtg. just before refinement meeting. Basically 79 is standard in Python (PEP-8). Personally I voted for 100 chars at least, but I was overvoted (is it the right word?)

bparees: >  I was overvoted (is it the right word?)
i think you mean "outvoted"

tisnik: that there were more votes for 79 limit (4 votes AFAIK)

bparees: I listened to the recording, though i couldn't see the votes.  I didn't really hear strong support for 79 char as a limit, it was just kinda "well, we don't know and the PEP says 79"(but also "the PEP dates from a different time and there's just no community inertia to update it to something more modern").  To the extent i heard people expressing an actual personal preference, it seemed to be for longer line lengths (100 was the lowest i heard proposed outside of 79)

for now https://github.com/openshift/lightspeed-service/pull/98 introduced a limit of 168 (reflecting the longest lines we currently have) and actually enforces it (so we don't get new lines that are longer than that)

If someone feels passionately that 168 is too long, i'd be open to reducing it, but not to 79.  I think 120 is about as low as i'd willingly go(and i'd be perfectly happy with 140 or even just leaving the 168 cap in place) barring a really convincing argument about why 120 is a problem.

Whoever feels that passionate about it is also going to have to fix the existing lines that exceed the new limit (i would really rather not mark them as exceptions to the linter.  that feels like admitting that it doesn't really matter that much, in which case we shouldn't be enforcing it)

imho the most compelling argument for constraining line length is that wrapped lines can be more difficult to read in an editor and it a github diff.  But forcibly broken up lines can also be harder to read, so i don't think one approach or the other is ever going to make everyone happy in all situations.

onmete: 168 or 140 is way too much, with this proposal, you are doing exactly what you pointed out. You are pushing your personal preference.

My personal preference is to have two windows in the editor, so such long lines will force me to either develop a new habit or scream every time I need to scroll horizontally. :)

Also, GitHub diff is a valid point. If GH is forcing 120 and we are too, we have much better control over how the long line is split and ensuring it is readable. You don't have this control with GH as it wraps it just "somehow".

100 would be a good compromise. WDYT?

bparees: > 168 or 140 is way too much, with this proposal, you are doing exactly what you pointed out. You are pushing your personal preference.

I didn't say having a personal preference is bad, i said i didn't hear anyone expressing a real preference for 79 char limits on the call.  Everyone else seemed to just kinda go along with it because it's what the PEP says despite people admitting the PEP was outdated and not really relevant.  And i don't want to see the team default into a choice that no one really wants just because they didn't care enough to speak up, which is why i'm reraising the topic here.

This is all absolutely personal preference (unlike, say, picking a dep manager where there are technical implications associated w/ the choice)

> My personal preference is to have two windows in the editor, so such long lines will force me to either develop a new habit or scream every time I need to scroll horizontally. :)

I don't really want to question your workflow/tooling choices, but since that seems to be the primary reason you are pushing for this..... why are you scrolling horizontally?  your editor doesn't wrap lines for you instead of scrolling?

bparees:
@tisnik is your intent to wait for other PRs to merge that fix the existing lines and then merge this once that is done?

if so, please put a "/hold" on it to make it clear what it's waiting for so we can ignore it in the meantime

bparees:
looks like we're going to land this change via https://github.com/openshift/lightspeed-service/pull/154

/close

PR #104, opened 2024-01-09
100% unit test coverage for logging module

tisnik:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

## Description

100% unit test coverage for logging module

- Related Issue #[OLS-187](https://issues.redhat.com//browse/OLS-187)

## Testing
- To be tested on CI

### DIFF
--- a/tests/unit/utils/test_logger.py
+++ b/tests/unit/utils/test_logger.py
@@ -1,0 +1,77 @@
+import logging
+import os
+from tempfile import gettempdir
+
+from utils.logger import Logger
+
+# log file should be stored in temporary directory
+log_file_name = os.path.join(gettempdir(), "test.log")
+
+
+def remove_logfile():
+    """Remove logfile, if exists."""
+    if os.path.exists(log_file_name):
+        os.remove(log_file_name)
+
+
+def setup_function(function):
+    """Setup test environment for this module."""
+    remove_logfile()
+
+
+def teardown_function(function):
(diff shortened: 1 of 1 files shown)


tests/unit/utils/test_logger.py
```diff
@@ -0,0 +1,75 @@
+import logging
+import os
+
+from utils.logger import Logger
+
+log_file_name = "test.log"
```
bparees, line 6: use a path in TMPDIR, e.g.: https://docs.python.org/3/library/tempfile.html

bparees:
/approve

lgtm pending update to use a tmpfile for the test logfile output.

tests/unit/utils/test_logger.py
bparees: (the motivation being we don't want to dump random files into the users working directory, even if we successfully clean up it could be stomping an existing file, and if we don't clean up it's littering the working directory)

tisnik: ah yes, that is good idea, indeed.

tisnik: updated

tisnik: @bparees ^

PR #105, opened 2024-01-09
Idiomatic checks in model config checking

tisnik:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Description

Idiomatic checks in model config checking

## Testing
- N/A

### DIFF
--- a/app/models/config.py
+++ b/app/models/config.py
@@ -35,10 +35,10 @@ class ProviderConfig(BaseModel):
         self.name = data.get("name", None)
         self.url = data.get("url", None)
         self.credential_path = data.get("credential_path", None)
-        if data.get("models", None) is None or len(data["models"]) == 0:
+        if "models" not in data or len(data["models"]) == 0:
             raise Exception(f"no models configured for provider {data['name']}")
         for m in data["models"]:
-            if m.get("name", None) is None:
+            if "name" not in m:
                 raise Exception("model name is missing")
             model = ModelConfig(m)
             self.models[m["name"]] = model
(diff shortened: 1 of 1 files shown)


app/models/config.py
```diff
@@ -99,11 +99,11 @@ def __init__(self, data: Optional[dict] = None):
...
-            if data.get("redis", None) is None:
+            if "redis" not in data:
                 raise Exception("redis config is missing")
-            self.redis = RedisConfig(data.get("redis", None))
+            self.redis = RedisConfig(data["redis"])
         elif self.type == "in-memory":
-            if data.get("in-memory", None) is None:
+            if "in-memory" not in data:
                 raise Exception("in-memory config is missing")
             self.memory = MemoryConfig(data.get("memory", None))
```
tisnik, line 108: @bparees here I'm not sure if we really want to access `data["memory"]` or `data["in-memory"]`

bparees: `memory` is correct here.

here is an example of the config section:
```
  conversation_cache:
    type: in-memory
    memory:
      max_entries: 1000
```

you can change `in-memory` to `memory` and be consistent everywhere, i think @vishnuchalla persuaded me to use `in-memory` at some point but i'm starting to doubt that choice, i think `memory` is sufficiently clear relative to the other option (some sort of persistent db storage)

cc @syedriko

bparees: i'll merge this as is and leave it up to @syedriko if he wants to change `in-memory` to `memory` in places where we currently use `in-memory`

PR #114, opened 2024-01-11
Update docstrings to not mention type when there is a type anotation

onmete:
…in the object

- Wrong docstring format fixed at some places.

## Type of change

- [X] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Description

If we are using type annotations in the arguments, we don't need to put the types again in the docstrings. See this post https://stackoverflow.com/a/67378139.

- Related Issue #
- Closes #

## Testing


tisnik (approved):
LGTM

ols/utils/logger.py
```diff
@@ -82,11 +82,6 @@ def __init__(
         self.set_handlers()
 
     def set_handlers(self) -> None:
-        """A simple logger function that logs messages at a specified level.
```
tisnik, line 85: why docstring was removed?

tisnik (approved):

ols/utils/logger.py
onmete: The docstring is about log messages, but the method is about handlers. Seems misplaced.

bparees: shouldn't we fix the docstring then, rather than remove it?

bparees: > Isn't the name of this method self-explanatory? 

yes...but if we're putting linting rules in that enforce a docstring requirement, there won't be a choice.  (I haven't closely followed all the rules you guys have proposed, but i was under the impression that was one of them)

> There is a different effort about writing docstrings. Could this be eventually solved in another PR?

i mean, sure....but why delete the existing one only to make someone else rewrite it correctly later(assuming it will be a requirement to have one) when you're already touching it?

if there is no planned requirement to require docstrings then i'm fine w/ deleting this because i agree it's unnecessary tedium for such a simple method.  The problem w/ not having a doc requirement and expecting people to doc the things that need it, is it quickly devolves into nothing getting docs.  (conversely, however, the problem w/ mandating docstrings is people will write useless docstrings that tell you nothing more than what the function name already tells you, just to meet the requirement.  But at least it makes people think about it)

PR #129, opened 2024-01-12
OLS-155: Add explicit env for logging level when logging to file

onmete:
## Description

Added explicit env for logging level when logging to file.

## Type of change

- [X] Refactor

### DIFF
--- a/default.env
+++ b/default.env
@@ -12,7 +12,10 @@ OLS_ENABLE_UI=True
 # LOG_FILE_SIZE = <size in bytes>
 
 LOG_LEVEL=INFO
-LOG_LEVEL_CONSOLE=INFO
+# you can use envs bellow to influence logging level depending on the
+# handler/output - those are optional, if not set LOG_LEVEL will be used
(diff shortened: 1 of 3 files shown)


onmete:
This leads to the question if we need logging to file. @bparees mentioned it here https://redhat-internal.slack.com/archives/C068JAU4Y0P/p1704753176067979?thread_ts=1704277142.129889&cid=C068JAU4Y0P

Do we want to store feedback via logging mechanism or something else?

bparees:
this is fine, but i'd like to remind you that we're replacing the env based config w/ yaml based config(and replacing direct env loading by components with a shared config object that is loaded in a single spot), so a better use of time would be to wire up the config objects (even if the config objects load from env right now) to the loghandler rather than adding more code to the loghandler that's directly loading env variables.

Something to think about in general (I appreciate that it's hard to look at the codebase today and understand what is "meaningful code that's worth refactoring/improving" and what is "throwaway code that's going to be replaced/rewritten and isn't worth investing in")

/approve
/lgtm

PR #131, opened 2024-01-12
OLS-204: logging refactoring

onmete:
- This unifies app logging. You can't have different logging configurations within a single app instance now (eg. some parts of app having a debug level while others have info).
- Third-party libs logging level can be changed.
- Renamed config to logging_config
- Logging config is initiated after config is loaded.
- I've added the logger name to the msg format. You can easily distinguish app and third-party loggers.
    Eg.
    ```
    2024-01-25 12:51:48,880 [httpcore.http11:_trace.py:45] DEBUG: send_request_header...
    2024-01-25 13:55:12,045 [ols.src.ui.gradio_ui:gradio_ui.py:34] INFO: Using histo...
    ```
- All application logging is done via module level `logger = logging.getLogger(__name__)` now.
	
Given the number of changes, I can't guarantee I've updated everything, but the configuration seems to be doing what it is supposed to do, and /the query endpoint works as usual, so that gives me at least some level of confidence.


olsconfig.yaml
```diff
@@ -11,12 +11,15 @@ LLMProviders:
     models:
       - name: gpt-4-1106-preview
 OLSConfig:
-  enable_debug_ui: false
+  enable_debug_ui: true
```
onmete, line 14: my mistake, this should stay false

ols/utils/logging.py
```diff
@@ -0,0 +1,18 @@
...
+def configure_logging(logging_config: LoggingConfig) -> None:
+    """Configure application logging according to the configuration."""
+    logging.basicConfig(
+        level=logging_config.app_log_level,
+        format="%(asctime)s [%(name)s:%(filename)s:%(lineno)d] %(levelname)s: %(message)s",
+    )
+    # manage logging levels for external libraries
+    logging.getLogger("httpcore").setLevel(logging_config.httpcore_log_level)
+    logging.getLogger("httpx").setLevel(logging_config.httpx_log_level)
+    logging.getLogger("urllib3").setLevel(logging_config.urllib3_log_level)
```
onmete, line 18: we can do this dynamically too (setting some pattern for config object), but now there are just these three (at least these three are doing some logging)

ols/app/models/config.py
```diff
@@ -271,27 +271,62 @@ def validate(self) -> None:
...
+    urllib3_log_level: Optional[str] = None
 
     def __init__(self, data: Optional[dict] = None):
         """Initialize configuration and perform basic validation."""
         super().__init__()
         if data is None:
             return
-        level = logging.getLevelName(data.get("default_level", "INFO"))
-        if level is None:
+        app_log_level = logging.getLevelName(data.get("app_log_level", "info").upper())
```
tisnik, line 287: when it will be None?

ols/app/models/config.py
```diff
@@ -271,27 +271,62 @@ def validate(self) -> None:
...
-                f"invalid log level for default log: {data.get('default_level',None)}"
+                f"invalid log level for default log: {data.get('app_log_level')}"
             )
-        self.default_level = level
+        self.app_log_level = app_log_level
+
+        httpcore_log_level = logging.getLevelName(
+            data.get("httpcore_log_level", "warning").upper()
+        )
+        if httpcore_log_level is None:
```
tisnik, line 297: dtto, can it be None?

tisnik (commented):
some comments Ondro

ols/app/main.py
```diff
@@ -15,11 +17,18 @@
                   """,
 )
 
+# load configuration
```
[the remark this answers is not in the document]
onmete: I wanted a friend for `# configure logging`. :)
I can remove both comments.

ols/app/models/config.py
onmete: Maybe some value for `getLevelName` exists that would return `None`. But most likely, it will be never `None`. I didn't want to change the actual validation here. Seems there is still some room for improvement. But I can, if you think it belongs here.

ols/app/models/config.py
onmete: The same response as above.

ols/app/models/config.py
onmete: Yea, this would fail when you provide `app_log_level: false`.

olsconfig.yaml
```diff
@@ -11,12 +11,15 @@ LLMProviders:
...
     type: memory
     memory:
       max_entries: 1000
-  logger_config:
-    default_level: INFO
+  logging_config:
+    app_log_level: info
+    httpcore_log_level: warning
+    httpx_log_level: warning
+    urllib3_log_level: warning
```
bparees, line 23: this is ok to add to the config yaml, but we should not be exposing these fields on the OLS CRD, it's too much of an internal implementation detail.

perhaps "app_log_level" and "library_log_level" or something.  (Yes that means we'd use the same loglevel for all the libraries)

bparees:
/approve

@tisnik can finish iterating on the review+lgtm

tisnik:
> /approve
> 
> @tisnik can finish iterating on the review+lgtm

on it

ols/app/models/config.py
```diff
@@ -271,27 +271,62 @@ def validate(self) -> None:
...
+        if isinstance(other, LoggingConfig):
+            return (
+                self.app_log_level == other.app_log_level
+                and self.httpcore_log_level == other.httpcore_log_level
+                and self.httpx_log_level == other.httpx_log_level
+                and self.urllib3_log_level == other.urllib3_log_level
+            )
         return False
 
     def validate(self) -> None:
```
syedriko, line 332: @onmete We've renamed the validate() -> validate_yaml() or it clashed with a base class method, in https://github.com/openshift/lightspeed-service/pull/240. Just a heads up.

olsconfig.yaml
syedriko: - Can we have one global log level, too?
- Why [just] these packages? Can we make the list of packages dynamic?
- Can we please expose a knob for log level(s) through the API so it can be changed at runtime? It would be invaluable for tracking down production issues once it's in the field

bparees: > Can we please expose a knob for log level(s) through the API so it can be changed at runtime? It would be invaluable for tracking down production issues once it's in the field

i don't think that this is needed, certainly not right now anyway.  OLS itself is stateless, so i can't think of a scenario in which we'd need to increase the loglevel and would lose critical debugging state if OLS was restarted to change the loglevel.  Do you have a scenario in mind?

I appreciate the idea and i'm not going to claim it would never be useful, but adding something like that means we need to stop and design the interface for changing the log level (who can change it? is it protected by auth?  is it controlled by the operator or CRD somehow?  does that mean new logic in the operator too?) as well as (ideally) test scenarios for it to ensure it functions as desired.  I just don't see the benefits being worth that cost right now

To my knowledge none of the core OCP payload binaries support changing loglevel dynamically other than maybe some networking level pieces.  I won't claim we've never wished we had it, but it's also never been a significant obstacle to supporting our users.

syedriko: I've mostly been thinking about tracking down connectivity and authentication/authorization issues, such as this thing is not talking to that thing and it's not giving a clear reason why, so the CX folks will likely need to ask the CU to enable the highest level of logging for the networking packages/libraries all the way down to OpenSSL, see what the issue is and then restore the normal level of logging. Changing the log level dynamically would start capturing debug data right when the issue is happening. If a restart is needed, it's a less clean situation where something can change due to the restart and the issue would go away for a while or it would take extra effort to reproduce the situation again. But I agree it's not the first priority.

onmete: I'll do a single setting `library_log_level` now. Once there is a need to further break it down, its easy to change.

ols/app/main.py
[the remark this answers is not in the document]
onmete: removed

ols/app/models/config.py
onmete: refactored

tisnik:
It looks ok + coverage is still 100%. I'm ok :)

/lgtm

PR #139, opened 2024-01-15
OLS-83: Specify optional types where it is needed

tisnik:
## Description

It is now required (PEP-484) that all optional types need to be specified explicitly.
Additionally, log level can be specified as int or as string
This PR fixes some errors reported by mypy/Pyright

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Testing
- Use mypy or Pyright

### DIFF
--- a/ols/utils/logger.py
+++ b/ols/utils/logger.py
@@ -41,8 +41,8 @@ class Logger:
     def __init__(
         self,
         logger_name: str = "default",
-        log_level: int = logging.INFO,
-        logfile: str = None,
+        log_level: str = logging.getLevelName(logging.INFO),
+        logfile: str | None = None,
         show_message: bool = False,
     ):
         """Initializes the Logger instance.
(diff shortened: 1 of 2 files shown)


ols/utils/logger.py
```diff
@@ -41,8 +41,8 @@ class Logger:
     def __init__(
         self,
         logger_name: str = "default",
-        log_level: int = logging.INFO,
-        logfile: str = None,
+        log_level: int | str = logging.INFO,
```
bparees, line 44: this should not be either/or.  

if we are representing/passing the loglevel as a string in some places and an int in others, let's straighten that out now.

I'd expect that users will provide the loglevel as a string (e.g. `info`) and then we will validate that they've provided as a valid loglevel that the loghandler understands (possibly by converting it to an int and then using that int value consistently internally)

tisnik: well the standard logging module supports both. We can indeed remove integer one as it is not user-friendly

bparees: yeah let's just support string from a user-facing/config perspective.

tisnik: but ok, can change it

tisnik: updated @bparees

PR #141, opened 2024-01-15
OSL-83: No need for optional config values - just use default values

tisnik:
## Description

There's no need for optional config values - just use default values. It seems like o "local" change, but strong type checkers would require explicit tests for `is not None` everywhere the values are used. Which is innapropriate as we know, that these values will be set to default ones later.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #OLS-83

## Testing

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -82,8 +82,8 @@ class RedisConfig(BaseModel):
 
     host: Optional[str] = None
     port: Optional[int] = None
-    max_memory: Optional[str] = None
-    max_memory_policy: Optional[str] = None
+    max_memory: str = constants.REDIS_CACHE_MAX_MEMORY
+    max_memory_policy: str = constants.REDIS_CACHE_MAX_MEMORY_POLICY
 
     def __init__(self, data: Optional[dict] = None):
         """Initialize configuration and perform basic validation."""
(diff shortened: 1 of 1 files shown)


onmete:
It's set again to the same value in the init, so it would deserve some refactoring, but I assume this partial module will be re-done with https://issues.redhat.com/browse/OLS-89

But if linter complains, let's proceed with this.

/lgtm

bparees:
> There's no need for optional config values - just use default values. 

the reason to leave them empty is that if you're not using "Redis" as the caching solution, then it doesn't make sense for the config to contain values for the Redis configuration.

Right now it's possible to understand whether Redis should be used or not due to the "Cache type" field, but it's still not ideal imho for the runtime config to contain values/config that are irrelevant to how the system is running (at best it may lead to confusion if someone dumps the config).

In fact a good validation would be to throw an error if any of the Redis config values are populated when the cache type is not Redis (and vice versa for the memory cache)

tisnik:
> > There's no need for optional config values - just use default values.
> 
> the reason to leave them empty is that if you're not using "Redis" as the caching solution, then it doesn't make sense for the config to contain values for the Redis configuration.
> 
> Right now it's possible to understand whether Redis should be used or not due to the "Cache type" field, but it's still not ideal imho for the runtime config to contain values/config that are irrelevant to how the system is running (at best it may lead to confusion if someone dumps the config).
> 
> In fact a good validation would be to throw an error if any of the Redis config values are populated when the cache type is not Redis (and vice versa for the memory cache)

so I guess we can close this one?

bparees:
> so I guess we can close this one?

i guess so.  We should do defaulting in the config loading logic itself where we can make intelligent choices about whether a particular field should have a default value or not (and in some cases possibly what that default should be)

/close

PR #143, opened 2024-01-15
Check if provider is supported

tisnik:
## Description

- Fix logging message template - avoid key error
- Unit test checks that key error is not thrown
- Program logic will probably need to be updated, but not in this simple bug fix

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit test

### DIFF
--- a/tests/unit/llms/test_llm_loader.py
+++ b/tests/unit/llms/test_llm_loader.py
@@ -1,0 +1,21 @@
+"""Unit tests for LLMLoader class."""
+
+import pytest
+
+from ols.src import constants
+from ols.src.llms.llm_loader import LLMLoader, UnsupportedProvider
+
+
+def test_constructor_no_provider():
+    """Test that constructor checks for provider."""
+    # we use bare Exception in the code, so need to check
+    # message, at least
+    with pytest.raises(Exception, match="ERROR: Missing provider"):
+        LLMLoader(provider=None)
+
+
+def test_constructor_wrong_provider():
+    """Test how wrong provider is checked."""
+    # currently, it just logs error and that's it
+    with pytest.raises(UnsupportedProvider):
+        LLMLoader(provider="invalid-provider", model=constants.GRANITE_13B_CHAT_V1)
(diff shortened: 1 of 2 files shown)


tests/unit/llms/test_llm_loader.py
```diff
@@ -0,0 +1,20 @@
...
+    # we use bare Exception in the code, so need to check
+    # message, at least
+    with pytest.raises(Exception, match="ERROR: Missing provider"):
+        LLMLoader(provider=None)
+
+
+def test_constructor_wrong_provider():
+    """Test how wrong provider is checked."""
+    # currently, it just logs error and that's it
+    LLMLoader(provider="foo-bar-baz", model=constants.GRANITE_13B_CHAT_V1)
```
bparees, line 20:
```suggestion
    LLMLoader(provider="invalid-provider", model=constants.GRANITE_13B_CHAT_V1)
```

makes it clearer what the expected result is.

bparees: Also for something like this:

1) separate PR from fixing the message template
2) would it be that hard to add error handling/exception raising in the llm_loader for this so we can have a meaningful test?

tisnik: fixed the provider name @bparees 

Ok so is the expected behavior to raise an exception immediately the unknown provider is ... provided?

tisnik: okay make sense, I'm on it

tisnik: better now @bparees ? :)

bparees:
/approve

fix the PR title and commit message to reflect that this is actually changing and i'll lgtm

tisnik:
> /approve
> 
> fix the PR title and commit message to reflect that this is actually changing and i'll lgtm

Fixed. TY in advance

PR #148, opened 2024-01-16
OLS-155: Merge constants under single project level module

onmete:
## Description

Having two different constants.py under app and src makes the codebase a bit confusing. This PR merges both into `ols/constants.py` and fixes imports.

## Type of change

- [X] Refactor

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -1,5 +1,12 @@
 """Constants used in business logic."""
 
+VALID = "VALID"
+INVALID = "INVALID"
+YAML = "YAML"
+NOYAML = "NOYAML"
+SOME_FAILURE = "some failure"
+
+
 # templates
 SUMMARIZATION_TEMPLATE = """
 The following context contains several pieces of documentation. Please summarize the context for the user.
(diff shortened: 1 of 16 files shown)


ols/constants.py
```diff
@@ -1,5 +1,12 @@
 """Constants used in business logic."""
```
bparees, line 1: seems odd to have this at the top level as opposed to under src or utils, are you following a best practice recommendation here?

onmete: You mean having the `ols/constants.py`? I don't think there is a best practice in terms of saying "this" should be always at "some" location.

Constants are not utils, nor specific to src (which we want to eventually want to get rid of). Constants are project-wide. Hence, they should be in the project root (`ols/`). Having multiple `constants.py` per specific directory is an option too, but I don't think it would be a better approach than a single file (it's better to have unique module names).

bparees:
/lgtm
/approve

this doesn't reflect a general approval of putting more code in the top-level/root `ols` dir, but it's ok for constants.

PR #149, opened 2024-01-16
OLS-155: Use constants in prompt template

onmete:
## Description

We have constants, let's use them.

Merge after https://github.com/openshift/lightspeed-service/pull/148

## Type of change

- [X] Refactor

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -90,40 +90,41 @@ Statement: {statement}
 Response:
 """
 
-QUESTION_VALIDATOR_PROMPT_TEMPLATE = """
+# {{query}} is escaped because it will be replaced as a parameter at time of use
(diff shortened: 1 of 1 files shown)


ols/constants.py
```diff
@@ -83,40 +90,42 @@
 Response:
 """
 
-QUESTION_VALIDATOR_PROMPT_TEMPLATE = """
+# we are escaping the curly braces there so that we can use the format
+# function to fill query later
```
bparees, line 94: I don't understand this comment.....aren't the now-bracketed values just immediately getting substituted by the constant values?  (which is fine, it's a good change to use the constants, but i don't understand the part about using the format function later)

bparees: @onmete ^^^

PR #150, opened 2024-01-16
Update PR template

onmete:
## Description

Up for consideration.

We are using Jira issues in the PR name. With that, "Related Tickets & Documents" seems unnecessary. When you need to add more links, you still can put those in the description, but given the atomic nature of PR's, it shouldn't happen that often.

## Type of change

- [X] Documentation Update

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,11 +13,6 @@
 - [ ] Configuration Update
 - [ ] Bump-up dependent library
 
-## Related Tickets & Documents
-
-- Related Issue #
-- Closes #
-
 ## Checklist before requesting a review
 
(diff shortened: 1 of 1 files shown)


onmete:
Might break existing GH to Jira workflow. Closing.

bparees:
> Might break existing GH to Jira workflow. Closing.

@onmete I don't think it would have broken anything.  And in fact I would prefer that the team standardize on tracking issues in jira (meaning that if someone opens a GH issue we should turn it into a jira(Bug or story) and close the github issue w/ a link to the jira ticket.

And then jira tickets should be referenced in the PR title.

so if we do all those things, there should be no need to ref issues(jira or GH) in the PR description body.

cc @joshuawilson for your thoughts on issue tracking and mapping between GH + Jira.

PR #153, opened 2024-01-16
OLS-155: Remove commented code

onmete:
## Description

There is a bunch of commented code. Seems like left-overs from developing.

## Type of change

- [X] Refactor

### DIFF
--- a/ols/app/main.py
+++ b/ols/app/main.py
@@ -8,7 +8,6 @@ from ols.utils import config
 
 app = FastAPI()
 
-# config = load_config(os.environ.get("OLS_CONFIG_FILE","olsconfig.yaml"))
 config.load_config_from_env()
 if config.ols_config.enable_debug_ui:
     app = gradioUI(logger=config.default_logger).mount_ui(app)
--- a/ols/utils/config.py
+++ b/ols/utils/config.py
@@ -124,30 +124,3 @@ def load_config_from_env() -> None:
             m.name = model
             oai_provider.models[m.name] = m
         llm_config.providers[oai_provider.name] = oai_provider
-
-
-# def load_config(config_file):
-#     try:
-#         with open(config_file, "r") as f:
-#             data = yaml.safe_load(f)
-#             c=Config(data)
-#     except Exception as e:
-#         print(f"Failed to load config file {config_file}: {str(e)}")
-#         print(traceback.format_exc())
-#         exit(1)
-#     try:
-#         c.validate()
-#     except Exception as e:
-#         print(f"Failed to validate config file {config_file}: {str(e)}")
-#         print(traceback.format_exc())
-#         exit(1)
-#     global config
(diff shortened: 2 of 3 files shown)


ols/app/main.py
```diff
@@ -8,7 +8,6 @@
 
 app = FastAPI()
 
-# config = load_config(os.environ.get("OLS_CONFIG_FILE","olsconfig.yaml"))
```
bparees, line 11: this isn't a leftover, it's a future thing we'll need when we shift to config.yaml based configuration.  (Well we won't necessarily need exactly this, we'll probably pass the config filename as an argument rather than loading it as an env var, but that's something for @syedriko to work out/propose).

for now it's ok to remove this, but just calling it out for @syedriko's awareness.

ols/utils/config.py
```diff
@@ -124,30 +124,3 @@ def load_config_from_env() -> None:
...
-#     global config
-#     global default_logger
-#     global feedback_logger
-#     global conversation_cache
-#     config = c
-#     default_logger = Logger(logger_name="default", logfile=c.ols_config.logger_config.default_filename,log_level=c.ols_config.logger_config.default_level).logger
-#     feedback_logger = Logger(logger_name="feedback", logfile=c.ols_config.logger_config.feedback_filename,log_level=c.ols_config.logger_config.feedback_level).logger
-
-#     conversation_cache = CacheFactory.conversation_cache(c.ols_config.conversation_cache)
-#     return c
```
bparees, line 153: this is also related to the future implementation of the yaml config loading.  @syedriko may want to start fresh, but i think it's better if we let him remove this as part of his work to introduce the config.yaml loading officially.

bparees (commented):
i think we should leave the config related commented code in place and let @syedriko remove it as part of his work.

i'd approve the cache test change if it was separate

ols/utils/config.py
onmete: Is necessary to have it in code though? It is more like a snippet that should be part of the Jira story.

bparees: it's not, but it's there now and it'll be removed shortly (this sprint) by the work Sergey is doing, so rather than remove it(and make him dig it up from git history for reference) or moving it elsewhere, let's just let him remove it as part of the work he's already doing, when he's ready.

bparees: (The reason it's here is this is code i wrote when testing the config work i was doing, and knowing that it would become relevant again when we were ready to adopt my config work, i left it in place but commented out for the future time when we'd want it again, or at least want to use it as a reference.  I'm not recommending it as a standard practice)

onmete: I still think a snippet is a better option here, leaving Sergey already a clean slate for work. But ok. I'll close this and open a new one with the test one (as its oneliner, I'm too lazy to revert and force push here:)).

PR #157, opened 2024-01-16
OLS-83: Loglevel as name everywhere

tisnik:
## Description

Loglevel as name everywhere

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/utils/config.py
+++ b/ols/utils/config.py
@@ -35,7 +35,9 @@ def load_config_from_env() -> None:
     logger_config = config_model.LoggerConfig()
     ols_config.logger_config = logger_config
 
-    logger_config.default_level = os.getenv("LOG_LEVEL", logging.INFO)
+    logger_config.default_level = os.getenv(
+        "LOG_LEVEL", logging.getLevelName(logging.INFO)
+    )
     default_logger = Logger(
         logger_name="default",
         log_level=logger_config.default_level,


ols/utils/config.py
```diff
@@ -35,7 +35,9 @@ def load_config_from_env() -> None:
     logger_config = config_model.LoggerConfig()
     ols_config.logger_config = logger_config
 
-    logger_config.default_level = os.getenv("LOG_LEVEL", logging.INFO)
+    logger_config.default_level = os.getenv(
+        "LOG_LEVEL", logging.getLevelName(logging.INFO)
```
onmete, line 39: What about just `os.getenv("LOG_LEVEL", "info")`? Would that work too?

The function actually returns different things based on the input. I guess those are all valid values, but its funny.
image

tisnik: just bare "info" will introduce yet another anonymous magic constant :/

onmete: Ok.

bparees: it's surprising that logging doesn't define its own string constants for "INFO", "DEBUG", etc..but i guess they only chose to define numeric constants.

Anyway @tisnik i think you've made the right call here to avoid us defining another constant.  This ties the behavior/constants to the core logging library expectations

PR #164, opened 2024-01-16
Use LLMChain for base prompt llm invocation

bparees:
## Description

LLM loader returns a `ChatOpenAI` instance(derived from `BaseChatModel`) for openai, whereas for BAM the loader returns a `LangChainInterface` (derived from `LLM`), and those two base classes don't have the same invocation arguments(one takes the prompt in the form of a raw string, the other expects a List of BaseMessages), 

This meant we weren't able to use the result of llmloader directly(it would fail when switching to openai, the code worked for the LLM base class but not for the BaseChatModel class).  

All the rest of our code seems to wrap the result of llmloader into a `LLMChain` which seems to abstract the argument difference, so i've updated the base_prompt implementation to do the same.....but this is going to be something to keep an eye on (using LLM vs BaseChatModel interfaces in OLS, and whether there are other differences under the covers even when wrapping them in a LLMChain, or other behaviors of LLMChain that we do not necessarily want)

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/tests/integration/test_ols.py
+++ b/tests/integration/test_ols.py
@@ -1,12 +1,14 @@
 """Integration tests for basic OLS REST API endpoints."""
 
+from unittest.mock import patch
+
 import requests
 from fastapi.testclient import TestClient
 
 from ols import constants
-from ols.app.endpoints import ols
 from ols.app.main import app
 from ols.src.query_helpers.question_validator import QuestionValidator
+from tests.mock_classes.llm_chain import mock_llm_chain
 
 client = TestClient(app)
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_ols.py
```diff
@@ -45,19 +47,10 @@ def test_status() -> None:
     }
 
 
+# the raw prompt should just return stuff from LLMChain, so mock that base method in ols.py
+@patch("ols.app.endpoints.ols.LLMChain", new=mock_llm_chain("test response"))
 def test_raw_prompt(monkeypatch) -> None:
```
onmete, line 52: You can remove `monkeypatch` here. It's not used anymore.

thoraxe:
/approve

but do we still want this endpoint? It was mostly a hack/test

tests/integration/test_ols.py
bparees: good catch, we'll get it cleaned up in a follow up

PR #166, opened 2024-01-17
OLS-89: Update OLS to consume config yaml

syedriko:
## Description

Implemented loading of OLS configuration from a YAML file.
- The location of the configuration file is taken from the OLS_CONFIG_FILE env variable and defaults to olsconfig.yaml in the current directory.
- An example of the config file is in olsconfig.yaml
- Providers and models carry the `credentials` and `credential_path` attributes. `credentials` takes precedence. If `credentials` is not set, `credential_path` it taken to contain the name of the file to load the credential from. It is then assigned to the `credentials` attribute.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # https://issues.redhat.com/projects/OLS/issues/OLS-89
- Closes #

## Testing


ols/utils/config.py
```diff
@@ -23,130 +23,36 @@ def load_empty_config() -> None:
     llm_config = config_model.LLMConfig()
 
 
-def load_config_from_env() -> None:
-    """Load configuration from environment variables."""
+def load_config_from_stream(stream):
+    """Load configuration from a stream."""
     global ols_config
```
tisnik, line 28: IMHO it might be better to try not to use global variables if possible. But it might need some refactoring there

ols/utils/config.py
```diff
@@ -23,130 +23,36 @@ def load_empty_config() -> None:
...
-#     global config
-#     global default_logger
-#     global feedback_logger
-#     global conversation_cache
-#     config = c
-#     default_logger = Logger(logger_name="default", logfile=c.ols_config.logger_config.default_filename,log_level=c.ols_config.logger_config.default_level).logger
-#     feedback_logger = Logger(logger_name="feedback", logfile=c.ols_config.logger_config.feedback_filename,log_level=c.ols_config.logger_config.feedback_level).logger
+def load_config(config_file):
+    """Load configuration from a file."""
+    global config
```
tisnik, line 49: dtto, it uses global var and returns it at the same time, IMHO that's confusing

ols/app/models/config.py
```diff
@@ -211,22 +256,18 @@ def __init__(self, data: Optional[dict] = None):
         super().__init__()
         if data is None:
             return
-        self.llm_config = LLMConfig(data.get("llm_providers", None))
-        self.ols_config = OLSConfig(data.get("ols_config", None))
+        v = data.get("LLMProviders")
+        if v is not None:
+            self.llm_config = LLMConfig(v)
```
raptorsun, line 261: 1. Shall we keep the key using underscore as delimiter as other models do, for a better consistency? (`llm_providers` vs `LLMProviders`)
2. I think it more readable in this way:
```python
if "LLMProviders" in data:
    self.llm_config = LLMConfig(data["LLMProviders"])
```

ols/utils/config.py
raptorsun: It seems that this test does not require the variable ols_config, llm_config, conversation_cache and default_logger to be global to make it working.

ols/app/models/config.py
bparees: > Shall we keep the key using underscore as delimiter as other models do, for a better consistency?

yes, when i wrote the initial example config.yaml i went with python-style naming, i think we should stick with that.  So "llm_providers" and not "LLMProviders", as the field name in the yaml.

ols/utils/config.py
bparees: i agree w/ the sentiment but in this case it's effectively read-only(config is loaded at startup and then nothing should be changing it) so i don't think it's a big problem or that passing it around buys us much.

but it's definitely not a pattern we should be repeating in general.

ols/utils/config.py
bparees: on this point i agree.  Can probably just call it "init_config" and have it initialize the global var w/o returning it.

tests/unit/test_config.py
```diff
@@ -0,0 +1,170 @@
...
+        """foobar""", AttributeError, "'str' object has no attribute 'get'"
+    )
+    check_expected_exception(
+        """123""", AttributeError, "'int' object has no attribute 'get'"
+    )
+    check_expected_exception(
+        """12.3""", AttributeError, "'float' object has no attribute 'get'"
+    )
+    check_expected_exception(
+        """[1,2,3]""", AttributeError, "'list' object has no attribute 'get'"
```
bparees, line 46: this seems like it's just testing the yaml parser and confirming that the config loader doesn't mask/wrap any parsing exceptions so i question the value of having so many of these tests.  To be honest we probably *should* catch+wrap parsing exceptions....we still want to show the user the parse exception, but having the config loader catch+wrap it so we can give them a sane error like `failed to load config due to yaml parse error: $original_parse_error`

tests/unit/test_config.py
```diff
@@ -0,0 +1,170 @@
...
+        p1 = config.config.llm_config.providers["p1"]
+        assert p1.name == "p1"
+        assert p1.url == "url1"
+        assert p1.credential_path == "secret.txt"
+        assert p1.credentials == "secret_key"
+        assert len(p1.models) == 2
+        m1 = p1.models["m1"]
+        assert m1.name == "m1"
+        assert m1.url == "murl1"
+        assert m1.credentials == "mkey1"
```
bparees, line 167: if we're going to have a test like this it should test all the fields of the config (so the ols config fields too, not just the llm config ones)

tests/unit/test_config.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+    check_expected_exception(
+        """12.3""", AttributeError, "'float' object has no attribute 'get'"
+    )
+    check_expected_exception(
+        """[1,2,3]""", AttributeError, "'list' object has no attribute 'get'"
+    )
+
+
+def test_invalid_config() -> None:
```
bparees, line 50: some additional invalid config scenarios:

`xxxx_model` field(e.g. default_model) doesn't refer to a valid provider/model combination

redis configuration supplied when cache type is memory
memory config supplied when cache type is redis

we should probably also go ahead and validate that any credential_paths that are provided, can be read.  (the creds may not be valid, but at least ensure that the path the user provided exists and is readable so we can fail early if it's not)

tests/unit/test_config.py
```diff
@@ -0,0 +1,170 @@
...
+        assert False
+
+
+def test_valid_config_file() -> None:
+    """Check if a valid configuration file is handled correctly."""
+    try:
+        config.load_config("olsconfig.yaml")
+        # it would be great to roundtrip the config back to YAML and diff the two YAMLs
+        assert len(config.config.llm_config.providers) == 2
+        p1 = config.config.llm_config.providers["p1"]
```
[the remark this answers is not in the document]
syedriko: Thanks for the idea! You must've meant __eq__ though.

tests/unit/test_config.py
```diff
@@ -0,0 +1,170 @@
...
+
+E = TypeVar("E", bound=Exception)
+
+
+def check_expected_exception(
+    yaml_stream: str, expected_exception_type: Type[E], expected_error_msg: str
+) -> None:
+    """Check that an expected exception is raised."""
+    caught_expected_exception = False
+    try:
```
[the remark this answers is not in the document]
syedriko: Done

ols/app/models/config.py
syedriko: I'm leaning towards renaming the LLMConfig class into LLMProviders, the Config.llm_config attribute into Config.llm_providers.

syedriko: @raptorsun 
```
if "LLMProviders" in data:
    self.llm_config = LLMConfig(data["LLMProviders"])
```
is not entirely equivalent, if there's a dict entry "LLMProviders"->None, it'll pass that None on.

raptorsun: yes, let's keep it that way to prevent the case where "LLMProviders" entry has None value.

ols/app/models/config.py
```diff
@@ -2,12 +2,25 @@
...
 from typing import Dict, Optional
+from urllib.parse import urlparse
 
 from pydantic import BaseModel
 
 from ols import constants
 
 
+def is_valid_http_url(url) -> bool:
+    """Check is a string is a well-formed http or https URL."""
```
bparees, line 13: let's use a regex for this so we can fully validate the url.  See: https://www.makeuseof.com/regular-expressions-validate-url/ for an example.

ols/utils/config.py
```diff
@@ -23,130 +23,36 @@ def load_empty_config() -> None:
...
-    )
+    data = yaml.safe_load(stream)
+    c = config_model.Config(data)
+    c.validate()
+    ols_config = c.ols_config
+    llm_config = c.llm_config
     default_logger = Logger(
-        logger_name="default",
-        log_level=logger_config.default_level,
+        logger_name="default", log_level=c.ols_config.logger_config.default_level
```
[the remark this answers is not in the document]
bparees: also cc @onmete as he was poking around the logging configuration/initialization space

ols/app/models/config.py
syedriko: Assuming OLS doesn't interpret provider and model URLs but simply passes them through, I don't think we should validate them too strictly ourselves or may be even not at all. I'd rather pass them along and let the code that makes use of them fail if it needs to, and we will report that error back to the user.
The example that you found thinks that "https://127.0.0.1:1234/" or "https://localhost:80/" are invalid URLs and we haven't touched IPv6 yet. ChatGPT generated me 
```
import re

def is_valid_url(url):
    regex = re.compile(
        r'^(https?://)'  # http:// or https://
        r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+[A-Z]{2,6}\.?|'  # domain
        r'localhost|'  # localhost
        r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|'  # ...or ipv4
        r'\[?[A-F0-9]*:[A-F0-9:]+\]?)'  # ...or ipv6
        r'(?::\d+)?'  # optional port
        r'(?:/?|[/?]\S+)$', re.IGNORECASE)

    return re.match(regex, url) is not None
```
which at least attempts to handle IPv4 and IPv6 addresses. But unless there's a good reason for us to do the validation and we can lift the validation code from somewhere and that somewhere stays current on the RFCs that define URLs, I'd rather stay away from it. Otherwise we're going to hear users complaining that OLS is rejecting their valid URLs.

bparees: I agree we don't want to define/implement/maintain our own validation.  I'd be very surprised if there's not an existing python lib that will do the validation for us so my proposal is to find+use one of those.  But if there's not clearly an existing library we can use for this, i'm fine w/ deferring it for now.

syedriko: But that is exactly what the code currently is doing, it's calling into just such a library https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse and checks that the scheme is http[s] and there's a netloc, so there's a network address to try to connect to. If you look at https://docs.python.org/3/library/urllib.parse.html#url-parsing-security, they're even admitting that "What constitutes a URL is not universally well defined." And if someone plugs in a bad URL, the network layer will figure it out. I just don't think this is something we need to get too deep into.

bparees: ugh. i am very sorry.  My eyes went right over the "urlparse" call and directly into the checking for an http or https protocol.  this is fine.

ols/utils/config.py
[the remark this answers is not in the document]
syedriko: @bparees If you don't mind, I'd like to make refactoring of the global logger and conversation cache a separate PR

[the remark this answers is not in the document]
bparees: perfectly fine

tests/unit/test_config.py
syedriko: @bparees When loading configuration from a file with ols.utils.config.init_config(), the user will see, just as you say, "Failed to load config file {config_file}: {str(e)}". These tests call load_config_from_stream() directly, it doesn't handle exceptions.
I added these ^^^ tests to see how yaml.safe_load() handles 'bad' YAML. Of the four tests, only the first one is actually malformed, the other three are valid YAML but useless as OLS config and the AttributeErrors are raised by the config model code like
```
       if data is None:
            return
        v = data.get("LLMProviders") // AttributeError gets raised here
```
The AttributeError tests cover all the possible Python types we can encounter in this situation.
What can be a better option is 
```
        if not isinstance(data, dict):
            return
        v = data.get("LLMProviders")
```
which for the AttributeError tests would bail immediately and later fail validation with `InvalidConfigurationError: no LLMProviders found`

Gotta love dynamic typing. You declare `data: Optional[dict]`, check that it's not None, but it still be not a dict.

ols/utils/config.py
onmete: What about wrapping all the config objects to `OLSConfig`? The `init_config` can then return the config as a whole in the `main` (at the app entrypoint). So we would have something like this there
```python
config = init_config()
```
Then you can access this global everywhere in the app like
```python
from ols.main import config

redis_host = config.ols_config.conversation_cache.redis.url
```

syedriko: Agree 100%. I'm hoping to get rid of global ols_config and llm_config, just in a different PR, or this one will never merge :)

tests/unit/test_config.py
```diff
@@ -0,0 +1,352 @@
...
+        )
+    except Exception:
+        print(traceback.format_exc())
+        assert False
+
+
+def test_valid_config_file() -> None:
+    """Check if a valid configuration file is handled correctly."""
+    try:
+        config.init_config("olsconfig.yaml")
```
bparees, line 285: if we're going to have a test based on a fixture data file, please create a directory structure for test data and put this there.

And then separately you can provide a sample config yaml in the top level of the repository

ols/app/models/config.py
```diff
@@ -27,15 +39,34 @@ def __init__(self, data: Optional[dict] = None):
             return
         self.name = data.get("name", None)
         self.url = data.get("url", None)
-        self.credential_path = data.get("credential_path", None)
+        self.credentials = data.get("credentials", None)
```
bparees, line 42: we now have two fields named "credentials" one of which takes a filesystem path and the other which takes the actual credentials as the value.  (one is for provider creds, the other is for redis creds).

we should be consistent in aligning the naming and values, so please keep this name as `credentials_path` so it's clear to what sort of value is expected

bparees (changes requested):

ols/app/models/config.py
syedriko: @bparees There're two perspectives. From the point of view of someone writing a config file, they are specifying a path to a file with credentials and `credentials_path` makes sense. From the point of view of someone writing code using config, there's just the value of credentials, so `credentials` makes sense. I tried to accommodate both perspectives in one of the past iterations where there were both fields in ProviderConfig and ModelConfig and I lifted the name of the file from `credentials_path` and populated `credentials' with the contents of that file. Or we can only use `credentials_path` in the config file, and have a `credentials` config attribute to access from the code, like so:
```
class ModelConfig(BaseModel):
...
    credentials: Optional[str] = None

   def __init__(self, data: Optional[dict] = None): 
       credentials_path = data.get("credentials_path", None)
        if credentials_path is not None:
            with open(credentials_path, mode="r") as f:
                self.credentials = f.read().rstrip()
...                
```

syedriko: @bparees On the Redis creds, don't we want to load them from a file as well? They are only mentioned in the sample config at https://gist.github.com/bparees/89419b57a54d52dfdfe5fa1506a7d174 but didn't make it to the code https://github.com/openshift/lightspeed-service/blob/aa8e5891849c69bdc7d0f1f0c81e4415fb9943bf/ols/app/models/config.py#L78

bparees: > On the Redis creds, don't we want to load them from a file as well?

yes, i'd expect them to be mounted into the OLS pod on disk also and then referenced from the config yaml

bparees: > Or we can only use credentials_path in the config file, and have a credentials config attribute to access from the code

this approach works for me

tisnik:
it looks pretty good right now. I guess that those `global` will be refactored in new PR.

syedriko:
> it looks pretty good right now. I guess that those `global` will be refactored in new PR.

Right, together with the logger and  the conversation cache

bparees (commented):
this gives us a solid foundation for a proper yaml based config, thanks!

bparees:
i've got some follow up items, but this gets the core code in place so we can continue iterating on the exact structure of the config and ensure all the rest of the OLS code is consuming the config properly

/approve
/lgtm

olsconfig.yaml
```diff
@@ -0,0 +1,22 @@
...
+OLSConfig:
+  enable_debug_ui: false
+  conversation_cache:
+    type: memory
+    memory:
+      max_entries: 1000
+  logger_config:
+    default_level: INFO
+  default_provider: bam
+  default_model: ibm/granite-13b-chat-v2
```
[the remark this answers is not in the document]
bparees: it looks like this was an accident/leftover, the implementation doesn't expect this format so i'll just fix the example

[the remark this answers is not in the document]
bparees: sorry, forgot this is actually part of the model name with BAM

PR #168, opened 2024-01-17
Fix enabling dev ui via env

onmete:
## Description

Enabling `/ui` via env is broken now. There are three different envs in different parts of the code that should do the trick and enable the ui. 
```
OLS_ENABLE_UI=false
ENABLE_DEBUG_UI=false
ENABLE_DEV_UI=true
```
but just `ENABLE_DEV_UI` is actually used for configuration.

We should fix this regardless of [config refactoring PR](https://github.com/openshift/lightspeed-service/pull/166).

## Type of change

- [X] Bug fix

### DIFF
--- a/default.env
+++ b/default.env
@@ -1,8 +1,8 @@
 #######################################
 ## GLOBAL SETTINGS
 #######################################
-## enabe OLS UI
-OLS_ENABLE_UI=True
+## enable OLS UI
+OLS_ENABLE_DEV_UI=True
  
 # logging config
 LOG_LEVEL=INFO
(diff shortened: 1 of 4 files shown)


default.env
```diff
@@ -1,8 +1,8 @@
 #######################################
 ## GLOBAL SETTINGS
 #######################################
-## enabe OLS UI
-OLS_ENABLE_UI=True
+## enable OLS UI
+ENABLE_DEV_UI=True
```
bparees, line 5: thanks for bringing sanity/consistency here, but please keep the OLS prefix on the env var name.

onmete: changed

PR #174, opened 2024-01-17
OLS-185: improve unit test coverage of ols.app module.

raptorsun:
## Description
This PR improves unit test coverage of ols.app module.
The current coverage is 90%+ for ols.app module and 90% overall.
image

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue [OLS-185](https://issues.redhat.com//browse/OLS-185), [OLS-89](https://issues.redhat.com//browse/OLS-89)
- Closes [OLS-185](https://issues.redhat.com//browse/OLS-185)

- [x] If it is a core feature, I have added thorough tests. (unit test only)

## Testing

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -131,7 +131,7 @@ class ConversationCacheConfig(BaseModel):
         elif self.type == "in-memory":
             if "in-memory" not in data:
                 raise InvalidConfigurationError("in-memory config is missing")
-            self.memory = MemoryConfig(data.get("memory", None))
+            self.memory = MemoryConfig(data.get("in-memory", None))
 
 
 class LoggerConfig(BaseModel):
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -124,14 +124,14 @@ class ConversationCacheConfig(BaseModel):
         if data is None:
             return
         self.type = data.get("type", None)
-        if self.type == "redis":
-            if "redis" not in data:
+        if self.type == constants.REDIS_CACHE:
+            if constants.REDIS_CACHE not in data:
                 raise InvalidConfigurationError("redis config is missing")
-            self.redis = RedisConfig(data["redis"])
-        elif self.type == "in-memory":
-            if "in-memory" not in data:
-                raise InvalidConfigurationError("in-memory config is missing")
-            self.memory = MemoryConfig(data.get("in-memory", None))
+            self.redis = RedisConfig(data[constants.REDIS_CACHE])
+        elif self.type == constants.IN_MEMORY_CACHE:
+            if constants.IN_MEMORY_CACHE not in data:
+                raise InvalidConfigurationError("memory config is missing")
+            self.memory = MemoryConfig(data.get(constants.IN_MEMORY_CACHE, None))
 
 
 class LoggerConfig(BaseModel):
--- a/tests/unit/test_app_endpoints.py
+++ b/tests/unit/test_app_endpoints.py
@@ -1,0 +1,124 @@
+"""Unit tests for app endpoints."""
+
+import os
+from http import HTTPStatus
+from unittest.mock import patch
+
+import pytest
+from fastapi import HTTPException
+
+from ols import constants
+from ols.app.endpoints import feedback, ols
+from ols.app.models.models import FeedbackRequest, LLMRequest
+from ols.utils import config
+from tests.mock_classes.llm_chain import mock_llm_chain
+from tests.mock_classes.llm_loader import mock_llm_loader
+
+
+@pytest.fixture(scope="module")
+@patch.dict(os.environ, {"LOG_LEVEL": "INFO"})
+def load_config():
+    """Load config before unit tests."""
+    config.load_config_from_env()
+
+
+def test_user_feedback(load_config):
+    """Test user feedback API endpoint."""
+    feedback_request = FeedbackRequest(
+        conversation_id=123,
+        feedback_object='{"rating": 5, "comment": "Great service!"}',
+    )
+    response = feedback.user_feedback(feedback_request)
+    assert response == {"status": "feedback received"}
+
+
+@patch("ols.src.query_helpers.question_validator.QuestionValidator.validate_question")
+@patch("ols.src.docs.docs_summarizer.DocsSummarizer.summarize")
+@patch("ols.src.query_helpers.yaml_generator.YamlGenerator.generate_yaml")
+@patch("ols.utils.config.conversation_cache.get")
+def test_conversation_request(
+    mock_conversation_cache_get,
+    mock_generate_yaml,
+    mock_summarize,
+    mock_validate_question,
+    load_config,
+):
+    """Test conversation request API endpoint."""
+    # valid question, no yaml
+    mock_validate_question.return_value = [constants.VALID, constants.NOYAML]
+    mock_summarize.return_value = (
+        "Kubernetes is an open-source container-orchestration system...",  # summary
+        "",  # referenced documents
+    )
+    llm_request = LLMRequest(query="Tell me about Kubernetes")
+    response = ols.conversation_request(llm_request)
+    assert (
+        response.response
+        == "Kubernetes is an open-source container-orchestration system..."
+    )
+
+    # valid question, yaml
+    mock_validate_question.return_value = [constants.VALID, constants.YAML]
+    mock_generate_yaml.return_value = "content: generated yaml"
+    llm_request = LLMRequest(query="Generate a yaml")
+    response = ols.conversation_request(llm_request)
+    assert response.response == "content: generated yaml"
+
+    # valid question, yaml, generator failure
+    mock_validate_question.return_value = [constants.VALID, constants.YAML]
+    mock_generate_yaml.return_value = constants.SOME_FAILURE
+    with pytest.raises(HTTPException) as excinfo:
+        llm_request = LLMRequest(query="Generate a yaml")
+        response = ols.conversation_request(llm_request)
+        assert excinfo.status_code == HTTPStatus.INTERNAL_SERVER_ERROR
+
+    # question of unknown type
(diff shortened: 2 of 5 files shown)


ols/app/models/config.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, data: Optional[dict] = None):
         elif self.type == "in-memory":
             if "in-memory" not in data:
                 raise InvalidConfigurationError("in-memory config is missing")
-            self.memory = MemoryConfig(data.get("memory", None))
+            self.memory = MemoryConfig(data.get("in-memory", None))
```
raptorsun, line 136: make it consistent with line 134 `if "in-memory" not in data`

tisnik: @bparees I recall we talked about it. Is it correct?^

ols/src/llms/requirements.txt
```diff
@@ -1,4 +1,4 @@
 openai
 ibm_watson_machine_learning
-ibm-generative-ai
+ibm-generative-ai==0.6.1
```
tisnik, line 3: why is it needed for unit tests?

tests/unit/test_app_endpoints.py
```diff
@@ -0,0 +1,121 @@
...
+        llm_request = LLMRequest(query="Generate a yaml")
+        response = ols.ols_request(llm_request)
+        assert excinfo.status_code == 500
+
+    # question of unknown type
+    mock_validate_question.return_value = [constants.VALID, "unknown question type"]
+    with pytest.raises(HTTPException) as excinfo:
+        llm_request = LLMRequest(query="ask a question of unknown type")
+        response = ols.ols_request(llm_request)
+        assert excinfo.status_code == 500
```
tisnik, line 78: there are named constants in `requests.codes`

tests/unit/test_app_endpoints.py
```diff
@@ -0,0 +1,121 @@
...
+    response = ols.ols_request(llm_request)
+    assert response.response == "content: generated yaml"
+
+    # valid question, yaml, generator failure
+    mock_validate_question.return_value = [constants.VALID, constants.YAML]
+    mock_generate_yaml.return_value = constants.SOME_FAILURE
+    with pytest.raises(HTTPException) as excinfo:
+        llm_request = LLMRequest(query="Generate a yaml")
+        response = ols.ols_request(llm_request)
+        assert excinfo.status_code == 500
```
tisnik, line 71: there are named constants in requests.codes

tisnik: hmm does it try to perform real HTTP request?

tests/unit/utils/test_app_utils.py
```diff
@@ -0,0 +1,9 @@
+"""Unit Test for the Utils class."""
+from ols.app.utils import Utils
+
+
+def test_get_suid():
+    """Test get_suid method."""
+    suid = Utils.get_suid()
+    assert len(suid) == 32
+    assert isinstance(suid, str)
```
tisnik, line 9: in theory you might test it using regexp

tisnik (commented):
It looks nice on overall! Thank you!
just have some questions and nitpicks

tisnik:
OT: the overall coverage will be higher if you rebase :)

ols/src/llms/requirements.txt
raptorsun: because the latest version of `ibm-generative-ai` is 2.0.0 that breaks our code on [an import of genai.schema](https://github.com/openshift/lightspeed-service/blob/2a458752ec0908e32480ba0935c469b5c3db8c9b/ols/src/llms/llm_loader.py#L153)
The v2.0.0 has removed that module of genai.schema.

tests/unit/test_app_endpoints.py
raptorsun: I will update the PR using requests.codes.
It does not perform real HTTP request, just run the handler function's instructions with some mocks.

ols/src/llms/requirements.txt
tisnik: yup, but you can mock the LLM loader completely

raptorsun:
> OT: the overall coverage will be higher if you rebase :)

The code has been rebased and updated on the HTTP status code constants. 

The total coverage is 90% now :D

ols/app/models/config.py
bparees: we talked about it.  my recommendation was to use `memory` consistently, and not use `in-memory`

bparees: cc @syedriko since this applies to the config.yaml format/loading also.

tests/unit/test_app_endpoints.py
```diff
@@ -0,0 +1,122 @@
...
+
+from ols import constants
+from ols.app.endpoints import feedback, ols
+from ols.app.models.models import FeedbackRequest, LLMRequest
+from ols.utils import config
+
+
+@pytest.fixture(scope="module")
+@patch.dict(os.environ, {"LOG_LEVEL": "INFO"})
+@patch.dict(os.environ, {"BAM_API_KEY": "BAM API KEY"})
```
bparees, line 18: more stuff that's going to be impacted by the switch to config.yaml loading instead of env loading.

cc @syedriko

tests/unit/test_app_endpoints.py
```diff
@@ -0,0 +1,122 @@
...
+
+@patch("ols.src.query_helpers.question_validator.QuestionValidator.validate_question")
+@patch("ols.src.docs.docs_summarizer.DocsSummarizer.summarize")
+@patch("ols.src.query_helpers.yaml_generator.YamlGenerator.generate_yaml")
+@patch("ols.utils.config.conversation_cache.get")
+def test_ols_request(
+    mock_conversation_cache_get,
+    mock_generate_yaml,
+    mock_summarize,
+    mock_validate_question,
```
bparees, line 42: is there some python magic that is mapping the mocks created by the patch commands to these argument names?  I don't doubt it works but i'd like to understand how.

bparees (commented):
please split out the tests for the config models and work with @syedriko on the best way to land them.  It may make more sense to include them in his PR, or on top of his PR.  He may also be introducing his own versions of these sorts of tests.

tests/unit/test_app_endpoints.py
raptorsun: when we remove the function `config.load_config_from_env()`, we can update this fixture function. The only purpose is to instantiate the logger which is called in many places in the codes to test. 
Using a mock on `LLMLoader` and `LLMChain` the variable `BAM_API_KEY`  becomes useless and will be removed.

tests/unit/test_app_endpoints.py
raptorsun: These mock functions' references are added by the `@patch` decorator in a "first in last out" order. In the test we can use them to modify their behavior such as their return values for different test cases. 
For example we can let the function `ols.src.query_helpers.question_validator.QuestionValidator.validate_question` return `[constants.VALID, constants.NOYAML]` for the 1st call while return `[constants.VALID, constants.YAML]` for the 2nd call in this test.

raptorsun:
> please split out the tests for the config models and work with @syedriko on the best way to land them. It may make more sense to include them in his PR, or on top of his PR. He may also be introducing his own versions of these sorts of tests.

@syedriko and I discussed about putting our works together. We agree that in PR#166 we test the config from parsing YAML string to the config's behavior. In this PR#174 we test only the data model of the config structures.

raptorsun:
A new commit is added to make the key consistent for in-memory cache configuration, too. @bparees

tests/unit/test_app_endpoints.py
bparees: ok, so there's an ordering relationship between the @patch annotations and the args being passed to the test function.  But the order is inverted(first @patch annotation -> last argument).  Got it.

bparees:
/approve

I am going to leave it to others to do the detailed review(lgtm) on this, but I do want to call out that this is doing a lot of mock based testing that is going to be fragile as we refactor the things it is mocking (e.g. the entire /ols flow of doing question validation and prompt submission)

tests/unit/test_app_endpoints.py
tisnik: nice

tests/unit/test_app_endpoints.py
tisnik: nice

tests/unit/utils/test_app_utils.py
tisnik: yes

tests/unit/test_app_endpoints.py
tisnik: yes. The fact is that any decorator __transforms__ the original function into some other function. So the "last" decorator takes the original function X, producing function Y. That function Y is basically passed into "last-1" decorator, producing function Z etc. That means - decorators are higher order function called during module import.

(that idea is super nice, so IMHO it's good to know what's going on - decorators are different beasts than, for example, annotations in Java, just by coincidence are written similarly)

raptorsun:
PR rebased on main branch, incorperating the function name changes in `ols.app.endpoints`.

tests/unit/utils/test_app_utils.py
raptorsun: pattern check has been added now using the pattern `[a-f0-9]{32}`.
The test for length is remove as the pattern already checks the length of 32 characters.

tests/unit/test_models.py
```diff
@@ -0,0 +1,501 @@
+"""Unit tests for data models."""
```
bparees, line 1: this seems to have significant overlap with the work @syedriko is doing on config.yaml here: https://github.com/openshift/lightspeed-service/pull/166

are you working talking with him?

PR #175, opened 2024-01-17
OLS-186: Checks how improper LLM loading is handled by LLM loader

tisnik:
## Description

Checks how improper LLM loading is handled by LLM loader

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

- Related Issue #[OLS-186](https://issues.redhat.com//browse/OLS-186)

## Testing
- Will be done on CI

### DIFF
--- a/tests/unit/llms/test_llm_loader.py
+++ b/tests/unit/llms/test_llm_loader.py
@@ -1,5 +1,7 @@
 """Unit tests for LLMLoader class."""
 
+from unittest.mock import patch
+
 import pytest
 
 from ols import constants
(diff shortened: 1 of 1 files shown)


tisnik:
This or https://github.com/openshift/lightspeed-service/pull/177 is correct, not both. Need to decide :)

tests/unit/llms/test_llm_loader.py
```diff
@@ -44,7 +47,7 @@ def test_constructor_wrong_provider():
 
 
 @pytest.mark.parametrize("provider, model", llm_cfgs)
-def test_constructor_correct_provider_no_models(provider, model):
+def _test_constructor_correct_provider_no_models(provider, model):
```
bparees, line 50: is the intent here to disable this test?  if it's being replaced by the new test why not just delete it?

bparees:
> This or https://github.com/openshift/lightspeed-service/pull/177 is correct, not both. Need to decide :)

#177 is closer to correct in that it raises the exception in llmloader when the libraries for a particular provider can't be loaded.

tests/unit/llms/test_llm_loader.py
tisnik: actually it was leftover :/

tisnik:
> > This or #177 is correct, not both. Need to decide :)
> 
> #177 is closer to correct in that it raises the exception in llmloader when the libraries for a particular provider can't be loaded.

then let's merge #177 and then update these tests - inverse the logic basically ;)

tisnik:
updated @bparees

tests/unit/llms/test_llm_loader.py
```diff
@@ -56,3 +58,32 @@ def test_constructor_correct_provider_no_models(provider, model):
...
+    providerConfig = ProviderConfig()
+    providerConfig.models = {constants.GRANITE_13B_CHAT_V1: None}
+    config.llm_config.providers = {provider: providerConfig}
+
+    def mock_import(module, *args, **kwargs):
+        """Mock the import and from x import statements."""
+        pass
+
+    # check what happens if LLM libraries can not be loaded
+    with patch("builtins.__import__", side_effect=mock_import):
```
bparees, line 87: for my own understanding, how is this raising an exception on import?  the side_effect function you're defining is a no-op, so why doesn't the import succeed?

tisnik: the import should return "something" (it's module). Pass technically is the same as `return None` in Python (yes, that language has it's dark corners) so it fails later with something like "Module not Found"

bparees: i assumed the default Mock would also return None by default.

tisnik: yes it does

bparees: i guess the mock returns a mock by default, so that is why you need to override the behavior via the side effect?

PR #176, opened 2024-01-17
OLS-40: Use pydantic ability to provide default uuid

onmete:
## Description

Changes:
- UUID is generated for request when not provided (at the model init)
- using full uuid instead of just hex
  - 81fde9cae75143238f71949faf40a23b -> 81fde9ca-e751-4323-8f71-949faf40a23b
- conversation cache is asked every time
  - As the retrieval is very fast (compared to other processing logic), would it be a problem? Maybe if there are 100/1000 paralell queries?

Also, it seems we are adding conversation to the cache only for yaml questions. Is that correct behavior?

If you agree with the changes, I'll proceed with the tests.

## Type of change

- [X] Refactor

## TODO
- [ ] tests

### DIFF
--- a/ols/app/models/models.py
+++ b/ols/app/models/models.py
@@ -1,8 +1,9 @@
 """Data models representing payloads for REST API calls."""
 
 from typing import Union
+from uuid import UUID, uuid4
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class LLMRequest(BaseModel):
@@ -10,7 +11,7 @@ class LLMRequest(BaseModel):
 
     Attributes:
         query: The query string.
-        conversation_id: The optional conversation ID.
+        conversation_id: The optional conversation id in valid UUID format.
         response: The optional response.
 
     Example:
@@ -18,7 +19,7 @@ class LLMRequest(BaseModel):
     """
 
     query: str
-    conversation_id: Union[str, None] = None
+    conversation_id: UUID = Field(default_factory=uuid4)
     response: Union[str, None] = None
 
 
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -27,16 +27,15 @@ def ols_request(llm_request: LLMRequest) -> LLMRequest:
     logger = config.default_logger
 
     # Initialize variables
-    previous_input = None
-    conversation = llm_request.conversation_id
+    conversation = str(llm_request.conversation_id)
+    previous_input = config.conversation_cache.get(conversation)
 
-    # Generate a new conversation ID if not provided
-    if conversation is None:
-        conversation = Utils.get_suid()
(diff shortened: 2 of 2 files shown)


tisnik:
it looks like the usage of Pydantic there is pretty idiomatic. Not sure about UUID etc. WDY other means?

bparees:
> Also, it seems we are adding conversation to the cache only for yaml questions. Is that correct behavior?

no, the conversation cache should be used for all request flows in which a valid conversation id is provided on the request.

ols/app/models/models.py
```diff
@@ -1,24 +1,25 @@
...
+        conversation_id: The optional conversation id in valid UUID format.
         response: The optional response.
 
     Example:
         llm_request = LLMRequest(query="Tell me about Kubernetes")
     """
 
     query: str
-    conversation_id: Union[str, None] = None
+    conversation_id: UUID = Field(default_factory=uuid4)
```
bparees, line 22: does this change break the existing gradio UI?

also if it doesn't break any existing tests that means we are missing tests that should be added a part of this change to enforce the api.

onmete: Some tests are failing already. I wanted to first ensure that this change makes sense to you. If so, I'll update tests and cover missing bits.

onmete: UI is not broken with this change

onmete: And I file a bug about missing conv cache and do it as a part of OLS-40, ok?

bparees:
> UUID is generated for request when not provided (at the model init)

I don't think we actually do want this behavior, because we lose track of the client/user intent.

If the user sent a uuid, then they assume there is a previous conversation to reference.  And if we can't find that previous conversation, that's worthy of note (it may have expired out of the cache, or something may have gone wrong).

If they didn't send a uuid, then it is a new conversation.

So we should not be defaulting it in the model, we need to understand whether the client request included a conversation id or not.

ols/app/models/models.py
bparees: > And I file a bug about missing conv cache and do it as a part of OLS-40, ok?

incorporation the conversation cache into all the flows is part of https://issues.redhat.com/browse/OLS-31 and in particular https://issues.redhat.com/browse/OLS-59

onmete:
We are able to tell if it's a new conversation or not. According to the cache entry. But true is, if we somehow lost the cache, we wouldn't be able to differentiate between this edge case and a new conversation.

bparees:
> if we somehow lost the cache, we wouldn't be able to differentiate between this edge case and a new conversation.

that's my point.  And at least right now we do intend to expire cache entries (based on cache size) so it will be possible (though hopefully we set the cache size large enough that this never actually happens to a user).

regardless, i think it's important to be able to distinguish between a uuid we expect to find in the cache, and one we do not expect to find in the cache.

PR #177, opened 2024-01-17
Raise exception when selected LLM can not be loaded

tisnik:
## Description

Raise exception when selected LLM can not be loaded

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Testing

### DIFF
--- a/tests/integration/test_ols.py
+++ b/tests/integration/test_ols.py
@@ -9,6 +9,7 @@ from ols import constants
 from ols.app.main import app
 from ols.src.query_helpers.question_validator import QuestionValidator
 from tests.mock_classes.llm_chain import mock_llm_chain
+from tests.mock_classes.llm_loader import mock_llm_loader
 
 client = TestClient(app)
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_ols.py
```diff
@@ -49,7 +50,9 @@ def test_status() -> None:
 
 # the raw prompt should just return stuff from LLMChain, so mock that base method in ols.py
 @patch("ols.app.endpoints.ols.LLMChain", new=mock_llm_chain("test response"))
-def test_raw_prompt(monkeypatch) -> None:
+# during LLM init, exceptins will occur on CI, so let's mock that too
+@patch("ols.app.endpoints.ols.LLMLoader", new=mock_llm_loader(None))
```
bparees, line 54: i don't understand how this is working....how does the mock llm loader end up ensuring that the raw prompt request returns "test response" as the result?

tisnik: well we just need to skip the `LLMLoader.__init__` failure, because it is initialized in the code, but unused (because the `LLMChain` is mocked too).

bparees: i missed the existing patch command for mock_llm_chain, my bad

PR #184, opened 2024-01-18
OLS-40: Llm loader parent error

onmete:
## Description

Contains two commits:
- Custom errors class naming is formatted according to google's standards.
- Adds parent configuration error to make catching these errors easy on the endpoint side (except just parent instead of 5 different ones.)
  - the assumption here is that all those errors are the same from the user's perspective -> 422 unprocessable entity

## Type of change

- [X] Refactor
- [X] New feature

- [X] If it is a core feature, I have added thorough tests.

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -17,23 +17,23 @@ from ols.utils.logger import Logger
 warnings.simplefilter("ignore", UserWarning)
 
 
-class MissingProvider(Exception):
+class MissingProviderError(Exception):
     """Provider is not specified."""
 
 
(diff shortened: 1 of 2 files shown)


tisnik:
/lgtm

(not sure about class hierarchy test, but it can't do any harm anyway)

onmete:
As the intention is to catch the parent error instead of all children, I think it is important to ensure those are actually children. But yeah, it might be an overkill.

raptorsun:
/lgtm
It should be a good idea to have this parent exception. Most times we just need to distinguish configuration error from other errors, no need to dig into the details.

PR #193, opened 2024-01-19
OLS-84: Refactoring: require LLM init to return the right object or fail in all cases

tisnik:
## Description

Refactoring: require LLM init to return the right object or fail in all cases
Why:
1) previously the code allows some initializers to simply don't touch `self.llm` and still be "fine"
2) the code that uses `self.llm` later (all query helpers etc.) will need to check `llm is not None` everywhere which is not there in the current code (and will be overkill)
3) now any LLM to be initialized **must** return the right object **or** throws an exception (the other possibilities will be detected)
4) it might help in the future with modularization

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-84](https://issues.redhat.com//browse/OLS-84)
- Closes #

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -3,7 +3,7 @@
 import inspect
 import os
 import warnings
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from langchain.callbacks.manager import CallbackManager
 from langchain.callbacks.streaming_stdout import StreamingStdOutCallbackHandler
@@ -11,6 +11,9 @@ from langchain.callbacks.streaming_stdout import StreamingStdOutCallbackHandler
 from ols import constants
 from ols.utils import config
 from ols.utils.logger import Logger
+
+if TYPE_CHECKING:
+    from langchain.llms.base import LLM
 
 # workaround to disable UserWarning
 warnings.simplefilter("ignore", UserWarning)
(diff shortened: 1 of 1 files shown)


ols/src/llms/llm_loader.py
```diff
@@ -78,25 +81,24 @@ def __init__(
 
         # return empty dictionary if not defined
         self.llm_params = params if params else {}
-        self.llm = None
-        self._set_llm_instance()
+        self.llm: LLM = self._llm_instance()
 
-    def _set_llm_instance(self):
+    def _llm_instance(self):
```
bparees, line 86: @tisnik shouldn't this specify the return type since we seem to be moving to doing that everywhere?

ols/src/llms/llm_loader.py
```diff
@@ -12,6 +12,9 @@
 from ols.utils import config
 from ols.utils.logger import Logger
 
+if TYPE_CHECKING:
+    from langchain.llms.base import LLM
```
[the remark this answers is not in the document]
tisnik: Python itself basically ignores type hints completely. So it does not need to import the library in runtime. Just Mypy or Pyright need this info. This import will be needed in so called typing-only context.

ols/src/llms/llm_loader.py
tisnik: yes it would be better do done in explicitly there, correct! (Mypy can infer the type, but return type is part of function/method documentation)

ols/src/llms/llm_loader.py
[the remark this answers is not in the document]
bparees: got it, thanks

ols/src/llms/llm_loader.py
bparees: you want to submit a follow up for that?

for us python newbies there's gonna be a lot of copying+pasting from the existing code examples, so we should try to ensure the existing code reflects how we want new code to look

tisnik: https://github.com/openshift/lightspeed-service/pull/194

bparees: ah, looks like you already did in https://github.com/openshift/lightspeed-service/pull/194

PR #199, opened 2024-01-22
Remove unused url parameter in LLMLoader

onmete:
## Description

Remove unused url parameter in LLMLoader.

## Type of change

- [X] Refactor

## Checklist before requesting a review

- [x] I have performed a self-review of my code.
- [x] PR has passed all pre-merge test jobs.

PR #200, opened 2024-01-22
OLS-233: Llmloader unify error handling

onmete:
## Description

This is a small step towards [OLS-233](https://issues.redhat.com/browse/OLS-233).

I've removed the parametrized test (as it was testing the same thing) and added new tests instead.

Watson loader seems to be incompatible a bit with the rest of the LLMLoader code. I can't peek around because I don't have credentials for it (or BAM).

Merge after https://github.com/openshift/lightspeed-service/pull/199

## Type of change

- [x] Refactor

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -10,6 +10,7 @@ from typing import Optional
 from langchain.llms.base import LLM
 
 from ols import constants
+from ols.app.models.config import ProviderConfig
 from ols.utils import config
 from ols.utils.logger import Logger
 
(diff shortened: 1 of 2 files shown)


ols/src/llms/llm_loader.py
```diff
@@ -58,28 +59,47 @@ def __init__(
...
         provider: Optional[str] = None,
         model: Optional[str] = None,
-        url: Optional[str] = None,
         params: Optional[dict] = None,
         logger=None,
     ) -> None:
         """Initialize loader using provided provider, model, and other parameters."""
         self.logger = logger if logger is not None else Logger("llm_loader").logger
-        if provider is None:
+        if not provider:
```
bparees, line 67: i don't know what python style guides say, but i find `provider is None` to be more readable/obvious than `not provider`, given that provider isn't a boolean.

ols/src/llms/llm_loader.py
```diff
@@ -58,28 +59,47 @@ def __init__(
...
+    # TODO: refactor after config implementation OLS-89
+    def _get_provider_config(self) -> ProviderConfig:
+        cfg = config.llm_config.providers.get(self.provider)
+        if not cfg:
+            raise UnsupportedProvider()
+
+        model = cfg.models.get(self.model)
+        if not model:
+            # NOTE: temporary condition for WatsonX - implementation is
+            # incompatible with the rest of LLMLoader now
```
bparees, line 92: the WatsonX loader implementation can't handle us throwing a ModelConfigMissingException?

ols/src/llms/llm_loader.py
onmete: `if not provider` check for any "falsy" value, but I agree `provider is None` is explicit here, reverting

ols/src/llms/llm_loader.py
onmete: It's more like we don't care about it in that case. Currently, we don't have any Watson model in constants or config. It is incompatible in the sense it is not using provider/model values, it is going for envs `WATSON_MODEL` and `WATSON_PROJECT_ID` directly. As it doesn't make sense to use `LLMLoader` with `provider="watsonx"` and `model="something"`, when model is not used, I wanted to separate it.
But I can remove that condition, and leave it for refactoring when someone actually gets credentials to it. WDYT?

onmete: ok, removed, rebased

PR #206, opened 2024-01-23
Don't expect all types libs

tisnik:
## Description

Don't expect all types libs

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,5 +16,5 @@ line-length = 100
 fail_under = 50
 
 [tool.mypy]
-disable_error_code = ["union-attr", "return-value", "arg-type", "misc", "call-arg"]
+disable_error_code = ["union-attr", "return-value", "arg-type", "misc", "call-arg", "import-untyped"]
 ignore_missing_imports = true


bparees:
@tisnik can you provide a little more detailed description on what this does?  I am assuming it means don't do type checking on imported libs, but it's not clear from the commit/PR description.

(I care less about fixing this PR in particular and more about providing better descriptions in the future)

tisnik:
> @tisnik can you provide a little more detailed description on what this does? I am assuming it means don't do type checking on imported libs, but it's not clear from the commit/PR description.
> 
> (I care less about fixing this PR in particular and more about providing better descriptions in the future)

Mypy has one downside - if it does not find type hints in imported library, it shows an error saying "you should install this" or just "I don't know what to do". The second problem is covered by `ingore_missing_imports` configuration option. The first one is tricky as any new version of Mypy have better knowledge about what to install. It means that even w/o changing anything in our code it might start to fail. IMHO it's better to simply ignore this case for now than to try install every library during CI check (for now).

bparees:
>  if it does not find type hints in imported library, it shows an error saying "you should install this" 

I think I ran into this here:
https://github.com/openshift/lightspeed-service/pull/203/files#diff-685da804fbcac569d75387e475e57d1de687a54c6c41b3aa4057694cfb5abc4bR7

let me know if we should remove that dep.

PR #215, opened 2024-01-24
feat: [OLS-104] Add token handler utility for context

asamal4:
## Description

We need a mechanism to limit tokens based on context window.
The modification covers below
- utility to convert text to tokens
- get approximate token count
- limit the context based on context window.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-104](https://issues.redhat.com//browse/OLS-104)
- Closes #

### DIFF
--- a/ols/src/utils/token_handler.py
+++ b/ols/src/utils/token_handler.py
@@ -1,0 +1,93 @@
+"""Utility to handle tokens."""
+import tiktoken
+
+from ols.src.utils.constants import (
+    MINIMUM_CONTEXT_LIMIT,
+    TOKENIZER_MODEL,
+)
+from ols.utils.logger import Logger
+
+
+class TokenHandler:
+    """This class handles tokens.
+
+    Convert text to tokens.
+    Get rough estimation of token count.
+    Truncate text based on token limit.
+    """
+
+    def __init__(self):
+        """Initialize the class instance."""
+        self._logger = Logger("token_handler").logger
+        self._encoder = tiktoken.get_encoding(TOKENIZER_MODEL)
+
+    def text_to_tokens(self, text: str) -> list[int]:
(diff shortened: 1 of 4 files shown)


tisnik:
It looks nice so far.
But are you going to integrate it with some existing code?

ols/src/utils/token_handler.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Args:
+            tokens: ex: [1, 2, 3, 4]
+
+        Returns:
+            text: ex "This is my doc"
+        """
+        return self._encoder.decode(tokens)
+
+    def handle_context(self, retrieved_nodes, prompt_token_len: int) -> list[dict]:
```
bparees, line 50:
```suggestion
    def truncate_rag_context(self, retrieved_nodes, prompt_token_len: int) -> list[dict]:
```

"context" is going to be a very overloaded term, let's be more specific.  Also truncate better reflects what this actually does.

bparees, line 50: i think this will be more modular/easier to use if instead of taking the prompt token len as an arg, it just took a max length for the context.

The caller can decide how many tokens they have available or want to dedicate to RAG context content.  This function should just be responsible for returning content that is constrained to that desired length.

bparees:
> @tisnik This will be called from doc_summarizer. And multiple changes required in summarizer module (including unit test) so I was thinking of a separate PR for that.. But integration coverage is marginally less than the threshold for this.. I will let you guys decide whether to hold this or not..

separate PR is totally reasonable.  Deliver the utility logic in one PR, wire it up in another.

PR #217, opened 2024-01-24
Add QueryHelper parent

onmete:
## Description

This PR unifies the query helpers interface. If there is a need to configure provider/model per helper, we can still do something like this at the place where it's initialized (in ols endpoint now). 
```python
If config.yaml_provider:
   yaml_gen = YamlGenerator(provider=config.yaml_provider, …)
# it will use default/provided one otherwise
```

Allows dependency injection in tests.

The `camel_to_snake` is there only to ensure the same logger name as before (it will be refactored eventually).

## Type of change

- [X] Refactor

### DIFF
--- a/ols/src/query_helpers/__init__.py
+++ b/ols/src/query_helpers/__init__.py
@@ -1,1 +1,28 @@
 """Question validators, statament classifiers, and response generators."""
+import re
+
+from ols.utils import config
+from ols.utils.logger import Logger
+
+
+def camel_to_snake(string: str) -> str:
+    """Convert a string from camel case to snake case.
+
+    Args:
+        string: The string to be converted.
+
+    Returns:
+        The converted string.
+    """
+    string = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", string)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", string).lower()
+
+
+class QueryHelper:
+    """Base class for query helpers."""
+
+    def __init__(self, provider: str | None = None, model: str | None = None):
+        """Initialize query helper."""
+        self.provider = provider or config.ols_config.default_provider
+        self.model = model or config.ols_config.default_model
+        self.logger = Logger(camel_to_snake(self.__class__.__name__)).logger
--- a/tests/unit/query_helpers/test_query_helper.py
+++ b/tests/unit/query_helpers/test_query_helper.py
@@ -1,0 +1,39 @@
+"""Unit tests for query helper class."""
+
+from ols.src.query_helpers import QueryHelper, camel_to_snake
+from ols.utils import config
+
+
+def test_camel_to_snake():
+    """Test camel case to snake case conversion."""
+    assert camel_to_snake("") == ""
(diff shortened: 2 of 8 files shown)


tests/unit/query_helpers/test_query_helper.py
```diff
@@ -0,0 +1,41 @@
...
+    """Test camel case to snake case conversion."""
+    assert camel_to_snake("") == ""
+    assert camel_to_snake("dog") == "dog"
+    assert camel_to_snake("LOUD") == "loud"
+    assert camel_to_snake("quiet") == "quiet"
+    assert camel_to_snake("OCP4") == "ocp4"
+    assert camel_to_snake("Is there a space?") == "is there a space?"
+    assert camel_to_snake("rainyDay") == "rainy_day"
+    assert camel_to_snake("NiceDay") == "nice_day"
+    assert camel_to_snake("AGoodDay") == "a_good_day"
```
onmete, line 17: Yes, this can be parametrized, but I found this better readable. I can change it though.

ols/src/query_helpers/__init__.py
```diff
@@ -1 +1,28 @@
...
+        string: The string to be converted.
+
+    Returns:
+        The converted string.
+    """
+    string = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", string)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", string).lower()
+
+
+class QueryHelper:
```
onmete, line 21: We also have `DocsSummarizer`. It is currently loading configuration from env. I would propose to move it into `query_helpers/` and make it a subclass of  `QueryHelper` too. WDYT?

tisnik: probably worth to put into utils if needed in other places?

tisnik:
it looks ok in general, thank you Ondro

ols/src/query_helpers/__init__.py
onmete: What part? This `QueryHelper` parent?

bparees:
/approve

general concept makes sense

bparees:
/lgtm

merging so https://github.com/openshift/lightspeed-service/pull/219 can continue forward, but i believe you intended to follow up on this with an update to move DocsSummarizer under this same structure

tests/unit/query_helpers/test_query_helper.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TestQueryHelper:
+    """Test the query helper class."""
+
+    def test_defaults_used(self):
+        """Test that the defaults are used when no inputs are provided."""
+        # TODO: I don't like this setting - should be possible to
+        # refactor after OLS-89
+        config.load_config_from_env()
```
bparees, line 27: this will need to be fixed now that the config yaml change has landed

```
FAILED tests/unit/query_helpers/test_query_helper.py::TestQueryHelper::test_defaults_used - AttributeError: module 'ols.utils.config' has no attribute 'load_config_from_env'
```

PR #223, opened 2024-01-24
OLS-103 - Completed basic retriever

mpaulgreen:
## Description

This is the code for basic retriever as described in 
https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=18820&projectKey=OLS&view=detail&selectedIssue=[OLS-103](https://issues.redhat.com//browse/OLS-103)
## Type of change

- [ ] Refactor
- [ X] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

- [X ] I have performed a self-review of my code.

## Testing
- unit test/integration test included

### DIFF
--- a/ols/src/query_helpers/query_docs.py
+++ b/ols/src/query_helpers/query_docs.py
@@ -1,0 +1,114 @@
+"""This module has the components to retrieve docs from vector store."""
+from typing import Any
+
+from langchain_core.documents.base import Document
+from langchain_core.vectorstores import VectorStore, VectorStoreRetriever
+
+from ols.utils.logger import Logger
+
+
+class RetrieveDocsExceptionError(Exception):
+    """Issue with retrieving the docs."""
+
+
+class QueryDocs:
+    """Generic interface for retrieving docs from any vector store."""
+
+    def __init__(self) -> None:
+        """Initialize the `QueryDocs` instance."""
+        self.logger = Logger("QueryDocs").logger
+
+    def get_relevant_docs(
+        self,
+        query: str,
+        vectordb: VectorStore = None,
+        search_type: str = "similarity",
+        **search_kwargs: Any,
+    ) -> list[Document] | None:
+        """Return list of documents from the vectorstore.
+
+        Args:
+            query: string to find relevant documents for
+            vectordb: Vector db from which you need to retrieve the docs.
+            search_type (Optional[str]): Defines the type of search that
+                the retriever should perform.
+                Can be "similarity" (default), "mmr", or
+                "similarity_score_threshold".
+            search_kwargs (Optional[Dict]): Keyword arguments to pass to the
+                search function. Can include things like:
+                    k: Amount of documents to return (Default: 4)
+                    score_threshold: Minimum relevance threshold
+                        for similarity_score_threshold
+                    fetch_k: Amount of documents to pass to MMR algorithm (
+                    Default: 20)
+                    lambda_mult: Diversity of results returned by MMR;
+                        1 for minimum diversity and 0 for maximum. (Default:
+                        0.5)
+                    filter: Filter by document metadata
+
+        Returns:
+            List[Document]: Returns a list of docs.
+
+        Examples:
+           code-block:: python
+
+            # Retrieve more documents with higher diversity
+            # Useful if your dataset has many similar documents
+            docsearch.as_retriever(
+                search_type="mmr",
+                search_kwargs={'k': 6, 'lambda_mult': 0.25}
+            )
+
+            # Fetch more documents for the MMR algorithm to consider
+            # But only return the top 5
+            docsearch.as_retriever(
+                search_type="mmr",
+                search_kwargs={'k': 5, 'fetch_k': 50}
+            )
+
+            # Only retrieve documents that have a relevance score
+            # Above a certain threshold
+            docsearch.as_retriever(
+                search_type="similarity_score_threshold",
+                search_kwargs={'score_threshold': 0.8}
+            )
+
+            # Only get the single most similar document from the dataset
+            docsearch.as_retriever(search_kwargs={'k': 1})
+
+            # Use a filter to only retrieve documents from a specific paper
+            docsearch.as_retriever(
+                search_kwargs={'filter': {'paper_title':'GPT-4 Technical
+                Report'}}
+            )
+        """
+        self.logger.info(
+            f"""Retrieving docs for
+                query: {query},
+                vectordb: {vectordb},
+                search_type: {search_type},
+                search_kwargs: {search_kwargs}
+        """
+        )
+
+        if search_type not in ("mmr", "similarity", "similarity_score_threshold"):
+            self.logger.error(f"incorrect search type {search_type}")
+            raise RetrieveDocsExceptionError(
+                f"""search type is invalid: {search_type}"""
+            )
+
+        db_retriever: VectorStoreRetriever = vectordb.as_retriever(
+            search_type=search_type, search_kwargs=search_kwargs["search_kwargs"]
+        )
+
+        try:
+            docs = db_retriever.get_relevant_documents(query=query)
+        except Exception:
+            self.logger.error(
+                f"exception raised while getting the docs for query: {query}"
+            )
+            raise RetrieveDocsExceptionError(
+                "error in getting the  docs from vectorstore"
+            )
+
+        return docs
--- a/tests/integration/query_helpers/test_query_docs.py
+++ b/tests/integration/query_helpers/test_query_docs.py
@@ -1,0 +1,111 @@
+"""Integration tests using light weight FAISS index."""
+import os
+
+import pytest
+from langchain_community.embeddings import HuggingFaceEmbeddings
+from langchain_community.vectorstores.faiss import FAISS
+from langchain_core.documents.base import Document
+
+from ols.src.query_helpers.query_docs import QueryDocs, RetrieveDocsExceptionError
+
+
+@pytest.fixture
+def setup_faiss():
+    """Documents stored as  up FAISS index."""
+    list_of_documents = [
+        Document(page_content="foo", metadata={"page": 1, "source": "adhoc"}),
+        Document(page_content="bar", metadata={"page": 1, "source": "adhoc"}),
+        Document(page_content="foo", metadata={"page": 2, "source": "adhoc"}),
+        Document(page_content="barbar", metadata={"page": 2, "source": "adhoc"}),
+        Document(page_content="foo", metadata={"page": 3, "source": "adhoc"}),
+        Document(page_content="bar burr", metadata={"page": 3, "source": "adhoc"}),
+        Document(page_content="foo", metadata={"page": 4, "source": "adhoc"}),
+        Document(page_content="bar bruh", metadata={"page": 4, "source": "adhoc"}),
+    ]
+    embeddings = HuggingFaceEmbeddings(model_name="all-MiniLM-L6-v2")
+
+    file_path = "./index/faiss_index"
+
+    if os.path.exists(file_path):
+        db = FAISS.load_local(file_path, embeddings)
+    else:
+        db = FAISS.from_documents(list_of_documents, embeddings)
+    yield db
+    db.save_local(file_path)
+
+
+def test_retrieve_top_k_similarity_search(setup_faiss: FAISS):
+    """Fetch top k similarity search."""
+    docs = QueryDocs().get_relevant_docs(
+        vectordb=setup_faiss, query="foo", search_kwargs={"k": 1}
+    )
+
+    assert len(docs) == 1
+    assert docs[0].page_content == "foo"
+    assert docs[0].metadata["page"] == 1
+    assert docs[0].metadata["source"] == "adhoc"
+
+
+def test_retrieve_mmr(setup_faiss: FAISS):
+    """Fetch more documents for the MMR algorithm to consider. But only return the top 1."""
+    docs = QueryDocs().get_relevant_docs(
+        vectordb=setup_faiss,
+        search_type="mmr",
+        query="foo",
+        search_kwargs={"k": 1, "fetch_k": 4},
+    )
+
+    assert len(docs) == 1
+    assert docs[0].page_content == "foo"
+    assert docs[0].metadata["page"] == 1
+    assert docs[0].metadata["source"] == "adhoc"
+
+
+def test_similarity_score(setup_faiss: FAISS):
+    """Fetch only the docs that has scores above similarity score threshold."""
+    docs = QueryDocs().get_relevant_docs(
+        vectordb=setup_faiss,
+        search_type="similarity_score_threshold",
+        query="foo",
+        search_kwargs={"score_threshold": 0.3},
+    )
+
+    assert len(docs) == 4
+    assert docs[0].page_content == "foo"
+    assert docs[0].metadata["page"] == 1
+    assert docs[1].page_content == "foo"
+    assert docs[1].metadata["page"] == 2
+    assert docs[2].page_content == "foo"
+    assert docs[2].metadata["page"] == 3
+    assert docs[3].page_content == "foo"
+    assert docs[3].metadata["page"] == 4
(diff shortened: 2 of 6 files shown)


mpaulgreen:
@bparees his implementation is a generic implementation of VectorStoreRetriever that will work with most vector dbs. 
It will work with most common vector db
 - There are various tuning option that is provided within the this retriever.
    - The various tuning options have been covered as part tests and docs.

The search by vectors are still not implemented by langchain for this generic interface
    https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores.py#L386

Once we decide upon a vector db for OLS we can extend this class to implment search by vector methods 
 - As an example this is the code for FAISS(VectorStore)
    - https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/faiss.py#L272

From the ADR it seems like no vector db has been chosen yet
    https://docs.google.com/document/d/1ChhHpmB-EZZ0hQXVbBuWbN6Fa0XAngQTDIMPyNF0SII/edit#heading=h.66y4kqbj468a

Lastly it is sufficiently modular to allow tuning at this point.

mpaulgreen:
@tisnik @bparees when I run make verify on my local it passes.
- on local
```
black . --check
All done! ✨ 🍰 ✨
61 files would be left unchanged.
ruff . --per-file-ignores=tests/*:S101

```, but it is failing on CI. Any recommendation?

File: ols/src/query_helpers/query_docs.py [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+                "error in getting the  docs from vectorstore"
+            )
+
+        return docs
+
+
+class RetrieveDocsExceptionError(Exception):
+    """Raise this exception when there is an issue in retrieving the docs."""
+
+    pass
```
onmete, line 124: The common practise (we follow) is to define exceptions at the top (after imports). Could you move it there?

Also, you can remove the `pass` here. The docstring itself satisfies the requirement for the class to have some content.

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,124 @@
...
+        self.logger = Logger("QueryDocs").logger
+
+    def get_relevant_docs(
+        self,
+        query: str,
+        perform_query: bool = True,
+        vectordb: VectorStore = None,
+        search_type: str = "similarity",
+        **search_kwargs: Any,
+    ) -> List[Document] | None:
```
onmete, line 24: Instead of `List` from typing, use the built-in generic type `list`.

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,124 @@
...
+                    lambda_mult: Diversity of results returned by MMR;
+                        1 for minimum diversity and 0 for maximum. (Default:
+                        0.5)
+                    filter: Filter by document metadata
+
+        Returns:
+            List[Document]: Returns a list of docs.
+
+        Examples:
+        .. code-block:: python
```
onmete, line 51: This format `.. code-block:: python` seems more specific for sphinx style. Google's style should support 
```
Examples:
   ```python
   some code
   ```.
```
(ignore the dot, I'm fighting with formatting there)

But I guess this could be influence by the actual tool we choose for documentation build (IRC we don't know at the moment).

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,124 @@
...
+                query: {query},
+                perform_query: {perform_query},
+                vectordb: {vectordb},
+                search_type: {search_type},
+                search_kwargs: {search_kwargs}
+        """
+        )
+
+        if not perform_query:
+            return None
```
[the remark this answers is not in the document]
mpaulgreen: This is one of the acceptance criteria. This was specifically discussed in backlog refinement and was asked to be part of this component

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,124 @@
...
+            self.logger.error(f"incorrect search type {search_type}")
+            raise RetrieveDocsExceptionError("search type in invalid")
+
+        db_retriever: VectorStoreRetriever = vectordb.as_retriever(
+            search_type=search_type, search_kwargs=search_kwargs["search_kwargs"]
+        )
+
+        try:
+            docs = db_retriever.get_relevant_documents(query=query)
+        except Exception:
```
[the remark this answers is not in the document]
mpaulgreen: Good point. Unfortunately langchain is throwing a catch all exception - https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/retrievers.py#L222

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+        if not perform_query:
+            return None
+
+        if not (
+            search_type == "mmr"
+            or search_type == "similarity"
+            or search_type == "similarity_score_threshold"
+        ):
```
onmete, line 100: Can be refactored as
```python
if search_type not in ("a", "b", ...):
```

ols/src/query_helpers/query_docs.py
mpaulgreen: @onmete The whole class is a wrapper around langchain code. The reason why some of the elements are what it is there is because of the way langchain code has been written. Please check the return type here https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/retrievers.py#L182

ols/src/query_helpers/query_docs.py
mpaulgreen: https://peps.python.org/pep-0020/ ```Explicit is better than implicit```

onmete: I don't think its applicable here. What explicit piece of information is `pass` providing? [This ](https://github.com/openshift/lightspeed-service/blob/main/ols/src/llms/llm_loader.py#L21) is how we are defining exceptions in code.

ols/src/query_helpers/query_docs.py
onmete: That shouldn't matter. You can try with linter, but we are typing the return type of our code. Given aside the `List` and `list` is actually the same thing, since python 3.9 (give or take) it is recommended to use python build-ins, just because you don't need to import "the same" thing  from somewhere else. (it wasn't possible in earlier versions)

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,124 @@
...
+            )
+            raise RetrieveDocsExceptionError(
+                "error in getting the  docs from vectorstore"
+            )
+
+        return docs
+
+
+class RetrieveDocsExceptionError(Exception):
+    """Raise this exception when there is an issue in retrieving the docs."""
```
onmete, line 122: The [Google docstring standard](https://google.github.io/styleguide/pyguide.html#384-classes) (we use) implies a specific syntactic of the exception docstring. We are [already using](https://github.com/openshift/lightspeed-service/blob/main/ols/src/llms/llm_loader.py#L26) it. So this should be something like "Issue with retrieving the docs."

ols/src/query_helpers/query_docs.py
mpaulgreen: As I mentioned this is a wrapper and data structures will follow close to what has the class is wrapping around.

ols/src/query_helpers/query_docs.py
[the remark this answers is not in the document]
onmete: Not necessarily, it catches all exceptions, but promoting those to users (by raising it again). So when `ValueError` is raised there, it's catched (as its `Exception`), but we should be able to catch `ValueError` on our side.
But I guess this will be refactored when we'll use this and start encountering various exceptions.

ols/src/query_helpers/query_docs.py
onmete: You are missing the point here.

onmete: The langchain is not returning `List`, but `list`. They just use `List` in the typing. Because of reasons I explained above. Most likely because they are still supporting python 3.8, where you can't use buildins as type hints.

tisnik:
> @tisnik @bparees when I run make verify on my local it passes.
> 
>     * on local
> 
> 
> ```
> black . --check
> All done! ✨ 🍰 ✨
> 61 files would be left unchanged.
> ruff . --per-file-ignores=tests/*:S101
> 
> ```,
>  but it is failing on CI. Any recommendation?
> ```

Do you happen to have the latest version of black? Also is everything commit into this PR?

ols/src/query_helpers/query_docs.py
tisnik: `list` (lowercase) is perfectly fit there. You can check types - it will be ok

tests/integration/query_helpers/test_query_docs.py
```diff
@@ -0,0 +1,126 @@
...
+        Document(page_content="foo", metadata={"page": 3, "source": "adhoc"}),
+        Document(page_content="bar burr", metadata={"page": 3, "source": "adhoc"}),
+        Document(page_content="foo", metadata={"page": 4, "source": "adhoc"}),
+        Document(page_content="bar bruh", metadata={"page": 4, "source": "adhoc"}),
+    ]
+    embeddings = HuggingFaceEmbeddings(model_name="all-MiniLM-L6-v2")
+
+    file_path = "./index/faiss_index"
+
+    if os.path.exists(file_path):
```
tisnik, line 29: is it ok to have two possible ways how unit test will run? It will be different on CI, I guess

tisnik: prob worth to create it in all cases?

tisnik (commented):
some nitpicks, but looks ok in overall

tests/integration/query_helpers/test_query_docs.py
mpaulgreen: Think of it as a light weight db. So creating it for every case will slow the test down. So we create one index store and set of document embeddings in the store and then test all out cases. Is it helpful?

mpaulgreen:
@asamal4 Very good point. The vector store index will be specific to vector store and embeddings will depend on the model with which you create those.. Refer ADR, both langchain and llamaindex is in consideration.

ols/src/query_helpers/query_docs.py
mpaulgreen: done

ols/src/query_helpers/query_docs.py
mpaulgreen: done

ols/src/query_helpers/query_docs.py
mpaulgreen: done

ols/src/query_helpers/query_docs.py
mpaulgreen: done

ols/src/query_helpers/query_docs.py
mpaulgreen: done

tests/integration/query_helpers/test_query_docs.py
```diff
@@ -0,0 +1,126 @@
...
+        query="foo",
+        search_kwargs={"score_threshold": 0.3, "filter": {"page": 1}},
+    )
+
+    assert docs is None
+
+
+def test_invalid_search_type(setup_faiss: FAISS):
+    """Test for invalid search type."""
+    with pytest.raises(RetrieveDocsExceptionError) as ex:
```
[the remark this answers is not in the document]
mpaulgreen: done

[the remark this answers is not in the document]
mpaulgreen: Thanks I learnt something new ;)

tests/mock_classes/mock_retrievers.py
```diff
@@ -0,0 +1,36 @@
+"""Mocks for VectorStore and VectorStoreRetriever."""
+from langchain_core.documents import Document
+
+
+class MockVectorStore:
+    """Mock for VectorStore."""
+
+    def __init__(self):
+        """Init."""
+        pass
```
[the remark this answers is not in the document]
mpaulgreen: done

mpaulgreen:
@asamal4 llamaindex and langchain are two libraries that can be both used for RAG searches and both are in consideration. if you look into some docs Langchain is preferred over llamaindex for chats(conversation). I am not trying to debate langchain over llamaindex here as they are both in consideration as per ADR. The existing code has to be looked as a POC and several elements of it is addressed in JIRA issue. One such example is that of QuestionAnswer Validator. This is anyhow a basic retriever not to obtain necessarily a best response but to be tunable as per JIRA description

tests/integration/query_helpers/test_query_docs.py
```diff
@@ -0,0 +1,125 @@
+"""Integration tests using light weight FAISS index."""
```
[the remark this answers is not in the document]
mpaulgreen: This is not a mock vector store. mock vector is in unit test cases. This is an index. Anyhow if you dont write integration test for your components it will fail the test-integration check due to insufficient test coverage. This is very common practice when you testing db integration where you write you test cases with lightweight db.

ols/src/query_helpers/query_docs.py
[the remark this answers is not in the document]
bparees: +1 @onmete 

retrieval should be optional means "in the overall OLS flow, when a user sends us a question there may be cases where we cannot do, or choose not to do, retrieval augmented generation for that particular query" (maybe because no RAG index is configured, maybe because we understand the question won't benefit from it, maybe some other reason).

But that decision would be made closer to the request flow and the logic would simply not invoke `get_relevant_docs`.  Having logic that calls `get_relevant_docs` while passing an argument that says "but don't actually retrieve anything" doesn't make much sense....  you should just not make the function call in the first place.

The very nature of breaking this logic out into its own function so we can make an explicit choice about invoking it or not, is fulfilling the intent of "make retrieval optional" (or at least is one part of addressing that requirement)

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,118 @@
...
+                search_kwargs: {search_kwargs}
+        """
+        )
+
+        if not perform_query:
+            return None
+
+        if search_type not in ("mmr", "similarity", "similarity_score_threshold"):
+            self.logger.error(f"incorrect search type {search_type}")
+            raise RetrieveDocsExceptionError("search type in invalid")
```
bparees, line 98: i'd include the value of search_type that was provided, in the exception.  and i think `in` should be `is`

ols/src/query_helpers/query_docs.py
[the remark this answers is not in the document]
mpaulgreen: @bparees it did not make sense me too. But this was discussed in the backlog refinement meeting and therefore I put it in there. I  will remove that.

ols/src/query_helpers/query_docs.py
```diff
@@ -0,0 +1,118 @@
...
+
+        db_retriever: VectorStoreRetriever = vectordb.as_retriever(
+            search_type=search_type, search_kwargs=search_kwargs["search_kwargs"]
+        )
```
bparees, line 102: do you know if this logic is going to load the db content from disk every time (at least in the case of filebased indexes)?  If so we're going to want to ensure we're only doing this once at startup time.  (we'll restart OLS if the rag db content is changed)

bparees: ah, i guess it's probably loaded when the Store is loaded, not when the the retriever is created so this is probably ok

ols/src/query_helpers/query_docs.py
mpaulgreen: done, will be seen in next push

bparees:
> @bparees all the comments have been addressed. So you can merge it at your convenience. In case you still see any issues let me know.

i'll give it an approval as it establishes the right structure, i'll let @asamal4 and @onmete finalize their reviews though, before we merge it

/approve

PR #224, opened 2024-01-24
Updated linter rule for pytest.fixture

tisnik:
## Description

Updated linter rule for `pytest.fixture`
There are two ways how to call `pytest.fixture` without arguments:

```python
@pytest.fixture
def my_fixture():
    ...
```

or:

```python
@pytest.fixture()
def my_fixture():
    ...
```

Both variants are perfectly ok, but it's better to choose one to be consistent. I choosed the 1st one that's already in our sources, so we just need to inform linters about it.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Testing
- N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ ignore = []
 target-version = "py311"
 pydocstyle.convention = "google"
 line-length = 100
+flake8-pytest-style.fixture-parentheses = false
 
 [tool.coverage.report]
 # unit tests fails if the total coverage measurement is under this threshold value


bparees:
I think i would have chosen the rule that consistently enforces parens are always used but it's not like it's difficult to change this decision later.  Good at have the consistency enforcement either way.

/approve
/lgtm

PR #225, opened 2024-01-24
OLS-212: Add test case for missing doc store scenario

sauagarwa:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -11,6 +11,9 @@ llm_providers:
       - name: gpt-4-1106-preview
       - name: gpt-3.5-turbo
 ols_config:
+  reference_content:
+#    product_docs_index_path: "./vector-db/ocp-product-docs"
+#    product_docs_index_id: product
   conversation_cache:
     type: memory
     memory:
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -1,6 +1,7 @@
 """Config classes for the configuration structure."""
 
 import logging
+import os
 from typing import Optional
 from urllib.parse import urlparse
 
@@ -306,11 +307,61 @@ class LoggingConfig(BaseModel):
         """Validate logger config."""
 
 
+class ReferenceContent(BaseModel):
+    """Reference content configuration."""
+
+    product_docs_index_path: Optional[str] = None
+    product_docs_index_id: Optional[str] = None
+
+    def __init__(self, data: Optional[dict] = None):
+        """Initialize configuration and perform basic validation."""
+        super().__init__()
+        if data is None:
+            return
+
+        self.product_docs_index_path = data.get("product_docs_index_path", None)
+        self.product_docs_index_id = data.get("product_docs_index_id", None)
+
+    def __eq__(self, other):
+        """Compare two objects for equality."""
+        if isinstance(other, ReferenceContent):
+            return (
+                self.product_docs_index_path == other.product_docs_index_path
+                and self.product_docs_index_id == other.product_docs_index_id
+            )
+        return False
+
+    def validate_yaml(self) -> None:
+        """Validate reference content config."""
+        if self.product_docs_index_path is not None:
+            if os.path.exists(self.product_docs_index_path) is False:
+                raise InvalidConfigurationError(
+                    f"Reference content path '{self.product_docs_index_path}' does not exist"
+                )
+            if os.path.isfile(self.product_docs_index_path):
+                raise InvalidConfigurationError(
+                    f"Reference content path '{self.product_docs_index_path}' is not a directory"
+                )
+            if self.product_docs_index_id is None:
+                raise InvalidConfigurationError(
+                    "product_docs_index_path is specified but product_docs_index_id is missing"
+                )
+
+        if (
+            self.product_docs_index_id is not None
+            and self.product_docs_index_path is None
+        ):
+            raise InvalidConfigurationError(
+                "product_docs_index_id is specified but product_docs_index_path is missing"
+            )
+
+
 class OLSConfig(BaseModel):
     """OLS configuration."""
 
     conversation_cache: Optional[ConversationCacheConfig] = None
     logging_config: Optional[LoggingConfig] = None
+    reference_content: Optional[ReferenceContent] = None
 
     default_provider: Optional[str] = None
     default_model: Optional[str] = None
@@ -346,6 +397,7 @@ class OLSConfig(BaseModel):
             data.get("conversation_cache", None)
         )
         self.logging_config = LoggingConfig(data.get("logging_config", None))
+        self.reference_content = ReferenceContent(data.get("reference_content", None))
 
     def __eq__(self, other) -> bool:
         """Compare two objects for equality."""
@@ -361,6 +413,7 @@ class OLSConfig(BaseModel):
                 and self.summarizer_model == other.summarizer_model
                 and self.validator_provider == other.validator_provider
                 and self.validator_model == other.validator_model
+                and self.reference_content == other.reference_content
             )
         return False
 
@@ -370,6 +423,8 @@ class OLSConfig(BaseModel):
             raise InvalidConfigurationError("OSLConfig: conversation cache is not set")
         self.conversation_cache.validate_yaml()
         self.logging_config.validate_yaml()
+        if self.reference_content is not None:
+            self.reference_content.validate_yaml()
 
 
 class DevConfig(BaseModel):
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -154,8 +154,6 @@ GPT35_TURBO = "gpt-3.5-turbo"
 # indexing constants
 PRODUCT_INDEX = "product"
 PRODUCT_DOCS_PERSIST_DIR = "./vector-db/ocp-product-docs"
-SUMMARY_INDEX = "summary"
-SUMMARY_DOCS_PERSIST_DIR = "./vector-db/summary-docs"
 
 # cache constants
 IN_MEMORY_CACHE = "memory"
--- a/tests/config/singleprovider.e2e.template.config.yaml
+++ b/tests/config/singleprovider.e2e.template.config.yaml
@@ -5,6 +5,9 @@ llm_providers:
     models:
       - name: $MODEL
 ols_config:
+  reference_content:
+    # product_docs_index_path: "./vector-db/ocp-product-docs"
+    # product_docs_index_id: product
   conversation_cache:
     type: memory
     memory:
--- a/tests/integration/test_ols.py
+++ b/tests/integration/test_ols.py
@@ -7,9 +7,9 @@ import requests
 from fastapi.testclient import TestClient
 
 from ols import constants
-from ols.app.models.config import ProviderConfig
+from ols.app.models.config import ProviderConfig, ReferenceContent
 from ols.src.llms.llm_loader import LLMConfigurationError
-from ols.utils import suid
+from ols.utils import config, suid
 from tests.mock_classes.llm_chain import mock_llm_chain
 from tests.mock_classes.llm_loader import mock_llm_loader
 
@@ -315,3 +315,42 @@ class TestQuery:
                     "LLM provider test-provider'"
                 }
             }
+
+
+def test_post_question_on_noyaml_response_type() -> None:
+    """Check the REST API /ols/ with POST HTTP method when unknown response type is returned."""
+    config.init_empty_config()
+    config.ols_config.reference_content = ReferenceContent(None)
+    config.ols_config.reference_content.product_docs_index_path = "./invalid_dir"
+    config.ols_config.reference_content.product_docs_index_id = "product"
+    answer = constants.SUBJECT_VALID
+    with patch(
+        "ols.app.endpoints.ols.QuestionValidator.validate_question", return_value=answer
+    ):
+        from tests.mock_classes.langchain_interface import mock_langchain_interface
+
+        ml = mock_langchain_interface("test response")
+        with patch(
+            "ols.src.query_helpers.docs_summarizer.LLMLoader", new=mock_llm_loader(ml())
+        ):
+            with patch(
+                "ols.src.query_helpers.docs_summarizer.ServiceContext.from_defaults"
+            ):
+                with patch(
+                    "ols.utils.config.ols_config.reference_content.product_docs_index_path",
+                    "./invalid_dir",
+                ):
+                    conversation_id = suid.get_suid()
+                    response = client.post(
+                        "/v1/query",
+                        json={
+                            "conversation_id": conversation_id,
+                            "query": "test query",
+                        },
+                    )
+                    print(response)
+                    assert response.status_code == requests.codes.ok
+                    assert (
+                        "The following response was generated without access to reference content:"
+                        in response.json()["response"]
+                    )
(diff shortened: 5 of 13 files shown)


tests/integration/test_ols.py
```diff
@@ -97,3 +100,40 @@ def test_post_question_on_unknown_response_type() -> None:
...
+    config.load_empty_config()
+
+    def dummy_validator(
+        self, conversation: str, query: str, verbose: bool = False
+    ) -> list[str]:
+        return constants.VALID, constants.NOYAML
+
+    # let's pretend the question is valid without even asking LLM
+    # but the question type is unknown
+    monkeypatch.setattr(QuestionValidator, "validate_question", dummy_validator)
```
tisnik, line 118: you can use `with patch()` there

tests/integration/test_ols.py
```diff
@@ -97,3 +100,40 @@ def test_post_question_on_unknown_response_type() -> None:
         assert response.json() == {
             "detail": {"response": "Internal server error. Please try again."}
         }
+
+
+@patch("ols.constants.PRODUCT_DOCS_PERSIST_DIR", "./invalid_dir")
```
bparees, line 105: when you have to do something like this, it points out an area where the existing code is insufficiently configurable.  After https://github.com/openshift/lightspeed-service/pull/166 merges, please make the RAG directory a configurable item and then update this test to drive the config.

sauagarwa: Will do. Thanks

examples/olsconfig.yaml
```diff
@@ -11,6 +11,8 @@ LLMProviders:
     models:
       - name: gpt-4-1106-preview
 OLSConfig:
+  product_docs_persist_dir: "./vector-db/ocp-product-docs"
+  summary_docs_persist_dir: "./vector-db/summary-docs"
```
bparees, line 15: This should be nested under a new config structure for "reference_content" or something along those lines.  We are likely to going to end up with a handful of config items related to RAG stuff and we aren't going to want them to all be a bunch of top level fields.

ols/app/models/config.py
```diff
@@ -330,6 +330,8 @@ class OLSConfig(BaseModel):
     validator_model: Optional[str] = None
     yaml_provider: Optional[str] = None
     yaml_model: Optional[str] = None
+    product_docs_persist_dir: Optional[str] = constants.PRODUCT_DOCS_PERSIST_DIR
+    summary_docs_persist_dir: Optional[str] = constants.SUMMARY_DOCS_PERSIST_DIR
```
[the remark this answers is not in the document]
thoraxe: +1 to drop summary. old experiment, crusty code.

examples/olsconfig.yaml
sauagarwa: @bparees : What about rag_content, or rag_config or just rag?

bparees: i don't really want "rag" to appear in our config spec field names.  it's not sufficiently self-explanatory.

ols/app/models/config.py
```diff
@@ -307,12 +307,44 @@ def _get_log_level(self, data, key, default):
...
+        """Compare two objects for equality."""
+        if isinstance(other, ReferenceContent):
+            return (
+                self.product_docs_persist_dir == other.product_docs_persist_dir
+                and self.product_index == other.product_index
+            )
+        return False
+
+    def validate_yaml(self) -> None:
+        """Validate refernce content config."""
```
bparees, line 339:
```suggestion
        """Validate reference content config."""
```

ols/constants.py
```diff
@@ -194,4 +192,4 @@
 REDIS_CACHE_MAX_MEMORY_POLICY = "allkeys-lru"
 
 # provider and model roles
-PROVIDER_MODEL_ROLES = ("default", "classifier", "summarizer", "validator", "yaml")
+PROVIDER_MODEL_ROLES = ("default", "classifier", "summarizer", "validator", "yaml")
```
bparees, line 195: put the trailing newline back please

ols/src/indexer/indexer.py
bparees, line 1: this file is not used by OLS at runtime.  It is a utility that is used to generate the index from product docs.

As such it shouldn't be tied to the ols config format. 

We should also move it out of the "ols" directory tree and into something more like a "tools" directory or something.  You can work with @ilan-pinto and @syedriko about where to put it, but for now please leave it out of this PR.  (or if you need to fix it up so you can drop the shared constants, do that, but don't tie it to the ols config module)

ilan-pinto: I wonder if we should have it in a different repository.

sauagarwa: @bparees: I can move the configuration out of olsconfig and put it under indexConfig to fix this PR. Would that be ok? We can look into moving the indexer out in separate PR.

bparees: > I can move the configuration out of olsconfig and put it under indexConfig to fix this PR

we don't need any configuration mechanism for it.

bparees: > I wonder if we should have it in a different repository.

eventually it should probably live with the rest of the pipeline-related logic but i'm trying not to get sidetracked into yet another repo setup exercise.

ols/app/models/config.py
```diff
@@ -308,11 +308,43 @@ def validate_yaml(self) -> None:
...
+        if isinstance(other, ReferenceContent):
+            return (
+                self.product_docs_persist_dir == other.product_docs_persist_dir
+                and self.product_index == other.product_index
+            )
+        return False
+
+    def validate_yaml(self) -> None:
+        """Validate reference content config."""
+        pass
```
bparees, line 339: `pass` isn't needed, evidently python treats the code comment as sufficient.

but also we should be validating that the value of product_docs_persist_dir is a valid filesystem path format.  Should probably even confirm a file exists at that location so we can fail on startup if it doesn't exist, rather than failing when a request comes in and we can't load the index.

bparees (commented):
also needs rebasing

ols/app/models/config.py
sauagarwa: I thought about it, but then every one will start failing as they have been running without the index file. Let me know and I can fail if the directory or the files does not exist.

The other thing is that as per your previous comment to take the reference content configuration out of olsconfig i created a separate section in config called indexer_config where the configuration for indexer can go. Let me know your thoughts.

bparees: > I thought about it, but then every one will start failing as they have been running without the index file. 

they won't fail if we default the value to None and tolerate None as valid.

>  i created a separate section in config called indexer_config where the configuration for indexer can go.

indexer config does not belong in the ols config.yaml at all.  The indexer is a standalone utility that no user is ever going to use.  I am going to move it out of this directory so it stops causing people to be confused.

bparees: https://github.com/openshift/lightspeed-service/pull/299

bparees:
@sauagarwa I feel like there is a lot of confusion here about the "indexer" vs the "rag content"


we do not need any configuration for the "indexer", it has nothing to do with runtime OLS.

what we do need is configuration for where to load the rag content from, which is why i previously suggested the config field of "reference_content" for pointing to where the rag content should be loaded from by OLS at runtime.

tests/unit/docs/test_doc_summarizer.py
```diff
@@ -24,8 +26,57 @@ def test_is_query_helper_subclass():
...
+@patch(
+    "ols.src.query_helpers.docs_summarizer.LLMLoader",
+    new=mock_llm_loader(mock_langchain_interface("test response")()),
+)
+@patch("ols.src.query_helpers.docs_summarizer.ServiceContext.from_defaults")
+@patch("ols.src.query_helpers.docs_summarizer.StorageContext.from_defaults")
+@patch(
+    "ols.src.query_helpers.docs_summarizer.load_index_from_storage", new=MockLlamaIndex
+)
+@patch.dict(os.environ, {"TEI_SERVER_URL": "localhost"}, clear=True)
```
bparees, line 48: this environment variable shouldn't be used by the code any more.  @syedriko did we miss one or has a PR not merged yet?

tests/unit/docs/test_doc_summarizer.py
```diff
@@ -24,8 +26,57 @@ def test_is_query_helper_subclass():
...
+
+@patch(
+    "ols.src.query_helpers.docs_summarizer.LLMLoader",
+    new=mock_llm_loader(mock_langchain_interface("test response")()),
+)
+@patch("ols.src.query_helpers.docs_summarizer.ServiceContext.from_defaults")
+@patch(
+    "ols.src.query_helpers.docs_summarizer.load_index_from_storage", new=MockLlamaIndex
+)
+@patch.dict(os.environ, {"TEI_SERVER_URL": "localhost"}, clear=True)
```
bparees, line 68: should not be used/needed any more.  Even w/ the existing code referencing it, it should be fine to leave it unset so i'm not clear why you needed to set it?

```
        embed_model: str | TextEmbeddingsInference = "local:BAAI/bge-base-en"
        # TODO get this from global config instead of env
        # Not a priority because embedding model probably won't be configurable in the final product
        tei_embedding_url = os.getenv("TEI_SERVER_URL", None)
        if tei_embedding_url:
            logger.info(f"{conversation} using TEI embedding server")

            embed_model = TextEmbeddingsInference(
                model_name=constants.TEI_EMBEDDING_MODEL,
                base_url=tei_embedding_url,
            )
```

tests/unit/utils/test_config.py
```diff
@@ -359,6 +359,88 @@ def test_invalid_config():
...
+      max_entries: 1000
+dev_config:
+  llm_temperature_override: 0.1
+  enable_dev_ui: true
+  disable_question_validation: false
+  disable_auth: false
+
+""",
+        InvalidConfigurationError,
+        "product_docs_index_path is specified but product_docs_index_id is missing",
```
bparees, line 414: once you add the validation logic, you'll need the inverse test case:  index id specified but index path missing.

tests/unit/utils/test_config.py
```diff
@@ -359,6 +359,88 @@ def test_invalid_config():
...
+
+    check_expected_exception(
+        """
+---
+llm_providers:
+  - name: p1
+    credentails_path: tests/config/secret.txt
+    models:
+      - name: m1
+        credentails_path: tests/config/secret.txt
```
bparees, line 425:
```suggestion
        credentials_path: tests/config/secret.txt
```

tests/unit/utils/test_config.py
```diff
@@ -359,6 +359,88 @@ def test_invalid_config():
...
+        InvalidConfigurationError,
+        "product_docs_index_path is specified but product_docs_index_id is missing",
+    )
+
+    check_expected_exception(
+        """
+---
+llm_providers:
+  - name: p1
+    credentails_path: tests/config/secret.txt
```
bparees, line 422:
```suggestion
    credentials_path: tests/config/secret.txt
```

bparees: and if you cut+pasted this from elsewhere please fix it there too

tests/config/singleprovider.e2e.template.config.yaml
```diff
@@ -5,6 +5,9 @@ llm_providers:
     models:
       - name: $MODEL
 ols_config:
+  reference_content:
+    product_docs_index_path: "./vector-db/ocp-product-docs"
+    product_docs_index_id: product
```
bparees, line 10: this is going to break/fail the e2e test job because this content isn't available in the e2e environment currently.

did you try running the e2e job on your PR?

If you want an e2e test that references RAG content you'll need to update the e2e script to download the rag content.  You can see how Erik is doing that here: https://github.com/openshift/lightspeed-service/pull/293/files#diff-367f8fb601acfd2e337029840094a854585472d93ad5a7b98c0300da1e80e94cR7

of course if you do that you should also add an e2e test that actually expects to get a result from the LLM that is based on RAG content.

For now it's probably easiest/best to just not use RAG content in the e2e, but please create a follow up story for us to add e2e tests that do leverage RAG content.

bparees (commented):
this is still failing unit and integration tests, in addition to the fact that i think your changes break the e2e job since you updated the e2e config to reference content that isn't available in the e2e job environment.

tests/config/singleprovider.e2e.template.config.yaml
sauagarwa: I did ran the e2e test with RAG content and it passed.

sauagarwa: I had vector-db in my workspace that passed the test cases. Added the RAG content test case in ols-158. I will modify the e2e script to download the index in that PR.

tests/unit/utils/test_config.py
sauagarwa: fixed

ols/app/models/config.py
```diff
@@ -306,11 +307,53 @@ def validate_yaml(self) -> None:
...
+                    f"Reference content path '{self.product_docs_index_path}' does not exist"
+                )
+            if os.path.isfile(self.product_docs_index_path):
+                raise InvalidConfigurationError(
+                    f"Reference content path '{self.product_docs_index_path}' is not a directory"
+                )
+            if self.product_docs_index_id is None:
+                raise InvalidConfigurationError(
+                    "product_docs_index_path is specified but product_docs_index_id is missing"
+                )
```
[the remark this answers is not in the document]
sauagarwa: done

tests/unit/docs/test_doc_summarizer.py
sauagarwa: fixed

tests/config/singleprovider.e2e.template.config.yaml
bparees: > Added the RAG content test case in ols-158. I will modify the e2e script to download the index in that PR.

we can't merge this PR without that change (or by removing the reference content config here).  Even if you can pass the e2e yourself w/ the vector-db present, CI will fail and we can't have failing CI.

please just delete the reference_content config here for the e2e config file and add it back in w/ your PR that updates the e2e script to download the content.

bparees:
/test e2e-ols-openai-gpt35

to confirm my assumption that this PR breaks the e2e job due to the e2e job not having the rag content available currnetly

sauagarwa:
@bparees : I have updated the config file in examples and commented the code
ols_config:
  reference_content:
    #product_docs_index_path: "./vector-db/ocp-product-docs"
    #product_docs_index_id: product

Which ols_config file are you using because the one that is used for e2e the two lines are not commented and thats why the test is failing

Here is the ols_config file that is used with e2e

ols_config:
  reference_content:
    product_docs_index_path: "./vector-db/ocp-product-docs"
    product_docs_index_id: product

bparees:
> Which ols_config file are you using because the one that is used for e2e the two lines are not commented and thats why the test is failing. 

Our e2e job uses the single provider config yaml, which it looks like you have now updated to remove the vectordb ref, so this should pass now.

PR #227, opened 2024-01-24
OLS-89: Config yaml follow-ups

bparees:
## Description

some minor clean up as a follow on to https://github.com/openshift/lightspeed-service/pull/166

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ data
 vector-db
 vendor
 
+# ignore credential files
+*_api_key.txt
+
 # ignore files that begin with tmptest_
 tmptest_*
 
--- a/README.md
+++ b/README.md
@@ -1,66 +1,64 @@
-
-<!-- PROJECT LOGO -->
-
-
-
-<!-- ABOUT THE PROJECT -->
 ## About The Project
 
+OpenShift LightSpeed is an AI powered assistant that runs on OpenShift and provides answers to product questions using backend LLM services.
 
-
-
-### Built With
-
-This project is built using IBM Watson machine learning.
-
-
-
-<!-- GETTING STARTED -->
 ## Getting Started
 
 To get a local copy up and running follow these simple example steps.
 
 ### Prerequisites
 
-
-* Python 3.11 or newer
+* Python 3.11
 * Git and pip installed
-* BAM account
-    - BAM_API_KEY
-    - BAM_URL
+* An LLM api key, currently BAM(IBM's research environment) and OpenAI are offered as backends.
 
 ### Installation
 
-_Below is an example of how you can instruct your audience on installing and setting up your app. This template doesn't rely on any external dependencies or services._
+#### Get API keys
 
+##### BAM provided LLM
 1. Get a BAM API Key at [https://bam.res.ibm.com](https://bam.res.ibm.com)
    * Login with your IBM W3 Id credentials.
    * Copy the API Key from the Documentation section.
      ![BAM API Key](docs/bam_api_key.png)
-2. BAM URL: https://bam-api.res.ibm.com
-3. Clone the repo
+2. BAM API URL: https://bam-api.res.ibm.com
+
+##### OpenAI provided LLM
+Get an [OpenAI api key](https://platform.openai.com/api-keys)
+
+#### Install the code
+1. Clone the repo
    ```sh
    git clone <project git>
    cd lightspeed-service
    ```
-4. Install python packages
+2. Install python packages
    ```sh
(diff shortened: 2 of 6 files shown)


bparees:
@syedriko ptal

.gitignore
```diff
@@ -17,6 +17,10 @@ data
 vector-db
 vendor
 
+# ignore credential files
+bam_api_key.txt
+openai_api_key.txt
+
```
syedriko, line 23: `*_api_key.txt` to establish a pattern and leave a comment to that effect in olsconfig.yaml?

README.md
```diff
@@ -1,66 +1,64 @@
...
-
-<!-- GETTING STARTED -->
 ## Getting Started
 
 To get a local copy up and running follow these simple example steps.
 
 ### Prerequisites
 
-
 * Python 3.11 or newer
```
syedriko, line 11: Nothing to do with config, but I don't think 'newer' is correct here, 3.12 didn't work for me.

README.md
```diff
@@ -1,66 +1,64 @@
...
-     unzip -j <path-to-downloaded-file>/local.zip -d vector-db/ocp-product-docs
-     ```
-<!-- USAGE EXAMPLES -->
+
+#### Configure OLS
+
+OLS configuration is in YAML format. It is loaded from a file referred to by the `OLS_CONFIG_FILE` environment variable and defaults to `olsconfig.yaml` in the current directory. 
+You can find a sample configuration in the `olsconfig.yaml`` file in the top directory of this repository.  
+
+API credentials are in turn loaded from files specified in the config yaml by the `credentials_path` attributes in the configuration YAML. If these paths are relative,
```
syedriko, line 45: we say 'the config yaml' twice in this sentence.

.gitignore
bparees: i'll update the git ignore rule, but ultimately olsconfig.yaml is going to be machine generated by our operator, so i don't think we need to define strong patterns for naming conventions there.

README.md
bparees: interesting, i'll go ahead and change it to just 3.11 since i'm in the file already

README.md
bparees: fixed, thanks

bparees:
/hold

this is broken until i fix the integration tests which currently load the config yaml out of the root directory and expect to find the referenced api key files.  The integration tests, like the unit tests, need to be loading their config from test fixtures, not default files in the root of the repo

tests/integration/test_feedback.py
```diff
@@ -1,9 +1,13 @@
...
 
+import os
+
 import requests
 from fastapi.testclient import TestClient
 
-from ols.app.main import app
+# config file path needs to be set before importing app which tries to load the config
+os.environ["OLS_CONFIG_FILE"] = "tests/config/valid_config.yaml"
+from ols.app.main import app  # noqa: E402
```
bparees, line 10: @syedriko @tisnik @onmete is there a different preferred solution here?

basically our app module loads the config from the OLS_CONFIG_FILE env during import.  We need a way to control what config file is loaded during integration testing where we still want to do a normal init of the application, but we need to control the config file being loaded.  

Probably main.py itself should be refactored so that some of the logic it contains is only executed when run as main and not when imported (i.e. `if __name__ == "__main__"`)

bparees:
/hold cancel

integration test "fixed" though per https://github.com/openshift/lightspeed-service/pull/227#discussion_r1465842730 i think there's a better solution that should be implemented

tests/integration/test_feedback.py
tisnik: it can be done this way. Usually (especially for unit tests) the preferred way is to change environment variables just or one test or for given context via `setup/teardown` mechanism:

```
@patch.dict(os.environ, {"OLS_CONFIG_FILE": "tests/config/valid_config.yaml"})
```

it means that this test won't change behavior of other tests

tisnik: and the import, it's mainly ok for tests

onmete: Seems alright. You either patch env when you need to provide a full config to test or just patch relevant parts of the actual config object (regardless of what is loaded).

bparees: thanks.  i'll leave this as is for now then

PR #241, opened 2024-01-25
Ensure script returns error code from test execution

bparees:
## Description
the trailing echo command in the script lead to it exiting with code 0 instead
of the error code from the make test-e2e invocation, which in turn lead to
prow always considering the job to have succeeded.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/scripts/test-e2e-standalone.sh
+++ b/scripts/test-e2e-standalone.sh
@@ -21,7 +21,12 @@ OLS_LOGS=$ARTIFACT_DIR/ols.log
 
 # Generate ols config.yaml
 export PROVIDER="${PROVIDER:-openai}"
-export PROVIDER_KEY_PATH="${PROVIDER_KEY_PATH:-/home/bparees/creds/llms/openai.key}"
+export PROVIDER_KEY_PATH="${PROVIDER_KEY_PATH:-openai_api_key.txt}"
+if [ ! -e "$PROVIDER_KEY_PATH" ]; then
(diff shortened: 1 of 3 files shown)


tests/e2e/test_api.py
```diff
@@ -42,7 +42,8 @@ def test_invalid_question() -> None:
         "/v1/query", json={"conversation_id": "1234", "query": "test query"}, timeout=20
     )
     print(vars(response))
-    assert response.status_code == requests.codes.unprocessable
+    #assert response.status_code == requests.codes.unprocessable
+    assert response.status_code == 100
```
bparees, line 46: deliberate error for testing the job, will remove before merge

scripts/test-e2e-standalone.sh
```diff
@@ -58,4 +58,3 @@ fi
 
 echo Done waiting for OLS server start, running e2e
 make test-e2e
```
tisnik, line 60: `set -e` at beginning of script might do the magic

bparees: yeah i considered it and may still but i've been burned before by the surprising effects of set -e when you don't want them. For now this should be sufficient i think.

PR #242, opened 2024-01-26
OLS-158: Added e2e test case

sauagarwa:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-158](https://issues.redhat.com//browse/OLS-158)
- Closes # [OLS-158](https://issues.redhat.com//browse/OLS-158)

## Testing

### DIFF
--- a/scripts/test-e2e-standalone.sh
+++ b/scripts/test-e2e-standalone.sh
@@ -16,6 +16,17 @@ TMPDIR=$(mktemp -d)
 # ARTIFACT_DIR is defined when running in a prow job
 ARTIFACT_DIR=${ARTIFACT_DIR:-$TMPDIR}
 
+# temp directory for index store
+RAG_TMP_DIR=$(mktemp -d)
+# Download index store
+RAG_INDEX="https://github.com/ilan-pinto/lightspeed-rag-documents/releases/latest/download/local.zip"
+export RAG_INDEX_DIR="$RAG_TMP_DIR/vector-db/ocp-product-docs"
+# Configure index store
+mkdir -p $RAG_INDEX_DIR \
+  && wget $RAG_INDEX \
+	&& unzip -j local.zip -d $RAG_INDEX_DIR \
+	&& rm -f local.zip
+
 export OLS_CONFIG_FILE="$ARTIFACT_DIR/olsconfig.yaml"
 OLS_LOGS=$ARTIFACT_DIR/ols.log
 
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -63,3 +63,26 @@ def test_invalid_question():
         "response": expected_details,
     }
     assert response.json() == expected_json
+
+
+def test_valid_question() -> None:
+    """Check the REST API /v1/query with POST HTTP method for valid question and no yaml."""
+    response = client.post(
+        "/v1/query",
+        json={"conversation_id": conversation_id, "query": "what is kubernetes?"},
+        timeout=90,
+    )
+    print(vars(response))
+    assert response.status_code == requests.codes.ok
+    json_response = response.json()
+    json_response["query"] == "what is kubernetes?"
+    json_response["conversation_id"] == conversation_id
+    # assuming the response will be consistent
+    assert (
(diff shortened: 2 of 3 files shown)


tests/e2e/test_api.py
```diff
@@ -50,3 +50,64 @@ def test_invalid_question() -> None:
...
+    """Check the REST API /v1/query with POST HTTP method for same conversation id."""
+    response = client.post(
+        "/v1/query", json={"query": "what is kubernetes?"}, timeout=90
+    )
+    print(vars(response))
+    assert response.status_code == requests.codes.ok
+    json_response = response.json()
+    conversion_id = json_response["conversation_id"]
+
+    # sending second request with same conversation id
```
bparees, line 102: i don't see what this accomplishes/tests.  Nothing in this test would fail if you sent the second request without the previous conversation id.

bparees: Also it is probably premature to be trying to test the conversation history/reuse functionality of OLS.

sauagarwa: @bparees : Should we revisit the conversation_id test after the story OLS-110 is complete and mark this story as done?

bparees: yeah that might be best, there's some cache reuse implemented in OLS today, but probably not enough to be able to write a good test against.

but please delete this test from your PR, i don't want someone thinking we have coverage for the cache based on this test which doesn't really check anything

bparees:
/lgtm
/approve

the yaml portion of this will probably get dropped sooner rather than later as we're removing the yaml-specific logic in OLS, but in theory this should keep working regardless

bparees:
@sauagarwa your new tests appear to be failing the e2e suite.  You should be able to run it locally using the `scripts/test-e2e-standalone.sh` script

bparees:
/hold

need to see this pass the e2e job before it is merge.  The e2e job doesn't block merging currently, so manually holding the PR

scripts/test-e2e-standalone.sh
```diff
@@ -16,6 +16,15 @@ TMPDIR=$(mktemp -d)
 # ARTIFACT_DIR is defined when running in a prow job
 ARTIFACT_DIR=${ARTIFACT_DIR:-$TMPDIR}
 
+# Downloading index store
+export RAG_INDEX="https://github.com/ilan-pinto/lightspeed-rag-documents/releases/latest/download/local.zip"
+export RAG_INDEX_DIR="$ARTIFACT_DIR/vector-db/ocp-product-docs"
```
bparees, line 21: let's use a separate tmpdir for the RAG content, we don't want it in ARTIFACTS_DIR because we don't want it collected as part of the job artifacts (it's a waste of storage)

tests/e2e/test_api.py
```diff
@@ -63,3 +63,26 @@ def test_invalid_question():
...
+    json_response = response.json()
+    json_response["query"] == "what is kubernetes?"
+    json_response["conversation_id"] == conversation_id
+    # assuming the response will be consistent
+    assert (
+        "Kubernetes is an open source container orchestration tool"
+        in json_response["response"]
+    )
+    assert (
+        "The following response was generated without access to reference content:"
```
bparees, line 86: isn't the whole point of your updates above that OLS *should* have access to the reference content?

i'd expect a test that asks a question that is well answered by the RAG content and confirms that the answer we get back was generated based on RAG content.

sauagarwa: this is to test vector index is set. It checks for not in clause

bparees: sorry, missed the not

sauagarwa:
@bparees : The PR is on the state of not mergeable as it have do-not-merge/hold label. Can you remove the labels ?

bparees:
/hold cancel

for the future you are welcome to unhold it (via that command) if you've satisfied the criteria of the hold (in this case you have, by running and passing the e2e job)

PR #243, opened 2024-01-26
Handle INVALID,YAML as a possible reply from the QuestionValidator

syedriko:
## Description

Handle INVALID,YAML as a possible reply from the QuestionValidator

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing
Screenshot from 2024-01-25 21-23-39

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -112,9 +112,9 @@ Example Response:
 {INVALID},{NOYAML}
 
 Example Question:
-Why is the sky blue?
+Can you generate YAML for a Redis Transmogrifier?
 Example Response:
-{INVALID},{NOYAML}
+{INVALID},{YAML}
 
 Example Question:
 Can you help configure my cluster to automatically scale?
--- a/ols/src/query_helpers/question_validator.py
+++ b/ols/src/query_helpers/question_validator.py
@@ -55,11 +55,17 @@ class QuestionValidator(QueryHelper):
 
         self.logger.info(f"{conversation} response: {clean_response}")
 
-        if response["text"] not in ["INVALID,NOYAML", "VALID,NOYAML", "VALID,YAML"]:
+        if response["text"] not in [
+            "INVALID,NOYAML",
+            "INVALID,YAML",
+            "VALID,NOYAML",
+            "VALID,YAML",
+        ]:
             raise ValueError("Returned response did not match the expected format")
 
         # will return an array:
         # [INVALID,NOYAML]
+        # [INVALID,YAML]
         # [VALID,NOYAML]
         # [VALID,YAML]
         return clean_response.split(",")


ols/constants.py
```diff
@@ -112,9 +112,9 @@
 {INVALID},{NOYAML}
 
 Example Question:
-Why is the sky blue?
+Can you generate YAML for a Redis Transmogrifier?
 Example Response:
-{INVALID},{NOYAML}
+{INVALID},{YAML}
```
mpaulgreen, line 117: @syedriko I dont think we want to handle INVALID, YAML case. This signifies either prompt is not working or model is behaving indifferently.

syedriko: The calling code actually doesn't look at the YAML|NOYAML part if it sees INVALID https://github.com/openshift/lightspeed-service/blob/8b2c9ae05e20b2537a1f460a584473838c76a5b9/ols/app/endpoints/ols.py#L56 and then its well-behaved, but if we don't handle INVALID|YAML, a ValueError gets raised as if the model replied with complete gibberish.
May be the possibilities need to be {valid,{yaml|noyaml}}|{invalid} and we need a prompt to match. I fed the prompt that we have into ChatGPT and could get an INVALID,YAML responce, as in the screenshot ^^^.

If I add this to the prompt:
```
If you see that you need to reply with "INVALID,YAML" or ""INVALID,NOYAML"", just say INVALID
```
the replies match the {valid,{yaml|noyaml}}|{invalid} tree.

syedriko: This seems to be working well enough:
```
Instructions:
- You are a question classifying tool
- You are an expert in kubernetes and openshift
- Your job is to determine if a question is about kubernetes or openshift and to provide a one-word response
- If a question is not about kubernetes or openshift, answer with only the word INVALID
- If a question is about kubernetes or openshift, but not YAML, answer with VALID,NOYAML
- If a question is about kubernetes or openshift and YAML, answer with VALID,YAML
- Do not provide explanation, only respond with the chosen words

Example Question:
Can you make me lunch with ham and cheese?
Example Response:
INVALID

Example Question:
Why is the sky blue?
Example Response:
INVALID

Example Question:
Can you help configure my cluster to automatically scale?
Example Response:
VALID,NOYAML

Example Question:
please give me a vertical pod autoscaler configuration to manage my frontend deployment automatically.  Don't update the workload if there are less than 2 pods running.
Example Response:
VALID,YAML
```

mpaulgreen: The response type is VALID or INVALID signifying that it is an openshift question or not? If it is INVALID then it is not an openshift/k8s question then there is no significance of that. If existing code does not handle it then we probably need to fix it. Check with @bparees

bparees: > The response type is VALID or INVALID signifying that it is an openshift question or not?

that is the intent/desire, yes.

>  If it is INVALID then it is not an openshift/k8s question then there is no significance of that.

agree, if the question is deemed "invalid" then it does not matter if it is a "yaml" question or a "general" question.

syedriko: The way it's coded now, the invalid case breaks down into two, yaml|noyaml, and we don't handle one of them well. I'm suggesting to either fix that, or fix the prompt for the model to respond to invalid questions with just "invalid" and adjust the calling code accordingly.

mpaulgreen: I think we need to fix the code if not too big an effort.

PR #254, opened 2024-01-26
Use default logging level if nothing is specified in config file

tisnik:
## Description

- Previously the code "accepts" config with missing logging info, but then fail on the first call of any log method
- Now default logging level will be used

## Type of change

- [ ] Refactor
- [x] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Testing
- Unit tests and integration tests should pass

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -281,7 +281,7 @@ class LoggingConfig(BaseModel):
         """Initialize configuration and perform basic validation."""
         super().__init__()
         if data is None:
-            return
+            data = {}
 
         self.app_log_level = self._get_log_level(data, "app_log_level", "info")
         self.library_log_level = self._get_log_level(
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -1,6 +1,7 @@
 """Unit tests for the configuration model."""
 
 import io
+import logging
 import traceback
 from typing import TypeVar
 
@@ -402,3 +403,15 @@ def test_valid_config_file_with_redis() -> None:
     except Exception:
         print(traceback.format_exc())
         pytest.fail()
+
+
+def test_config_file_without_logging_config() -> None:
+    """Check how a configuration file without logging config is correctly initialized."""
+    # when logging configuration is not provided, default values will be used
+    # it means the following call should not fail
+    config.init_config("tests/config/config_without_logging.yaml")
+
+    # test if default values have been set
+    logging_config = config.ols_config.logging_config
+    assert logging_config.app_log_level == logging.INFO
+    assert logging_config.library_log_level == logging.WARNING
--- a/tests/config/config_without_logging.yaml
+++ b/tests/config/config_without_logging.yaml
@@ -1,0 +1,21 @@
+---
+LLMProviders:
+  - name: p1
+    url: "https://url1"
+    credentials_path: tests/config/secret.txt
+    models:
+      - name: m1
+        url: "https://murl1"
+OLSConfig:
+  conversation_cache:
+    type: redis
+    redis:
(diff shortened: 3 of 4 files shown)


onmete:
Makes sense.

/lgtm

bparees:
logging config should not be required, we should default to INFO level logging if the config file doesn't specify a level.

ols/app/models/config.py
```diff
@@ -281,7 +281,7 @@ def __init__(self, data: Optional[dict] = None):
         """Initialize configuration and perform basic validation."""
         super().__init__()
         if data is None:
-            return
+            data = {}
```
onmete, line 284: `data = data or {}`?

tests/unit/test_config.py
```diff
@@ -402,3 +402,11 @@ def test_valid_config_file_with_redis() -> None:
     except Exception:
         print(traceback.format_exc())
         pytest.fail()
+
+
+def test_config_file_without_logging_config() -> None:
+    """Check how a configuration file without logging config is correctly."""
```
onmete, line 408: "Check how a configuration file without logging config is correctly **initialized**."?

tests/unit/test_config.py
```diff
@@ -402,3 +402,11 @@ def test_valid_config_file_with_redis() -> None:
...
+    # when logging configuration is not provided, default values
+    # will be used
+    # it means the following call should not fail
```
onmete, line 411: can this be on the line above?

tests/unit/test_config.py
```diff
@@ -402,3 +402,11 @@ def test_valid_config_file_with_redis() -> None:
...
+    config.init_config("tests/config/config_without_logging.yaml")
```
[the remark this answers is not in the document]
onmete: I believe that happens in `tests/unit/test_models.py`.

ols/app/models/config.py
tisnik: actually I was thinking that later we might inform about that state (dunno why when logger is not setup at this stage ;)

tests/unit/test_config.py
tisnik: updated

tests/unit/test_config.py
tisnik: updated

ols/app/models/config.py
tisnik: WDYT @onmete ?

onmete: ok

tests/unit/test_config.py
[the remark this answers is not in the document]
bparees: it does, but i don't know why you wouldn't check it here as well.  this is a more complete flow of the config loading.

test_models.py only confirms that if you initialize an empty logconfig, it gets info as the default.

this only confirms that if you load an empty config, it doesn't fail.

you don't have anything that proves the connection between the two (that if you load an empty config, it initializes an empty logconfig to the default)

i'm reasonably certain I could introduce a code change that would break the default loglevel behavior in a way that didn't break either of the two tests you have here.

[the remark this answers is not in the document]
bparees: the test description even says `Check how a configuration file without logging config is correctly initialized.`

but all it checks is that an exception isn't thrown.  It doesn't validate that the config is "correctly initialized"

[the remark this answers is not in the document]
tisnik: agree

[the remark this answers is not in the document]
tisnik: added @bparees

PR #261, opened 2024-01-28
OLS-89: YAML configuration: redis conf from YAML, more validation

syedriko:
## Description

- Redis cache config now comes from YAML
- Added more YAML config validation

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-89](https://issues.redhat.com//browse/OLS-89)
- Closes #

## Testing

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -63,7 +63,9 @@ class ModelConfig(BaseModel):
         if self.name is None:
             raise InvalidConfigurationError("model name is missing")
         if self.url is not None and not _is_valid_http_url(self.url):
-            raise InvalidConfigurationError("model URL is invalid")
+            raise InvalidConfigurationError(
+                "model URL is invalid, only http:// and https:// URLs are supported"
+            )
 
 
 class ProviderConfig(BaseModel):
@@ -108,7 +110,9 @@ class ProviderConfig(BaseModel):
         if self.name is None:
             raise InvalidConfigurationError("provider name is missing")
         if self.url is not None and not _is_valid_http_url(self.url):
-            raise InvalidConfigurationError("provider URL is invalid")
+            raise InvalidConfigurationError(
+                "provider URL is invalid, only http:// and https:// URLs are supported"
+            )
         for v in self.models.values():
             v.validate_yaml()
 
@@ -144,7 +148,7 @@ class LLMProviders(BaseModel):
 class RedisCredentials(BaseModel):
     """Redis credentials."""
 
-    user: Optional[str] = None
+    username: Optional[str] = None
     password: Optional[str] = None
 
     def __init__(self, data: Optional[dict] = None) -> None:
@@ -152,22 +156,20 @@ class RedisCredentials(BaseModel):
         super().__init__()
         if not isinstance(data, dict):
             return
-        self.user = _get_attribute_from_file(data, "user_path")
+        self.username = _get_attribute_from_file(data, "username_path")
         self.password = _get_attribute_from_file(data, "password_path")
 
     def __eq__(self, other) -> bool:
         """Compare two objects for equality."""
         if isinstance(other, RedisCredentials):
-            return self.user == other.user and self.password == other.password
+            return self.username == other.username and self.password == other.password
         return False
 
     def validate_yaml(self) -> None:
         """Validate redis credentials."""
-        if (self.user is not None and self.password is None) or (
-            self.user is None and self.password is not None
-        ):
+        if self.username is not None and self.password is None:
             raise InvalidConfigurationError(
-                "both or neither user and password need to be specified for Redis"
+                "for Redis, if a username is specified, a password also needs to be specified"
             )
 
 
@@ -185,12 +187,24 @@ class RedisConfig(BaseModel):
         super().__init__()
         if data is None:
             return
-        self.host = data.get("host", None)
-        self.port = data.get("port", None)
+        self.host = data.get("host", constants.REDIS_CACHE_HOST)
+
+        try:
+            yaml_port = data.get("port", constants.REDIS_CACHE_PORT)
+            self.port = int(yaml_port)
+            if not (0 < self.port < 65536):
+                raise ValueError
+        except ValueError:
+            raise InvalidConfigurationError(
+                f"invalid Redis port {yaml_port}, valid ports are integers in the (0, 65536) range"
+            )
+
         self.max_memory = data.get("max_memory", constants.REDIS_CACHE_MAX_MEMORY)
+
         self.max_memory_policy = data.get(
             "max_memory_policy", constants.REDIS_CACHE_MAX_MEMORY_POLICY
         )
+
         credentials = data.get("credentials")
         if credentials is not None:
             self.credentials = RedisCredentials(credentials)
@@ -207,6 +221,22 @@ class RedisConfig(BaseModel):
             )
         return False
 
+    def validate_yaml(self) -> None:
+        """Validate Redis cache config."""
+        if (
+            self.max_memory_policy is not None
+            and self.max_memory_policy not in constants.REDIS_CACHE_MAX_MEMORY_POLICIES
+        ):
+            valid_polices = ", ".join(
+                str(p) for p in constants.REDIS_CACHE_MAX_MEMORY_POLICIES
+            )
+            raise InvalidConfigurationError(
+                f"invalid Redis max_memory_policy {self.max_memory_policy},"
+                f" valid policies are ({valid_polices})"
+            )
+        if self.credentials is not None:
+            self.credentials.validate_yaml()
+
 
 class MemoryConfig(BaseModel):
     """In-memory cache configuration."""
@@ -218,15 +248,27 @@ class MemoryConfig(BaseModel):
         super().__init__()
         if data is None:
             return
-        self.max_entries = data.get(
-            "max_entries", constants.IN_MEMORY_CACHE_MAX_ENTRIES
-        )
+
+        try:
+            self.max_entries = int(
+                data.get("max_entries", constants.IN_MEMORY_CACHE_MAX_ENTRIES)
+            )
+            if self.max_entries < 0:
+                raise ValueError
+        except ValueError:
+            raise InvalidConfigurationError(
+                "invalid max_entries for memory conversation cache,"
+                " max_entries needs to be a non-negative integer"
+            )
 
     def __eq__(self, other) -> bool:
         """Compare two objects for equality."""
         if isinstance(other, MemoryConfig):
             return self.max_entries == other.max_entries
         return False
+
+    def validate_yaml(self) -> None:
+        """Validate memory cache config."""
 
 
 class ConversationCacheConfig(BaseModel):
@@ -245,14 +287,22 @@ class ConversationCacheConfig(BaseModel):
         if self.type is not None:
             if self.type == constants.REDIS_CACHE:
                 if constants.REDIS_CACHE not in data:
-                    raise InvalidConfigurationError("redis configuration is missing")
+                    raise InvalidConfigurationError(
+                        "redis conversation cache type is specified,"
+                        " but redis configuration is missing"
+                    )
                 self.redis = RedisConfig(data.get(constants.REDIS_CACHE))
             elif self.type == constants.IN_MEMORY_CACHE:
                 if constants.IN_MEMORY_CACHE not in data:
-                    raise InvalidConfigurationError("memory configuration is missing")
+                    raise InvalidConfigurationError(
+                        "memory conversation cache type is specified,"
+                        " but memory configuration is missing"
+                    )
                 self.memory = MemoryConfig(data.get(constants.IN_MEMORY_CACHE))
             else:
-                raise InvalidConfigurationError("unknown conversation cache store")
+                raise InvalidConfigurationError(
+                    f"unknown conversation cache type: {self.type}"
+                )
 
     def __eq__(self, other) -> bool:
         """Compare two objects for equality."""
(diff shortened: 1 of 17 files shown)


ols/app/models/config.py
```diff
@@ -209,6 +220,15 @@ def __eq__(self, other):
             )
         return False
 
+    def validate_yaml(self) -> None:
+        """Validate Redis cache config."""
+        if self.port is not None:
+            if self.port <= 0 or self.port > 65535:
```
tisnik, line 226: this check is repeated in `__init__` so maybe some refactoring?

ols/app/models/config.py
```diff
@@ -209,6 +220,15 @@ def __eq__(self, other):
...
+                raise InvalidConfigurationError("invalid Redis port")
+        if self.max_memory_policy is not None:
+            if self.max_memory_policy not in constants.REDIS_CACHE_MAX_MEMORY_POLICIES:
```
tisnik, line 229: this check is repeated in `__init__` so maybe some refactoring?

ols/app/models/config.py
```diff
@@ -220,16 +240,31 @@ def __init__(self, data: Optional[dict] = None):
...
         """Compare two objects for equality."""
         if isinstance(other, MemoryConfig):
             return self.max_entries == other.max_entries
         return False
 
+    def validate_yaml(self) -> None:
+        """Validate memory cache config."""
+        if (
+            self.max_entries is None
+            or not isinstance(self.max_entries, int)
```
tisnik, line 263: is it needed? I mean, there are other "integer" config options, but these are just converted in `__init__` w/o further type checking

ols/app/models/config.py
```diff
@@ -268,7 +303,14 @@ def __eq__(self, other):
...
         """Validate conversation cache config."""
-        pass
+        if self.type is not None:
+            match self.type:
+                case constants.REDIS_CACHE:
+                    return self.redis.validate_yaml()
+                case constants.IN_MEMORY_CACHE:
+                    return self.memory.validate_yaml()
+                case _:
+                    raise InvalidConfigurationError("unknown conversation cache store")
```
tisnik, line 313: I'd specify the type in exception message, so the user will know that he (for example) made a typo...

ols/constants.py
```diff
@@ -192,6 +192,16 @@
 REDIS_CACHE_PORT = 6379
 REDIS_CACHE_MAX_MEMORY = "500mb"
 REDIS_CACHE_MAX_MEMORY_POLICY = "allkeys-lru"
+REDIS_CACHE_MAX_MEMORY_POLICIES = (
```
tisnik, line 195: nitpick - this might be set, not tuple

tisnik (commented):
looks ok, just few nitpicks

syedriko:
@tisnik I've addressed your comments, PTAL. For Redis auth, I used https://redis.io/commands/auth/ as a reference.

ols/src/cache/redis_cache.py
```diff
@@ -22,29 +18,34 @@ class RedisCache(Cache):
...
+                cls._instance.initialize_redis(config)
         return cls._instance
 
-    def initialize_redis(self) -> None:
+    def initialize_redis(self, config: RedisConfig) -> None:
         """Initialize the Redis client and logger.
 
         This method sets up the Redis client with custom configuration parameters.
         """
+        kwargs = {}
```
syedriko, line 34: @tisnik The idea here is to pass username or password only if they are defined in the passed-in config context, and leave them as defaults defined in StrictRedis(). Is this saying what I want it to say?

ols/constants.py
```diff
@@ -192,6 +192,20 @@
...
+REDIS_CACHE_MAX_MEMORY_POLICIES = frozenset(
+    {
+        "noeviction",
+        "allkeys-lru",
+        "allkeys-lfu",
+        "volatile-lru",
+        "volatile-lfu",
+        "allkeys-random",
+        "volatile-random",
+        "volatile-ttl",
```
bparees, line 204: we should only support LRU policies imho

bparees:
some nits to iterate on but the concept is sound
/approve

ols/app/models/config.py
```diff
@@ -217,16 +247,29 @@ def __init__(self, data: Optional[dict] = None):
...
 
     def __eq__(self, other):
         """Compare two objects for equality."""
         if isinstance(other, MemoryConfig):
             return self.max_entries == other.max_entries
         return False
 
+    def validate_yaml(self) -> None:
+        """Validate memory cache config."""
+        pass
```
bparees, line 271: no need for `pass` here as i'm told the comment is sufficient for "implementation"

ols/app/models/config.py
```diff
@@ -244,14 +287,22 @@ def __init__(self, data: Optional[dict] = None):
...
-                    raise InvalidConfigurationError("memory configuration is missing")
+                    raise InvalidConfigurationError(
+                        "memory conversation cache is specified,"
+                        " but memory configuration is missing"
+                    )
                 self.memory = MemoryConfig(data.get(constants.IN_MEMORY_CACHE))
             else:
-                raise InvalidConfigurationError("unknown conversation cache store")
+                raise InvalidConfigurationError(
+                    f"unknown conversation cache: {self.type}"
```
bparees, line 304:
```suggestion
                    f"unknown conversation cache type: {self.type}"
```

ols/app/models/config.py
```diff
@@ -265,6 +316,16 @@ def __eq__(self, other):
 
     def validate_yaml(self) -> None:
         """Validate conversation cache config."""
+        if self.type is not None:
```
bparees, line 319: if type is none isn't that also an error?

ols/src/llms/llm_loader.py
```diff
@@ -302,12 +301,12 @@ def _watson_llm_instance(self) -> LLM:
             "url": (
                 self.llm_params.get("url")
                 if self.llm_params.get("url") is not None
```
bparees, line 303: we should not gets the url from llm_params, we should only get it from the config object or from some hardcoded default.

ols/src/llms/llm_loader.py
```diff
@@ -302,12 +301,12 @@ def _watson_llm_instance(self) -> LLM:
...
-                else os.environ.get("WATSON_API_URL", None)
+                else self.provider_config.url
             ),
             "apikey": (
                 self.llm_params.get("apikey")
                 if self.llm_params.get("apikey") is not None
```
bparees, line 308: same here, we should not try to get this value from llm_params, only from the config.

ols/src/llms/llm_loader.py
```diff
@@ -331,14 +330,10 @@ def _watson_llm_instance(self) -> LLM:
 
         # WatsonX uses different parameter names
         llm_model = Model(
-            model_id=self.llm_params.get(
-                "model_id", os.environ.get("WATSON_MODEL", None)
-            ),
+            model_id=self.llm_params.get("model_id", None),
```
bparees, line 333: `model_id` should be the same value as `model name` from our config object, not come from llm_params

ols/src/llms/llm_loader.py
```diff
@@ -331,14 +330,10 @@ def _watson_llm_instance(self) -> LLM:
...
             credentials=creds,
             params=params,
-            project_id=self.llm_params.get(
-                "project_id", os.environ.get("WATSON_PROJECT_ID", None)
-            ),
+            project_id=self.llm_params.get("project_id", None),
```
bparees, line 336: we're going to have to sort this out when we start implementing watson support, but ultimately it needs to come from the config, it can't come from llm_params.

tests/config/with_openai_api_key.yaml
```diff
@@ -0,0 +1,16 @@
+---
+llm_providers:
+  - name: openai
+    url: "https://api.openai.com/v1"
+    credentials_path: tests/config/secret.txt
+    models:
+      - name: ibm/granite-13b-chat-v1
```
bparees, line 7: it doesn't really matter since it's just for a test, but to avoid confusion please use openai model names when defining openai providers (e.g. gpt35 in this case) because as written it makes me wonder if you're trying to test a case where someone accidentally mixed up a model from a different provider.

tests/config/without_openai_api_key.yaml
```diff
@@ -0,0 +1,16 @@
+---
+llm_providers:
+  - name: openai
+    url: "https://api.openai.com/v1"
+    credentials_path: tests/config/empty_secret.txt
+    models:
+      - name: ibm/granite-13b-chat-v1
```
bparees, line 7: same comment as above, please make this a valid openai model name

bparees (commented):
in general the only values we should pick up from llm_params as passed to llm_loader are values we expect OLS itself to need to override/set at runtime.  That should not include information we need the user to supply to use because it's specific to their environment/needs (e.g. modelids, custom api urls)

ols/app/models/config.py
syedriko: come to think of it, it is

syedriko: So `conversation_cache` is optional, but if it's there, it has to have a type and the config for the type.

bparees: we should make it required.  you can pick the in-memory one or redis, but you need to configure one.

ols/src/cache/in_memory_cache.py
```diff
@@ -4,8 +4,10 @@
 
 import threading
 from collections import deque
-from typing import Union
+from typing import TYPE_CHECKING, Union
 
+if TYPE_CHECKING:
+    from ols.app.models.config import MemoryConfig
```
bparees, line 10: is this a runtime performance optimization or what?

ols/src/llms/llm_loader.py
```diff
@@ -139,16 +156,12 @@ def _openai_llm_instance(self) -> LLM:
...
             raise e
         params: dict = {
-            "base_url": (
-                self.provider_config.url
-                if self.provider_config.url is not None
-                else "https://api.openai.com/v1"
-            ),
-            "openai_api_key": self.provider_config.credentials,
+            "base_url": self._get_llm_url("https://api.openai.com/v1"),
+            "openai_api_key": self._get_llm_credentials(),
```
bparees, line 160: it seems a bit odd and non-discoverable to me that we wouldn't just pass this value in as an arg to `_openai_llm_instance`.  Same for the url.

ols/src/llms/llm_loader.py
```diff
@@ -184,12 +197,8 @@ def _bam_llm_instance(self) -> LLM:
...
         # BAM Research lab
         creds = Credentials(
-            api_key=self.provider_config.credentials,
-            api_endpoint=(
-                self.provider_config.url
-                if self.provider_config.url is not None
-                else "https://bam-api.res.ibm.com"
-            ),
+            api_key=self._get_llm_credentials(),
+            api_endpoint=self._get_llm_url("https://bam-api.res.ibm.com"),
```
bparees, line 201: same comment here

tests/config/without_openai_api_key.yaml
bparees, line 1: the name on this test/file is odd.  It's not "without openai api key", it's an empty openai api key.

(do we also have a separate test where the file does not exist, and/or where the config does not specify a filepath?)

ols/src/cache/in_memory_cache.py
syedriko: if I make that import unconditional, I get
```
ruff . --fix --per-file-ignores=tests/*:S101
ols/src/cache/in_memory_cache.py:10:35: TCH001 Move application import `ols.app.models.config.MemoryConfig` into a type-checking block
```

tests/unit/llms/test_llm_loader.py
```diff
@@ -131,28 +130,16 @@ def mock_import(module, *args, **kwargs):
...
-
-
-@patch.dict(os.environ, {"OPENAI_API_KEY": ""})
 def test_constructor_openai_llm_instance_no_api_key():
     """Test the construction fo LLM instance for OpenAI when API key is not provided."""
-    _prepare_openapi_config()
-
     # no API key is provided so validation should fail
+    config.init_config("tests/config/without_openai_api_key.yaml")
     with pytest.raises(Exception, match="Did not find openai_api_key"):
```
bparees, line 137: this error message doesn't seem specific enough.  It seems like this test is checking what happens when the api key file is provided, but empty.

There are really 3 scenarios to test:
1) no filepath is provided
2) filepath is provided, but the file does not exist
3) filepath is provided, but the file is empty

we should ensure each of those scenarios has a unique/obvious error message specific to the situation.

ols/src/cache/in_memory_cache.py
bparees: weird.  @tisnik can you explain this?  It seems odd, especially that this seems to be the only place it's necessary to do this.

tisnik: it's runtime optimization indeed. The module is not used in the real code, "just" in type annotations. So in runtime it does not need to be imported.

tisnik: if `"MemoryConfig"` will be used instead (string), no conditional import will be needed

tests/unit/llms/test_llm_loader.py
syedriko: This tests the LLM loader, the "Did not find openai_api_key" message comes from langchain, https://github.com/langchain-ai/langchain/blob/cc173344737d4ac7ff96a04bfe2e93022b32d8b7/libs/core/langchain_core/utils/env.py#L42 most likely.
1. Credentials are optional in config: https://github.com/openshift/lightspeed-service/blob/00a3357045fd6aa7c3aaa242b30ccdfccf6cde60/ols/app/models/config.py#L48. There're many tests in https://github.com/openshift/lightspeed-service/blob/main/tests/unit/utils/test_config.py that don't specify credentials.
2. We have tests like that: https://github.com/openshift/lightspeed-service/blob/00a3357045fd6aa7c3aaa242b30ccdfccf6cde60/tests/unit/utils/test_config.py#L267 and https://github.com/openshift/lightspeed-service/blob/00a3357045fd6aa7c3aaa242b30ccdfccf6cde60/tests/unit/utils/test_config.py#L279
3. This case is not handled in the config code and but this test ^^^ does exactly that and langchain complains.

bparees: ok, thanks

PR #264, opened 2024-01-28
Added missing types into token handler

tisnik:
## Description

Added missing types into token handler

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Minor

### DIFF
--- a/ols/src/utils/token_handler.py
+++ b/ols/src/utils/token_handler.py
@@ -1,6 +1,7 @@
 """Utility to handle tokens."""
 
 import logging
+from abc import ABC
 
 import tiktoken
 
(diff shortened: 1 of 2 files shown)


ols/src/utils/token_handler.py
```diff
@@ -49,7 +50,7 @@ def tokens_to_text(self, tokens: list) -> str:
         return self._encoder.decode(tokens)
 
     def truncate_rag_context(
-        self, retrieved_nodes, max_tokens: int = 500
+        self, retrieved_nodes: list[Any], max_tokens: int = 500
```
bparees, line 53: there's no Node type defined?

tisnik: well it is list with this complex values:
```
            {
                "text": "a text text text text",
                "score": 0.6,
                "metadata": {"file_name": "doc1.pdf", "doc_link": "link1"},
            },
```

so it might be `list[dict[str, Any]]`. WDYT?

tisnik: actually no, I'm wrong. It's list of some object. Lemme find it

tisnik: ok so it indeed is/was unknown class. let's start with that one

tisnik: updated

PR #267, opened 2024-01-28
OLS-118: Ssar auth

nemerna:
## Description

Refactored the Code of the #66 ,
Enabled k8s SSAR authentication, Added propper logging according to the new logger usage,
added support for related Configuration in the Config loader.

made use of [FastAPI Dependencies](https://fastapi.tiangolo.com/tutorial/dependencies/#:~:text=Enforce%20security%2C%20authentication)

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-156](https://issues.redhat.com/browse/OLS-156)
- Closes #

## Testing
- have an Openshift/k8s Cluster connection and obtain a Token that represents your authenticated account.
Enable the Authentication (disabled by default) by using the config yaml:
Below is how to enable Authentication, and provide the k8s Cluster API .
```
Authentication:
     ols_auth_check: False
     k8s_cluster_api: "https://api.example.com:6443"
```
to enable ssl validation, add `verify_ssl: True`, for providing a certificated in case of self signed certs cluster,
use `k8s_ca_cert_path: "/home/user/ca.crt"`.
for a full config example, refer to examples/olsconfig.yaml

after configuring is done, run the code, and use the following curl to provide your token:

`curl -X 'POST' 'http://127.0.0.1:8080/v1/query' -H 'accept: application/json' -H "Authorization: Bearer $YOURTOKEN" -H 'Content-Type: application/json' -d '{"query": "write a deployment yaml for the mongodb image"}'`


examples/olsconfig.yaml
```diff
@@ -21,3 +21,9 @@ OLSConfig:
     library_log_level: warning
   default_provider: bam
   default_model: ibm/granite-13b-chat-v2
+
+Authentication:
```
onmete, line 25: This is probably a discussion for the other place, but why the top level keys are not in snake case (like the rest of the config)?

nemerna: @onmete 
Honestly no reason, i can patch hat.
simply i saw that the existing top level keys started with Capital letters so i followed the same.

if you think i have to patch, then i will.

onmete: @syedriko is there a reason behind that?

syedriko: That's how I found them. There was a discussion here https://github.com/openshift/lightspeed-service/pull/166#discussion_r1456459619 but tbh I didn't catch what was agreed on.

nemerna: Sure, will perform the Changes accordingly

examples/olsconfig.yaml
```diff
@@ -21,9 +21,13 @@ OLSConfig:
     library_log_level: warning
   default_provider: bam
   default_model: ibm/granite-13b-chat-v2
+Authentication:
+    ols_auth_check: False
```
bparees, line 25: i added a developer config setting for this in #278, you just need to wire it up to your logic.

examples/olsconfig.yaml
```diff
@@ -21,9 +21,13 @@ OLSConfig:
...
+    k8s_cluster_api: "https://api.example.com:6443"
+    k8s_ca_cert_path: "/Users/home/ca.crt"
+    verify_ssl: False
```
bparees, line 28: name this `skip_tls_verification` with a default of False so that the default behavior is that you get verification.

also ssl is an outdated term, it refers to an older version of TLS, so TLS is the preferred terminology.
```
Transport Layer Security (TLS) is the upgraded version of SSL that fixes existing SSL vulnerabilities. 
```

ols/app/models/config.py
```diff
@@ -445,6 +481,12 @@ def __init__(self, data: Optional[dict] = None):
         v = data.get("OLSConfig")
         if v is not None:
             self.ols_config = OLSConfig(v)
+        v = data.get("Authentication")
+        if v is not None:
+            self.authentication = AuthenticationConfig(v)
+        else:
+            self.authentication = AuthenticationConfig({})
```
bparees, line 488: we need to ensure that authentication is only turned off explicitly and deliberately.  We do not want someone to accidentally have a typo in their configuration which results in no authentication happening.

see my other comment about how i added an explicit developer-mode config setting to disable auth.  If that setting isn't enable(set to true), then missing/invalid authentication configuration should be an immediate error and failure.

examples/olsconfig.yaml
```diff
@@ -23,8 +23,13 @@ ols_config:
     lib_log_level: warning
   default_provider: bam
   default_model: ibm/granite-13b-chat-v2
+  authentication_config:
+    k8s_cluster_api: "https://api.example.com:6443"
+    k8s_ca_cert_path: "/Users/home/ca.crt"
+    skip_tls_verification: false
 dev_config:
   enable_dev_ui: true
+  disable_auth: false
```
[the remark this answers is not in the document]
nemerna: we will need a running k8s/ocp cluster for that
and the nto setup those parameters propeperly
  authentication_config:
    k8s_cluster_api: "https://api.example.com:6443"
    k8s_ca_cert_path: "/Users/home/ca.crt"
    skip_tls_verification: false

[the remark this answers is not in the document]
onmete: Yeah, I'm basically asking if it is something to have in contrib docs. But if so, I guess it would be worth of separate story.

tests/config/valid_config.yaml
```diff
@@ -34,3 +34,7 @@ ols_config:
   summarizer_model: m1
   validator_provider: p1
   validator_model: m1
+  yaml_provider: p2
+  yaml_model: m2
```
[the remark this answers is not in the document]
nemerna: resolved, looks like i made a mistake while resolving conflicts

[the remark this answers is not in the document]
onmete: yea, rebasing is killing me too these days :)

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,100 @@
...
+        configuration.verify_ssl = True
+
+        if config.ols_config.authentication_config.k8s_ca_cert_path:
+            configuration.ssl_ca_cert = (
+                config.ols_config.authentication_config.k8s_ca_cert_path
+            )
+        if config.ols_config.authentication_config.skip_tls_verification:
+            configuration.verify_ssl = False
+
+        k8s_client = kubernetes.client.ApiClient(configuration)
```
bparees, line 70: does the k8s client lib cache this object?  if not we need to be caching it so we're not constructing a new client on every request

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,100 @@
...
+        jd = json.loads(
+            '{"apiVersion":"authorization.k8s.io/v1","kind":"SelfSubjectAccessReview","spec":{"nonResourceAttributes":{"path":"/ols-access","verb":"get"}}}'
+        )
+        response = kubernetes.utils.create_from_dict(k8s_client, jd)
+        if response[0].status.allowed:
+            logger.info("passed authorization check")
```
bparees, line 76: should be debug level, also this message alone isn't too useful w/o some sort of request id, user id, or other information making it clear what request/user it pertained to

nemerna:
Closing the PR and moving (will be created again)

PR #272, opened 2024-01-29
Refactor logging configuration

onmete:
## Description

Changes here somehow disappeared from the https://github.com/openshift/lightspeed-service/pull/255

Refactors logging configuration to not configure each lib explicitly. As those are logging to root logger, we can just see its level according to our needs.

## Type of change

- [X] Refactor


tisnik (commented):
it looks ok

let's wait for CI ;)

PR #274, opened 2024-01-29
Fix integration tests: pydantic version

tisnik:
## Description

Fix integration tests: pydantic version

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


bparees:
/lgtm
/approve

followups:

1) pin our pydantic level (assuming that would help)
2) we should not be returning low level library errors back to the client like this.  We should be wrapping the error in our own validation error that we return

PR #281, opened 2024-01-30
OLS-256: Removed yaml category for Kubernetes queries

sauagarwa:
## Description
The fix addresses the following issues
a. BAM provider does not accurately return the expected classification response
b. Remove yaml categorization for Kubernetes related queries.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -5,19 +5,15 @@
 
 SUBJECT_VALID = "SUBJECT_VALID"
 SUBJECT_INVALID = "SUBJECT_INVALID"
-CATEGORY_YAML = "CATEGORY_YAML"
-CATEGORY_GENERIC = "CATEGORY_GENERIC"
-CATEGORY_UNKNOWN = "CATEGORY_UNKNOWN"
 POSSIBLE_QUESTION_VALIDATOR_RESPONSES = (
-    SUBJECT_VALID + "," + CATEGORY_YAML,
-    SUBJECT_VALID + "," + CATEGORY_GENERIC,
-    SUBJECT_INVALID + "," + CATEGORY_GENERIC,
+    SUBJECT_VALID,
+    SUBJECT_INVALID,
 )
 
 
 # templates
 SUMMARIZATION_TEMPLATE = """
-The following context contains several pieces of documentation. Please summarize the context for the user.
+The following context contains several pieces of documentation. Please answer the user's question based on this context.
 Documentation context:
 {context_str}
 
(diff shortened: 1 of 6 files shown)


ols/constants.py
```diff
@@ -107,7 +105,7 @@
 - If a question is not about kubernetes or openshift, answer with only the word {SUBJECT_INVALID}
 - If a question is about kubernetes or openshift, answer with the word {SUBJECT_VALID}
 - If a question is not about creating kubernetes or openshift yaml, answer with the word {CATEGORY_GENERIC}
-- If a question is about creating kubernetes or openshift yaml, add the word {CATEGORY_YAML}
+- If a question is about creating kubernetes or openshift yaml, add the word {CATEGORY_GENERIC}
```
bparees, line 108: i don't think we need to call out yaml at all. 

```
If a question is not about kubernetes or openshift, answer with only the word {SUBJECT_INVALID}
If a question is about kubernetes or openshift, answer with the word {SUBJECT_VALID}
```

those two alone should be sufficient in terms of categorization for now.

ols/src/query_helpers/question_validator.py
```diff
@@ -52,7 +52,9 @@ def validate_question(
         logger.info(f"{conversation} Validating query")
 
         bare_llm = LLMLoader(
-            self.provider, self.model, params={"min_new_tokens": 1, "max_new_tokens": 4}
+            self.provider,
+            self.model,
+            params={"min_new_tokens": 1, "max_new_tokens": 12},
```
bparees, line 57: does this work for both BAM and OpenAI?  At one point it worked for one but not the other

tests/integration/test_ols.py
```diff
@@ -61,14 +61,15 @@ def test_post_question_on_unexpected_payload() -> None:
...
+    print(response.json())
     assert response.json() == {
         "detail": [
             {
                 "input": "this is really not proper payload",
                 "loc": ["body"],
                 "msg": "Input should be a valid dictionary or object to extract fields from",
                 "type": "model_attributes_type",
-                "url": "https://errors.pydantic.dev/2.6/v/model_attributes_type",
+                "url": "https://errors.pydantic.dev/2.5/v/model_attributes_type",
```
bparees, line 72: we just pinned to v2.6, you may need to rebase

ols/constants.py
bparees: for now i think we have no need for categorization.  we can just do validation ("is this an openshift/k8s related question or not") and not worry about the sub-type of question (yaml-gen, generic, etc) for now.

ols/src/query_helpers/question_validator.py
sauagarwa: Yeah it was chopping off the CATEGORIZATION value. Increasing it to 12 worked for both. Since I have removed categorization now, I have reverted it back to 4

tests/integration/test_ols.py
sauagarwa: Done

PR #287, opened 2024-01-30
OLS-157: Rename conversation to conversation_id

onmete:
## Description

Rename conversation to conversation_id. Just conversation suggests it holds more than just id.

## Type of change

- [X] Refactor


tisnik (commented):
looks good

PR #288, opened 2024-01-30
OLS-157: Proper distribution of models unit tests

onmete:
## Description

Just moving tests around to reflect the main package (unit tests should mirror that).
As we have modules with the same name, we need to have some __init__.py in tests, to avoid tests naming collision.
No code is changed here.

## Type of change

- [x] Refactor

### DIFF
--- a/tests/unit/app/models/__init__.py
+++ b/tests/unit/app/models/__init__.py
@@ -1,0 +1,1 @@
+"""Init of tests/unit/app/models."""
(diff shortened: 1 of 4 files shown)


tisnik (commented):
ok

bparees:
/approve

merge conflict now

onmete:
But I can try at least:D

tisnik (commented):
it looks ok

PR #293, opened 2024-01-30
OLS-265: Implements an all-in-one container for testing

thoraxe:
## Description

Quick containerfile that makes it easy to test "e2e" with the rag data already present

## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

N/A

## Testing

### DIFF
--- a/Containerfile.rag
+++ b/Containerfile.rag
@@ -1,0 +1,24 @@
+# vim: set filetype=dockerfile
+
+# port, commands, and other details inherited from the base image
+ARG OLS_VERSION=latest
+FROM quay.io/openshift/lightspeed-service-api:${OLS_VERSION}
+
+ARG RAG_INDEX="https://github.com/ilan-pinto/lightspeed-rag-documents/releases/latest/download/local.zip"
+ARG APP_ROOT=/app-root
+USER 0
+
+RUN microdnf install -y unzip wget \
+  && mkdir -p vector-db \
+	&& cd vector-db \
+  && wget $RAG_INDEX \
+	&& unzip local.zip \
+	&& mv local ocp-product-docs \
+	&& rm -f local.zip \
+	&& microdnf remove -y unzip wget
+
+RUN chown -R 1001:0 ${APP_ROOT} && \
+    chmod -R g+rx ${APP_ROOT}
+
+# default user for Python app
+USER 1001
--- a/README.md
+++ b/README.md
@@ -64,6 +64,22 @@ in order to run the API service
 uvicorn ols.app.main:app --reload --port 8080
 ```
 
+### Optionally run with podman
+There is an all-in-one image that has the document store included already.
+
+1. Follow steps above to create your config yaml and your API key file(s). 
+1. Place your config yaml and your API key file(s) in a known location (eg:
+`/path/to/config`)
+1. Make sure your config yaml references the config folder for the path to your
+key file(s) (eg: `credentials_path: config/openai_api_key.txt`)
+1. Run the all-in-one-container. Example invocation:
+
+   ```sh
+    podman run -it --rm -v `/path/to/config:/app-root/config:Z \
+    -e OLS_CONFIG_FILE=/app-root/config/olsconfig.yaml -p 8080:8080 \
+    quay.io/openshift/lightspeed-api-service-rag:latest
+    ```
+
 ### Query the server
 
 To send a request to the server you can use the following curl command:


README.md
```diff
@@ -64,6 +64,20 @@ in order to run the API service
...
+### Optionally run with podman
+There is an all-in-one image that has the document store included already.
+
+1. Follow steps above to create your config yaml and your API key file(s). 
+2. Place your config yaml and your API key file(s) in a known location (eg:
+tmp/config)
+3. Run the all-in-one-container. Example invocation:
+
+   ```sh
+    podman run -it --rm -v `pwd`/tmp/config:/app-root/config:Z \
```
bparees, line 76: i'm not clear how this works, i think the instructions are a bit too handwavy about how to setup the config.yaml and api keys.

The config.yaml containers filepath refs to the apikeys.  The filepath ref in the config yaml either needs to be absolute (in which case it would need to be `/app-root/config/.....` given that's where you're mounting them here), or it needs to be relative to the working dir of the container which in this case is ???? but it certainly won't be `/app-root/config`.  Given you are setting OLS_CONFIG_FILE to `config/olsconfig.yaml` that would imply the working directory is `/app-root`.

Which would mean the api key refs in the config.yaml would need to be specified as `config/api-key.txt`

bparees: to make this somewhat easier on users, you could mount the host path to the same location inside the container.  Then you don't have to worry about the filepath ref in the config.yaml being valid on the host but not valid inside the container, or vice versa.

thoraxe: I can improve the instructions, but in this case some folder you create locally (somethingsomething/config) needs to be mounted into the container at `/app-root/config`. Your local config folder should have both your ols config yaml and your keys.

I have to test it, but my assumption is that the path to the API key files in the ols config yaml would just be `config/blah.txt`. Once I test, I'll improve the docs.

I'm not sure what you mean by "mount the host path to the same location inside the container" -- you can't mount anything directly to `/app-root` because it would completely overwrite all the contents of `/app-root`. You have to mount a folder into `/app-root` or into `/`. Podman/Docker do not support mounting files.

bparees: > in this case some folder you create locally (somethingsomething/config) needs to be mounted into the container at /app-root/config

why does it have to be mounted there, in particular?  The config.yaml can be mounted into the container anywhere you want, you just have to point the OLS_CONFIG_FILE env to the file.

> I'm not sure what you mean by "mount the host path to the same location inside the container" 

I mean that you can tell people to put config.yaml and their keys in a path like "/tmp/olsconfig" on their local host, and then mount it into the container at that same path ("/tmp/olsconfig"), that way the ols config yaml and the file references it contains will be valid both inside the container, and outside the container.  Assuming they use absolute paths in the file references.  For this to work of course the OLS_CONFIG_FILE env var has to be set correctly, but that's always true.

if they use relative paths, then it entirely depends on what OLS considers its working directory to be (which is not the same as which directory the config.yaml is in)

thoraxe: I tested the latest push locally, and it works fine with relative paths to the API key files.

If you're passionate about not mounting the config folder into the `/app-root` folder, I will test and update the readme again. LMK.

README.md
```diff
@@ -64,6 +64,23 @@ in order to run the API service
...
 
+### Optionally run with podman
+There is an all-in-one image that has the document store included already.
+
+1. Follow steps above to create your config yaml and your API key file(s). 
+1. Place your config yaml and your API key file(s) in a known location (eg:
+`/path/to/config``)
+1. Place your api key file(s) in the same folder (eg: `/path/to/config``)
+1. Make sure your config yaml references the config folder for the path to your
+key file(s) (eg: `credentials_path: config/openai_api_key.txt``)
```
bparees, line 75: are all these double backticks doing some magic i'm not familiar with?

README.md
```diff
@@ -64,6 +64,23 @@ in order to run the API service
 uvicorn ols.app.main:app --reload --port 8080
 ```
 
+### Optionally run with podman
+There is an all-in-one image that has the document store included already.
+
+1. Follow steps above to create your config yaml and your API key file(s). 
+1. Place your config yaml and your API key file(s) in a known location (eg:
+`/path/to/config``)
+1. Place your api key file(s) in the same folder (eg: `/path/to/config``)
```
bparees, line 73: the previous step already says to put your api key files in the same dir

Containerfile.rag
```diff
@@ -0,0 +1,20 @@
...
+RUN microdnf install -y unzip wget
+
+RUN mkdir -p vector-db/ocp-product-docs \
+  && cd vector-db \
+  && wget ${RAG_INDEX} \
+	&& unzip local.zip \
+	&& mv local vector-db \
+	&& rm -f local.zip
+
+RUN microdnf remove -y unzip wget
```
bparees, line 18: if you want this to be useful you need to do it in the same RUN command as the one where you installed it.  Otherwise you're just adding another layer to the image where these files are masked, but they're still in the lower layers.

README.md
thoraxe: user error. good catch.

README.md
thoraxe: thanks

Containerfile.rag
thoraxe: Ah, interesting. Will move.

Containerfile.rag
```diff
@@ -0,0 +1,37 @@
...
+  && wget $RAG_INDEX \
+	&& unzip local.zip \
+	&& mv local ocp-product-docs \
+	&& rm -f local.zip \
+	&& microdnf remove -y unzip wget
+
+RUN chown -R 1001:0 ${APP_ROOT} && \
+    chmod -R g+rx ${APP_ROOT}
+
+# activate virtualenv with workaround RHEL/CentOS 8+
```
bparees, line 22: i don't think you should need this:
https://github.com/moby/moby/issues/33242#issuecomment-302075518

were you getting failures without it?

also i'm not even sure our base image should be using a venv, there should be no reason we can't just install our deps into the system directories, so we should revisit the OLS base image containerfile too(but not in this PR of course).  It's kind of the whole idea of docker images.

Containerfile.rag
```diff
@@ -0,0 +1,37 @@
...
+    ENV="${APP_ROOT}/venv/bin/activate" \
+    PROMPT_COMMAND=". ${APP_ROOT}/venv/bin/activate" \
+    PATH="${APP_ROOT}/venv/bin:${PATH}"
+
+# default user for Python app
+USER 1001
+
+# Run the application
+EXPOSE 8080
+CMD ["uvicorn", "ols.app.main:app", "--host", "0.0.0.0", "--port", "8080"]
```
bparees, line 33: like ENV, i would expect this to be inherited from the base image you're extending:

https://www.reddit.com/r/docker/comments/sp4q7c/do_dockerfiles_inherit_their_parent_images/

Containerfile.rag
```diff
@@ -0,0 +1,37 @@
...
+ENV BASH_ENV="${APP_ROOT}/venv/bin/activate" \
+    ENV="${APP_ROOT}/venv/bin/activate" \
+    PROMPT_COMMAND=". ${APP_ROOT}/venv/bin/activate" \
+    PATH="${APP_ROOT}/venv/bin:${PATH}"
+
+# default user for Python app
+USER 1001
+
+# Run the application
+EXPOSE 8080
```
bparees, line 32: also inherited from the base image as i understand it.  this one is weird since there is no way to "unexpose" a port.  but it is what it is, as i understand it.

Containerfile.rag
thoraxe: this was copied from our existing container. Happy to remove it.

Containerfile.rag
thoraxe: Probably, but removing it makes it harder to look at our image and understand what it's doing. So, do you want "self documented" or do you want "inherited"? Granted the all-in-one is a hack, but the transparency is nice.

I don't have strong opinions either way, so I'm happy to defer to you.

Containerfile.rag
thoraxe: same as previous - it's also a form of documentation.

Containerfile.rag
bparees: I want inherited because if we change how we launch OLS, i don't want this image to break or need its dockerfile updated to re-align.  This image should only override/specify things that differ.

Containerfile.rag
bparees: same as previous, i wouldn't want this image to become misleading if we changed the port used in the base image.

Containerfile.rag
bparees: if you feel up to exploring it, i'd be interested to see if we can remove it from the ols image containerfile also.

Containerfile.rag
thoraxe: Good points.

thoraxe:
@bparees please re-check as I addressed your comments.

bparees:
/lgtm
/approve

@ilan-pinto fyi since you were asking about people consuming the rag content

note that the quay image won't actually exist until this merges:
https://github.com/openshift/release/pull/48233

PR #300, opened 2024-01-31
Remove return type in tests

onmete:
## Description

Removes return type from tests. It's not needed there.

## Type of change

- [x] Refactor


tisnik (commented):
ok

PR #305, opened 2024-01-31
OLS-261: Refactor SUID generator

tisnik:
## Description

- Refactor SUID generator
- We used to have three different `utils` which is too heavy for such a small project. That one utils can be refactored w/o problems.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

Related Issue #[OLS-261](https://issues.redhat.com//browse/OLS-261)


ols/utils/suid.py
```diff
@@ -0,0 +1,12 @@
...
+import uuid
+
+
+def get_suid() -> str:
+    """Generate a unique session ID (SUID) using UUID4.
+
+    Returns:
+        A unique session ID.
+    """
+    return str(uuid.uuid4().hex)
```
onmete, line 12: Just to be sure, we don't mind that uuid is not "partitioned" by `-`?
image

tisnik: we actually expect the 1st format

onmete: UUID without hyphens is still technically a valid UUID, but its format deviates from the standard and widely recognized hyphenated notation.
So what is the reason that we expect it without hyphens?

tisnik: - You need to ask original author Vishna
- Do you expect this change to be part of refactoring?
- Where do we get the info, that we must use UUID? We just need some session ID, that's it. Can be generated by other means

onmete: It is outside of this PR scope. We'll resolve this offline.

PR #310, opened 2024-01-31
OLS-157: Ensure endpoints have in/out models

onmete:
## Description

Adds request/response models for all endpoints.

I've removed the query from the response. It should be possible to track it via conversation_id. In theory, it might be a bit difficult to put those together in asynchronous scenarios, but hard to tell if it can happen.

I would like to hold it until https://github.com/openshift/lightspeed-service/pull/281 (don't want to fix tests that will be removed right after).

## Type of change

- [x] Refactor
- [x] New feature

## TODO

- [ ] refactor/update tests after PR-281

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -76,13 +76,9 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
             logger.info(
                 f"{conversation_id} - Query is not relevant to kubernetes or ocp, returning"
             )
-            response = str(
-                {
-                    "detail": {
-                        "response": "I can only answer questions about \
-            OpenShift and Kubernetes. Please rephrase your question"
-                    }
-                }
+            response = (
+                "I can only answer questions about OpenShift and Kubernetes. "
+                "Please rephrase your question"
             )
         case constants.SUBJECT_VALID:
             logger.info(
@@ -93,9 +89,14 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
                 docs_summarizer = DocsSummarizer(
                     provider=llm_request.provider, model=llm_request.model
                 )
-                response, _ = docs_summarizer.summarize(
+                llm_response, _ = docs_summarizer.summarize(
                     conversation_id, llm_request.query, previous_input
                 )
+                # TODO: There are some inconsitencies in the types to be solved.
+                # See the comment in `summarize` method in `docs_summarizer.py`
+                # Because of that, we are ignoring some type checks when we
+                # are creating response model.
+                response = llm_response.response
             except LLMConfigurationError as e:
                 raise HTTPException(
                     status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
@@ -115,9 +116,9 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
         config.conversation_cache.insert_or_append(
             user_id,
             conversation_id,
-            llm_request.query + "\n\n" + str(response or ""),
+            llm_request.query + "\n\n" + str(response or ""),  # type: ignore
         )
-    llm_response = LLMResponse(conversation_id=conversation_id, response=response)
+    llm_response = LLMResponse(conversation_id=conversation_id, response=response)  # type: ignore
     return llm_response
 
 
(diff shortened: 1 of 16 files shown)


tests/unit/app/models/test_models.py
```diff
@@ -1,38 +1,92 @@
...
+from ols.app.models.models import (
+    FeedbackRequest,
+    FeedbackResponse,
+    HealthResponse,
+    LLMRequest,
+    LLMResponse,
+)
+
+
+class TestLLM:
```
tisnik, line 12: what's the purpose of using classes there? Is not it bad behavior?

onmete: Maybe you meant "bad practice"? I'm not a dog. :P
And I wouldn't say so.
When there is a need to test more than one component/obj in a single unit test, I use those to clearly separate unit tests for different pieces.

Like
```python
class TestA:
  def test_a
  def test_b

class TestB:
  def test_a
  def test_b
```
instead of
```python
  def test_A_a
  def test_A_b
  def test_B_a
  def test_B_b
```
While I found it helpful to make unit tests visually distinct, it is just my preference.

Is there something in particular that is bothering you with this approach?

tisnik: so it is just to have separate "namespace"? do you think the tests will share some common values or so?

onmete: In general? Yes.
Here? They can share some inputs, like query, conversation_id, ...

ols/app/endpoints/ols.py
```diff
@@ -108,7 +106,7 @@ def conversation_request(llm_request: LLMRequest) -> LLMRequest:
             logger.info(
                 f"{conversation_id} - Query is not relevant to kubernetes or ocp, returning"
             )
-            llm_response.response = str(
+            response = str(
```
bparees, line 109: i guess you don't have to clean it up here, but as we refine the "response object" api, we definitely should not have a string field in our response object, into which we are putting json structured data.  If we want to have `response.detail.response="i can only answer question about......"` then we should define that overall structure of the response object.  (Note:  I do NOT think that is the overall structure we want for the response object)

ols/app/endpoints/ols.py
```diff
@@ -185,13 +183,16 @@ def conversation_request(llm_request: LLMRequest) -> LLMRequest:
         config.conversation_cache.insert_or_append(
             user_id,
             conversation_id,
-            llm_request.query + "\n\n" + str(llm_response.response or ""),
+            llm_request.query + "\n\n" + str(response or ""),
         )
+    llm_response = LLMResponse(
+        query=llm_request.query, conversation_id=conversation_id, response=response
```
bparees, line 189: to clarify my comment above:  in the finalized api i would expect each of these fields to represent primitive datatypes.  For fields that are strings, i'd expect the content to just be a raw string, not a json representation.

bparees (commented):
/approve

i'll leave it up to you if you want to address my comments about the structure of the actual data we are putting in the response field here, or as a follow up, but this PR establishes the sort of request/response object concept I was looking for as a starting point, thanks.

ols/app/endpoints/ols.py
bparees: let's change it now by dropping the `details` wrapper.  i don't think it will affect much existing code, if any (maybe the gradio ui) and it will avoid confusing people who are building on top of the api and think that we're going to have structured json embedded in these fields.

if it's more complicated than just dropping the details wrapper and putting the response directly into the response field as raw text, then i'm ok not fixing it here and handling it in a follow up, but even as a follow up i'd like to fix it immediately rather than fixing it later as we evolve the api.  I want our basic api to at least reflect the philosophy and intent of our api so everyone is on the same page going forward, and "json blobs in text fields" does not reflect that.

onmete: I'll do that as a separate PR right after this one.

onmete:
@bparees So I had to do the [follow-up](https://github.com/openshift/lightspeed-service/pull/310/commits/805c01f364c5bc9dc26d00e374992e019ec82fa5) in this PR (because of mypy).

The output I see in UI now:
image

Before it was:
image

tisnik (commented):
it looks ok so far. Have you tried to run e2e tests? That part might need changes too.

onmete:
Just finished fixing unit/integration tests. There were some leftovers from the "YAML thing" removal.

/test e2e

PR #311, opened 2024-01-31
Clean project deps

onmete:
## Description

I've removed some deps that seem extra or indirect. 

Eventually, we should pin everything we have in lock in pyproject.toml to be Konflux compliant. But now, it would be an overkill.

The app still works as usual. But I can't confirm if those aren't used in some corners. Like indexer.py?

## Type of change

- [x] Refactor

### DIFF
--- a/pdm.lock
+++ b/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev"]
 strategy = ["cross_platform", "inherit_metadata"]
 lock_version = "4.4.1"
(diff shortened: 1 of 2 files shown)


bparees:
> The app still works as usual. But I can't confirm if those aren't used in some corners. Like indexer.py?

indexer.py is deleted

but probably good to run the e2e:

/test e2e-ols-openai-gpt35

PR #314, opened 2024-01-31
Use UUID as conversation ID

tisnik:
## Description

Use UUID as conversation ID

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -5,7 +5,7 @@ from httpx import Client
 
 client = Client(base_url="http://localhost:8080")
 
-conversation_id = "0123456789abcdef0123456789abcdef"
+conversation_id = "12345678-abcd-0000-0123-456789abcdef"
 
 
 def test_readiness():
--- a/ols/src/cache/cache.py
+++ b/ols/src/cache/cache.py
@@ -1,8 +1,9 @@
 """Abstract class that is parent for all cache implementations."""
 
-import re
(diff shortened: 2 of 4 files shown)


tests/e2e/test_api.py
```diff
@@ -1,11 +1,13 @@
...
 
+import uuid
+
 import requests
 from httpx import Client
 
 client = Client(base_url="http://localhost:8080")
 
-conversation_id = "0123456789abcdef0123456789abcdef"
+conversation_id = str(uuid.uuid4())
```
bparees, line 10: it probably doesn't matter here but it's generally preferred to use constants for tests so we get consistent/reproducible results.  Or if you're using randomly generated values, you want to ensure you use a consistent seed for the generator so that the "Random" values are always the same when generated (only when testing of course!)

bparees (commented):
@onmete you had concerns about our uuid, does this address them or do you want to discuss them here?

tests/e2e/test_api.py
tisnik: ok not a problem to change, minute

onmete:
My concerns were addressed here.

/lgtm

tests/e2e/test_api.py
tisnik: updated @bparees

PR #322, opened 2024-02-01
OLS-261: Rename src/utils to rag

onmete:
## Description

The code in `token_handler.py` is not used anywhere at the moment. But it seems to be more RAG-related, hence the change.
The pleasant side-effect is we don't have spread utils in multiple places.

## Type of change

- [x] Refactor


tisnik (commented):
it looks ok from my perspective

ols/rag/token_handler.py
bparees, line 1: we're going to have other uses for the token handler beyond just counting tokens for RAG. We are also going to need to count tokens for conversation history and possibly other aspects of constraining the overall prompt length

onmete: Ok, so I would move it to `ols/utils/token_handler.py`, WDYT?

bparees: sounds good to me

tisnik (commented):
ok

PR #325, opened 2024-02-01
More unit tests for Gradio UI

tisnik:
## Description

More unit tests for Gradio UI

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

### DIFF
--- a/tests/unit/ui/test_gradio_ui.py
+++ b/tests/unit/ui/test_gradio_ui.py
@@ -1,10 +1,22 @@
 """Unit tests for gradioUI class."""
 
+import logging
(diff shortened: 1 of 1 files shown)


onmete:
/lgtm

but I would like to know if we should really cover with tests also code that serves solely dev purposes

bparees:
> but I would like to know if we should really cover with tests also code that serves solely dev purposes

we should not

PR #326, opened 2024-02-01
OLS-233: Use Optional type instead of | operator

onmete:
## Description

Seems we are using `Optional` instead of `something | None`. 

## Type of change

- [X] Refactor

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -18,7 +18,7 @@ def _is_valid_http_url(url: str) -> bool:
     ]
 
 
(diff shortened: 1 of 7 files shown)


tisnik (commented):
looks ok

bparees:
i had been pushing for the use of `|` syntax in PRs i reviewed because my understanding is that it is the newer/more modern syntax.

https://peps.python.org/pep-0655/ states:

```
Optional[] is too ubiquitous to deprecate, although use of it may fade over time in favor of the T|None notation specified by [PEP 604](https://peps.python.org/pep-0604/).
```

PR #330, opened 2024-02-02
OLS-116: Introducting Prometheus metrics

tisnik:
## Description

- Introducting Prometheus metrics
- Three example counters
- An example how to test metrics (counters)

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-116](https://issues.redhat.com//browse/OLS-116)

## Testing
- You can access the metrics as `/metrics` enpoint

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -7,6 +7,7 @@ from langchain.chains import LLMChain
 from langchain.prompts import PromptTemplate
 
 from ols import constants
+from ols.app import metrics
 from ols.app.models.models import LLMRequest
 from ols.src.llms.llm_loader import LLMConfigurationError, LLMLoader
 from ols.src.query_helpers.docs_summarizer import DocsSummarizer
@@ -53,15 +54,18 @@ def conversation_request(llm_request: LLMRequest) -> LLMRequest:
         question_validator = QuestionValidator(
             provider=llm_request.provider, model=llm_request.model
         )
+        metrics.llm_calls_total.inc()
         validation_result = question_validator.validate_question(
             conversation_id, llm_request.query
         )
     except LLMConfigurationError as e:
+        metrics.llm_calls_validation_errors_total.inc()
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail={"response": f"Unable to process this request because '{e}'"},
         )
     except Exception as validation_error:
+        metrics.llm_calls_failures_total.inc()
         logger.error("Error while validating question")
         logger.error(validation_error)
         raise HTTPException(
--- a/ols/app/metrics.py
+++ b/ols/app/metrics.py
@@ -1,0 +1,13 @@
+"""Prometheus metrics that are exposed by REST API."""
+
+from prometheus_client import Counter, make_asgi_app
+
+# just an example
+llm_calls_total = Counter("llm_calls_total", "LLM calls counter")
+llm_calls_failures_total = Counter("llm_calls_failures_total", "LLM calls failures")
+llm_calls_validation_errors_total = Counter(
+    "llm_validation_errors_total", "LLM validation errors"
+)
+
+# register metric
+metrics_app = make_asgi_app()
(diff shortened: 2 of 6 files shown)


onmete:
Nice and clean!

/lgtm

ols/app/metrics.py
```diff
@@ -0,0 +1,11 @@
+"""Prometheus metrics that are exposed by REST API."""
+
+from prometheus_client import Counter, make_asgi_app
+
+# just an example
+llm_calls_counter = Counter("llm_calls", "LLM calls counter")
+llm_calls_failures = Counter("llm_calls_failures", "LLM calls failures")
+llm_calls_validation_errors = Counter("llm_validation_errors", "LLM validation errors")
```
raptorsun, line 8: I suggest we follow naming conventions of [Prometheus metrics](https://prometheus.io/docs/practices/naming/) to name these metrics like:
- llm_calls_total
- llm_calls_failures_total
- llm_validation_errors_total
we add a "_total" for these 3 are unit-less accumulating counters.

tisnik: ok, on it. Thank you

ols/app/endpoints/ols.py
```diff
@@ -53,15 +54,18 @@ def conversation_request(llm_request: LLMRequest) -> LLMRequest:
...
             conversation_id, llm_request.query
         )
     except LLMConfigurationError as e:
+        metrics.llm_calls_validation_errors_total.inc()
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail={"response": f"Unable to process this request because '{e}'"},
         )
     except Exception as validation_error:
+        metrics.llm_calls_failures_total.inc()
```
bparees, line 68: Seems like these two metric names are reversed?  (validation errors and call failures)

or they are both validation failures and we need a better name for both of them that distinguishes them

(i realize they are just examples and may not even stick around, but they might, or people might copy them, so lets get the foundation right)

ols/app/metrics.py
```diff
@@ -0,0 +1,13 @@
+"""Prometheus metrics that are exposed by REST API."""
+
+from prometheus_client import Counter, make_asgi_app
+
+# just an example
```
bparees, line 5: just an example but it will establish the precedent that people will copy, so please provide comments explaining what the metrics are intended represent.  In particular it's not clear what the intended distinction is between "call failures" and "validation errors"

bparees (commented):
@tisnik some small items to follow up on please

ols/app/endpoints/ols.py
raptorsun: What I understand is the exception LLMConfigurationError is raised when validating LLM configuration, and the most probable way to raise the generic Exception is a unsuccessful query to OpenAI/BAM API. So they are named as validation_error and call_failures. 

(unrelated to this PR) It may be better to validate the LLM configuration just after loading the configuration file. We can use a Gauge to show whether the configuration is valid, for example `configuration_valid(component="llm") 1` .

bparees: > What I understand is the exception LLMConfigurationError is raised when validating LLM configuration, and the most probable way to raise the generic Exception is a unsuccessful query to OpenAI/BAM API. So they are named as validation_error and call_failures.

Failure to call the backend should have a clear error type that is distinct from any other failure type

I think this is fundamentally less about the metrics you defined and more about the fact that our error handling+reporting still has some maturing to do.  At a high level there are 3 categories of failure:

1) an external service we rely on failed (BAM/Watson/OpenAI is down/unavailable/timedout or otherwise errored for a reason that is out of our control)
2) An internal error occurred (we got an unexpected exception trying to process the request.  e.g. we failed to parse the response from the llm, we ran out of memory, something else surprising and bad happened)
3) A user input validation error occurred: the user's request was not valid (e.g. api request object was badly formatted, or the api request asked for a model/provider that doesn't exist, or the credentials used to talk to the llm are invalid/rejected)

>  It may be better to validate the LLM configuration just after loading the configuration file.

i think we are doing that validation during load already(@syedriko can confirm).  I agree that we should not be getting runtime errors in the request flow due to an invalid ols config yaml.  However individual api requests can specify a model+provider so it's possible that the api request specifies a model or provider that does not exist in the configuration yaml.

PR #331, opened 2024-02-02
OLS-157: Do request validation in pydantic model

onmete:
## Description

Use pydantic abilities to validate its own models instead of us doing the same work. It still raises 422, but the json is a bit different.

## Type of change

- [X] Refactor


bparees:
much cleaner, thanks

/lgtm
/approve

PR #339, opened 2024-02-03
OLS-89: LLM loader configuration cleanup

syedriko:
## Description

- Addressed outstanding review items from https://github.com/openshift/lightspeed-service/pull/261
- In the LLM loader, distinguish between providers that are not in configuration and those that are, but not supported (openai, bam, etc are supported).
- Cleaned up test configurations for LLM loader tests

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # https://issues.redhat.com/browse/OLS-89
- Closes #

## Testing


tests/unit/llms/test_llm_loader.py
```diff
@@ -62,43 +64,44 @@ def test_constructor_no_model():
...
+                    "name": test_provider,
+                    "models": [
+                        {
+                            "name": constants.GRANITE_13B_CHAT_V1,
+                        }
+                    ],
+                }
+            ]
+        }
+    )
```
bparees, line 89: this seems harder to read and maintain than the previous code that called the individual constructors was, what is the motivation for the refactor?  (same question/comment for the other spot where you did this)

syedriko: The way config was contructed in this code was breaking (the thing layer of) config encapsulation. For example, I couldn't quite tell what this
```
    provider_config = ProviderConfig()
    provider_config.models = {constants.GRANITE_13B_CHAT_V1: "mock model"}
    config.llm_config.providers = {provider: provider_config}
```
was even doing, `models` is a `dict[str, ModelConfig]`, it was plugging in a string literal instead of a ModelConfig.
Some other cases were impossible according to the config model, such as providers with no models raise an InvalidConfigurationError("no models configured for provider...").

PR #351, opened 2024-02-05
OLS-110: Added conversation history to OLS

xrajesh:
## Description

Added cache as Conversation List instead of String
Updated the QueryEngine to make use of Conversation History

## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue (https://issues.redhat.com/browse/OLS-110)

## Testing

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -9,6 +9,7 @@ from langchain.prompts import PromptTemplate
 from ols import constants
 from ols.app import metrics
 from ols.app.models.models import LLMRequest, LLMResponse
+from ols.src.cache.conversation import Conversation
 from ols.src.llms.llm_loader import LLMConfigurationError, LLMLoader
 from ols.src.query_helpers.docs_summarizer import DocsSummarizer
 from ols.src.query_helpers.question_validator import QuestionValidator
@@ -96,7 +97,13 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
                 # See the comment in `summarize` method in `docs_summarizer.py`
                 # Because of that, we are ignoring some type checks when we
                 # are creating response model.
-                response = llm_response.response
+                if isinstance(llm_response, str):
+                    response = llm_response
+                else:
+                    if llm_response.response:
+                        response = llm_response.response
+                    else:
+                        """"""
             except LLMConfigurationError as e:
                 raise HTTPException(
                     status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
@@ -113,10 +120,9 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
                 )
 
     if config.conversation_cache is not None:
+        conversation = Conversation(llm_request.query, str(response or ""))  # type: ignore
         config.conversation_cache.insert_or_append(
-            user_id,
-            conversation_id,
-            llm_request.query + "\n\n" + str(response or ""),  # type: ignore
+            user_id, conversation_id, conversation
         )
     llm_response = LLMResponse(conversation_id=conversation_id, response=response)  # type: ignore
     return llm_response
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -13,9 +13,11 @@ POSSIBLE_QUESTION_VALIDATOR_RESPONSES = (
 
 # templates
 SUMMARIZATION_TEMPLATE = """
-The following context contains several pieces of documentation. Please answer the user's question based on this context.
-Documentation context:
-{context_str}
+Instructions:
+- You are an expert in Kubernetes and OpenShift.
+- Respond to questions about topics other than Kubernetes and OpenShift with: "I can only answer questions about Kubernetes and OpenShift"
+- Refuse to participate in anything that could harm a human.
+- Use the previous chat history, and the context, to interact and help the user."
 
 User query:
 {query_str}
@@ -23,6 +25,7 @@ User query:
 Summary:
 
 """
+
 
 SUMMARY_TASK_BREAKDOWN_TEMPLATE = (
     """
@@ -167,6 +170,7 @@ REDIS_CACHE_PORT = 6379
 REDIS_CACHE_MAX_MEMORY = "500mb"
 REDIS_CACHE_MAX_MEMORY_POLICY = "allkeys-lru"
 REDIS_CACHE_MAX_MEMORY_POLICIES = frozenset({"allkeys-lru", "volatile-lru"})
+REDIS_MAX_RETRY = 5
 
 # provider and model roles
 PROVIDER_MODEL_ROLES = frozenset({"default"})
--- a/ols/src/cache/cache.py
+++ b/ols/src/cache/cache.py
@@ -1,8 +1,8 @@
 """Abstract class that is parent for all cache implementations."""
 
 from abc import ABC, abstractmethod
-from typing import Optional
 
+from ols.src.cache.conversation import Conversation
 from ols.utils.suid import check_suid
 
 
@@ -40,7 +40,7 @@ class Cache(ABC):
         return f"{user_id}{Cache.COMPOUND_KEY_SEPARATOR}{conversation_id}"
 
     @abstractmethod
-    def get(self, user_id: str, conversation_id: str) -> Optional[str]:
+    def get(self, user_id: str, conversation_id: str) -> list[Conversation] | None:
         """Abstract method to retrieve a value from the cache.
 
         Args:
@@ -52,7 +52,9 @@ class Cache(ABC):
         """
 
     @abstractmethod
-    def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> None:
+    def insert_or_append(
+        self, user_id: str, conversation_id: str, value: Conversation
+    ) -> None:
         """Abstract method to store a value in the cache.
 
         Args:
--- a/ols/src/cache/in_memory_cache.py
+++ b/ols/src/cache/in_memory_cache.py
@@ -4,10 +4,12 @@ from __future__ import annotations
 
 import threading
 from collections import deque
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from ols.app.models.config import MemoryConfig
+    from ols.src.cache.conversation import Conversation
+
 from ols.src.cache.cache import Cache
 
 
@@ -29,9 +31,9 @@ class InMemoryCache(Cache):
         """Initialize the InMemoryCache."""
         self.capacity = config.max_entries
         self.deque: deque[str] = deque()
-        self.cache: dict[str, str] = {}
+        self.cache: dict[str, list[Conversation]] = {}
 
-    def get(self, user_id: str, conversation_id: str) -> Optional[str]:
+    def get(self, user_id: str, conversation_id: str) -> list[Conversation] | None:
         """Get the value associated with the given key.
 
         Args:
@@ -50,7 +52,9 @@ class InMemoryCache(Cache):
         self.deque.appendleft(key)
         return self.cache[key]
 
-    def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> None:
+    def insert_or_append(
+        self, user_id: str, conversation_id: str, value: Conversation
+    ) -> None:
         """Set the value if a key is not present or else simply appends.
 
         Args:
@@ -65,9 +69,9 @@ class InMemoryCache(Cache):
                 if len(self.deque) == self.capacity:
                     oldest = self.deque.pop()
                     del self.cache[oldest]
-                self.cache[key] = value
+                values = [value]
+                self.cache[key] = values
             else:
                 self.deque.remove(key)
-                old_value = self.cache[key]
-                self.cache[key] = old_value + "\n" + value
+                self.cache[key].append(value)
             self.deque.appendleft(key)
--- a/ols/src/cache/redis_cache.py
+++ b/ols/src/cache/redis_cache.py
@@ -1,12 +1,20 @@
 """Cache that uses Redis to store cached values."""
 
+import logging
 import threading
-from typing import Optional
 
 import redis
 
+from ols import constants
 from ols.app.models.config import RedisConfig
 from ols.src.cache.cache import Cache
+from ols.src.cache.conversation import Conversation
+
+logger = logging.getLogger(__name__)
+
+
+class RedisMaxRetryError(Exception):
+    """RedisMaxRetryError."""
 
 
 # TODO
@@ -47,7 +55,7 @@ class RedisCache(Cache):
         self.redis_client.config_set("maxmemory", config.max_memory)
         self.redis_client.config_set("maxmemory-policy", config.max_memory_policy)
 
-    def get(self, user_id: str, conversation_id: str) -> Optional[str]:
+    def get(self, user_id: str, conversation_id: str) -> list[Conversation] | None:
         """Get the value associated with the given key.
 
         Args:
@@ -58,10 +66,13 @@ class RedisCache(Cache):
             The value associated with the key, or None if not found.
         """
         key = super().construct_key(user_id, conversation_id)
+        value = self.redis_client.get(key)
+        # GET operation might return Awaitable value .
+        return value if value else None
 
-        return self.redis_client.get(key)
-
-    def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> None:
+    def insert_or_append(
+        self, user_id: str, conversation_id: str, value: Conversation
+    ) -> None:
         """Set the value associated with the given key.
 
         Args:
@@ -70,14 +81,33 @@ class RedisCache(Cache):
             value: The value to set.
 
         Raises:
-            OutOfMemoryError: If item is evicted when Redis allocated
-                memory is higher than maxmemory.
+            RedisConnectionError: If item is unable to update after REDIS_MAX_RETRY.
         """
         key = super().construct_key(user_id, conversation_id)
-
         old_value = self.get(user_id, conversation_id)
-        with self._lock:
-            if old_value:
-                self.redis_client.set(key, old_value + "\n" + value)
-            else:
-                self.redis_client.set(key, value)
+        retry_count = 0
+        while retry_count < constants.REDIS_MAX_RETRY:
+            logger.debug("Updating redis cache ")
+            with self._lock:
+                if old_value:
+                    old_value.append(value)
+                    try:
+                        self.redis_client.set(key, old_value)
+                        break
+                    except Exception as redis_err:
+                        logger.error(
+                            f"Exception during updating the cache: {redis_err}"
+                        )
+                        retry_count += 1
+                else:
+                    values = [value]
+                    try:
+                        self.redis_client.set(key, values)
+                        break
+                    except Exception as redis_err:
+                        logger.error(
+                            f"Exception during creating the cache: {redis_err}"
+                        )
+                        retry_count += 1
+
+        raise RedisMaxRetryError("Updating cache failed after max retries")
--- a/ols/src/query_helpers/chat_history.py
+++ b/ols/src/query_helpers/chat_history.py
@@ -1,0 +1,46 @@
+"""Produce chat history object using list of Conversations."""
+
+from typing import Optional
+
+from langchain.schema import ChatMessage as langchainCM
+from llama_index.llms import ChatMessage, MessageRole
+
+from ols.src.cache.conversation import Conversation
+
+
+def get_llama_index_chat_history(
+    conversations: Optional[list[Conversation]],
+) -> list[ChatMessage]:
+    """Produce list of ChatMessage object using list of Conversations for llama_index."""
+    chat_history: list[Conversation] = []
+    if conversations is None or len(conversations) < 1:
+        return chat_history
+    for conversation in conversations:
+        user_chat_message = ChatMessage(
+            role=MessageRole.USER, content=conversation.user_message
+        )
+        chat_history.append(user_chat_message)
+        assistant_chat_message = ChatMessage(
+            role=MessageRole.ASSISTANT, content=conversation.assistant_message
+        )
+        chat_history.append(assistant_chat_message)
+    return chat_history
+
+
+def get_langchain_chat_history(
+    conversations: Optional[list[Conversation]],
+) -> list[langchainCM]:
+    """Produce list of ChatMessage object using list of Conversations for langchain."""
+    chat_history: list[Conversation] = []
+    if conversations is None or len(conversations) < 1:
+        return chat_history
+    for conversation in conversations:
+        user_chat_message = langchainCM(
+            role=MessageRole.USER, content=conversation.user_message
+        )
+        chat_history.append(user_chat_message)
+        assistant_chat_message = langchainCM(
+            role=MessageRole.ASSISTANT, content=conversation.assistant_message
+        )
+        chat_history.append(assistant_chat_message)
+    return chat_history
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -4,15 +4,24 @@ import logging
 from typing import Any, Optional
 
 import llama_index
+from langchain.chains import LLMChain
+from langchain.schema import ChatMessage
 from llama_index import ServiceContext, StorageContext, load_index_from_storage
 from llama_index.postprocessor import BaseNodePostprocessor
 from llama_index.prompts import PromptTemplate
+from llama_index.chat_engine.types import BaseChatEngine
+from llama_index.prompts import PromptTemplate as LlamaTemplate
 from llama_index.response.schema import Response
 from llama_index.schema import NodeWithScore, QueryBundle
 
 from ols import constants
 from ols.src.llms.llm_loader import LLMLoader
 from ols.src.query_helpers import QueryHelper
+from ols.src.query_helpers.chat_history import (
+    get_langchain_chat_history,
+    get_llama_index_chat_history,
+)
+from ols.src.query_helpers.constants import summary_prompt_for_langchain
 from ols.utils import config
 
 logger = logging.getLogger(__name__)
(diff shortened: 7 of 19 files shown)


ols/src/cache/redis_cache.py
```diff
@@ -74,10 +78,12 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
                 memory is higher than maxmemory.
         """
         key = super().construct_key(user_id, conversation_id)
-
         old_value = self.get(user_id, conversation_id)
         with self._lock:
             if old_value:
-                self.redis_client.set(key, old_value + "\n" + value)
+                old_value.append(value)
```
bparees, line 84: is this list guaranteed to stay ordered through redis storage + retrieval?

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -80,20 +89,21 @@ def summarize(
...
                 logger.info(f"{conversation} Submitting summarization query")
-                summary = query_engine.query(query)
+
+                chat_engine = index.as_chat_engine(
+                    service_context=service_context,
+                    chat_history=chat_history,
+                    chat_mode="context",
+                    system_prompt=summarization_template,
+                )
+                summary = chat_engine.chat(query)
```
bparees, line 106: this is ok for now but sooner rather than later we're going to drop the use of llamaindex for our RAG-based query flows and instead construct the prompt(with rag content) ourselves, specifically as part of https://issues.redhat.com/browse/OLS-104

So eventually i expect this codepath to look more like the no-RAG flow below.

bparees (commented):
this seems to be lacking tests that really confirm that conversation history is retrieved and injected into the prompt that goes to the llm.

xrajesh:
> this seems to be lacking tests that really confirm that conversation history is retrieved and injected into the prompt that goes to the llm.

Yes , will have a new PR . Wanted to get acceptance on the flow .

bparees:
> Yes , will have a new PR . Wanted to get acceptance on the flow .

tests for new function should go in the same PR that introduces the new function, so please add them to this PR.

The flow looks ok to me, aside from my comment about the use of llamaindex vs langchain which will change in the future.

ols/src/cache/cache.py
```diff
@@ -1,8 +1,9 @@
 """Abstract class that is parent for all cache implementations."""
 
 from abc import ABC, abstractmethod
-from typing import Union
+from typing import List, Union
```
tisnik, line 4: just use `list` please

ols/src/cache/in_memory_cache.py
```diff
@@ -4,10 +4,12 @@
 
 import threading
 from collections import deque
-from typing import TYPE_CHECKING, Union
+from typing import TYPE_CHECKING, List, Union
```
tisnik, line 7: dtto `list`

ols/src/cache/in_memory_cache.py
```diff
@@ -65,9 +71,10 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
                 if len(self.deque) == self.capacity:
                     oldest = self.deque.pop()
                     del self.cache[oldest]
-                self.cache[key] = value
+                values = []
+                values.append(value)
```
tisnik, line 75: `values = [value]` perhaps?

ols/src/cache/redis_cache.py
```diff
@@ -1,12 +1,13 @@
 """Cache that uses Redis to store cached values."""
 
 import threading
-from typing import Union
+from typing import List, Union
```
tisnik, line 4: dtto

ols/src/cache/redis_cache.py
tisnik: yes

ols/src/cache/redis_cache.py
```diff
@@ -74,10 +78,12 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
-
         old_value = self.get(user_id, conversation_id)
         with self._lock:
             if old_value:
-                self.redis_client.set(key, old_value + "\n" + value)
+                old_value.append(value)
+                self.redis_client.set(key, old_value)
             else:
-                self.redis_client.set(key, value)
+                values = []
```
tisnik, line 87: `values = [value]`

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,44 @@
+"""Produce chat history object using list of Conversations."""
+
+from typing import List
```
tisnik, line 3: `list`

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,44 @@
...
+
+from ols.src.cache.conversation import Conversation
+
+
+def get_llama_index_chat_history(
+    conversations: List[Conversation],
+) -> List[ChatMessage]:
+    """Produce list of ChatMessage object using list of Conversations for llama_index."""
+    chat_history: List[Conversation] = []
+    if conversations is None or len(conversations) < 1:
```
tisnik, line 16: can it be `None`?

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,44 @@
...
+            role=MessageRole.ASSISTANT, content=conversation.assistant_message
+        )
+        chat_history.append(assistant_chat_message)
+    return chat_history
+
+
+def get_langchain_chat_history(conversations: List[Conversation]) -> List[langchainCM]:
+    """Produce list of ChatMessage object using list of Conversations for langchain."""
+    chat_history: List[Conversation] = []
+    if conversations is None or len(conversations) < 1:
```
tisnik, line 33: dtto, can it be `None`?

tests/integration/test_ols.py
```diff
@@ -327,17 +327,21 @@ def test_post_question_on_noyaml_response_type() -> None:
...
-                            "query": "test query",
-                        },
-                    )
-                    print(response)
-                    assert response.status_code == requests.codes.ok
-                    assert (
-                        "The following response was generated without access to reference content:"
-                        in response.json()["response"]
-                    )
+                    ml.invoke = MagicMock(return_value={"text": "success"})
```
tisnik, line 330: you can use `patch` as in other tests

tisnik (commented):
some comments and nitpicks

ols/src/cache/redis_cache.py
```diff
@@ -74,10 +78,12 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
+                values.append(value)
+                self.redis_client.set(key, values)
```
raptorsun, line 89: shall we care about whether redis has saved our conversations?
If missing a message in the conversation, will it affect greatly the quality of the answer to user's questions?

ols/constants.py
```diff
@@ -13,14 +13,17 @@
...
 # templates
 SUMMARIZATION_TEMPLATE = """
-The following context contains several pieces of documentation. Please answer the user's question based on this context.
-Documentation context:
-{context_str}
+Instructions:
+- You are an expert in Kubernetes and OpenShift.
+- Respond to questions about topics other than Kubernetes and OpenShift with: "I can only answer questions about Kubernetes and OpenShift"
+- Refuse to participate in anything that could harm a human.
+- Use the previous chat history, and the context, to interact and help the user."
```
onmete, line 20: Is 
_You are an expert in Kubernetes and OpenShift._
instruction necessary? What is its outcome? Some confident tone in conversation?

Similarly the 3. 4. instruction. Aren't those encoded in the llms already?

ols/src/cache/cache.py
```diff
@@ -40,7 +41,9 @@ def construct_key(user_id: str, conversation_id: str) -> str:
         return f"{user_id}{Cache.COMPOUND_KEY_SEPARATOR}{conversation_id}"
 
     @abstractmethod
-    def get(self, user_id: str, conversation_id: str) -> Union[str, None]:
+    def get(
+        self, user_id: str, conversation_id: str
+    ) -> Union[List[Conversation], None]:
```
onmete, line 46: It is basically `Optional[List[Conversation]]`, but it somehow feels off using the optional type for return, wdyt?

ols/src/query_helpers/chat_history.py
xrajesh: I dont expect to be None , but even if we get a None we will return a empty History. You okay with that @tisnik

ols/src/query_helpers/chat_history.py
xrajesh: ditto

ols/src/query_helpers/chat_history.py
tisnik: I meant, if it can be None, the type of parameter should be `Optional[...]`.

ols/src/cache/redis_cache.py
xrajesh: @raptorsun Agree , but  this is how it was , can we open a bug to address this ?

ols/constants.py
bparees: not just tone, it frames the whole way the LLM attempts to answer the question and what knowledge it brings to bear on the problem.  It influences the llm to take on that persona and bring specific expertise to the conversation.

ols/src/cache/redis_cache.py
```diff
@@ -47,7 +48,9 @@ def initialize_redis(self, config: RedisConfig) -> None:
         self.redis_client.config_set("maxmemory", config.max_memory)
         self.redis_client.config_set("maxmemory-policy", config.max_memory_policy)
 
-    def get(self, user_id: str, conversation_id: str) -> Union[str, None]:
+    def get(
+        self, user_id: str, conversation_id: str
+    ) -> Union[List[Conversation], None]:
```
bparees, line 53: see https://github.com/openshift/lightspeed-service/pull/357/files#r1479975907

ols/constants.py
xrajesh: I dont think I have to change anything here @onmete . Please confirm.

ols/src/cache/cache.py
xrajesh: `list[Conversation] | None` Looks more clear . find okay @onmete  ?

ols/src/cache/in_memory_cache.py
xrajesh: ditto going with `list[Conversation] | None`

ols/src/query_helpers/chat_history.py
xrajesh: made it `list[Conversation] | None`

ols/src/cache/cache.py
onmete: We agreed on using the `Optional` type on Monday's arch meet. I would go with that.
To my understanding, `Optional` has a meaning outside of Python, it basically means when there is nothing meaningful to return, it returns nothing. So it is not wrong to use it for return typing.
Sorry for my original message, I should give it a bit more thought.

ols/constants.py
onmete: Yes, I'll try to poke in the area, but no changes are needed here.

ols/src/cache/redis_cache.py
bparees: > Agree , but this is how it was , can we open a bug to address this ?

the point of OLS-110 is to do the "real implementation" 

There are some things where it's reasonable to queue it up for a follow up PR, but this doesn't seem like one of those cases.  We need proper error handling+logging when we implement new functionality.

ols/app/endpoints/ols.py
```diff
@@ -96,7 +97,10 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
                 # See the comment in `summarize` method in `docs_summarizer.py`
                 # Because of that, we are ignoring some type checks when we
                 # are creating response model.
-                response = llm_response.response
+                if isinstance(llm_response, str):
+                    response = llm_response
+                else:
+                    response = llm_response.response  # type: ignore
```
onmete, line 103: Can you check if the `# type: ignore` here (and bellow) is still needed after your changes?

ols/src/cache/redis_cache.py
```diff
@@ -70,14 +83,30 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
+                    values = [value]
+                    try:
+                        self.redis_client.set(key, values)
+                        break
+                    except Exception as redis_err:
+                        logger.info(f"Exception during creating the cache: {redis_err}")
+                        retry_count += 1
+
+        if retry_count == constants.REDIS_MAX_RETRY:
+            raise RedisConnectionError("Updating cache failed after max retries")
```
onmete, line 112: The intent is to use `RedisConnectionError` with different messages? Otherwise it should be something like `RedisMaxConnectionError`.

ols/src/query_helpers/chat_history.py
onmete: I don't know if something changed, but we should always use the `Optional` instead of `something | None`.

ols/src/cache/redis_cache.py
```diff
@@ -1,12 +1,22 @@
...
 from ols.src.cache.cache import Cache
+from ols.src.cache.conversation import Conversation
+
+logger = logging.getLogger(__name__)
+
+
+class RedisConnectionError(Exception):
+    """RedisConnectionError."""
+
+    pass
```
onmete, line 19: The pass is not needed.

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -35,8 +43,161 @@ def summarize(
...
+        logger.info(f"{conversation_id} Referenced documents: {referenced_documents}")
+        return summary_no_rag, referenced_documents
+
+    def get_docs_no_rag(
+        self,
+        conversation_id: str,
+        query: str,
+        history: Optional[str] = None,
+    ) -> tuple[str, str]:
+        """Summarize the given query, with no RAG , based on the provided conversation context .
```
onmete, line 80: `Summarize the given query, with no RAG, based on the provided conversation context.`

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -57,7 +218,10 @@ def summarize(
         logger.info(f"{conversation_id} call settings: {settings_string}")
 
         # TODO: use history there
-        summarization_template = PromptTemplate(constants.SUMMARIZATION_TEMPLATE)
+        logger.debug(f"History:  {history}")
+        chat_history = get_llama_index_chat_history(history)
+        logger.debug(f"Chat history for chat_engine:  {chat_history}")
```
onmete, line 223: There are double spaces in those msgs, does it mean something?

tests/integration/test_ols.py
```diff
@@ -327,17 +327,69 @@ def test_post_question_on_noyaml_response_type() -> None:
...
+                            },
+                        )
+                        response = client.post(
+                            "/v1/query",
+                            json={
+                                "conversation_id": conversation_id,
+                                "query": "Second query",
+                            },
+                        )
+                        print(response)
```
onmete, line 392: We had a [discussion ](https://github.com/openshift/lightspeed-service/pull/364#discussion_r1481832220)with @bparees about prints in e2e tests. They are here for the same purpose?

tests/integration/test_ols.py
```diff
@@ -327,17 +327,69 @@ def test_post_question_on_noyaml_response_type() -> None:
...
+            "ols.src.query_helpers.docs_summarizer.LLMLoader", new=mock_llm_loader(ml())
+        ):
+            with patch(
+                "ols.src.query_helpers.docs_summarizer.ServiceContext.from_defaults"
+            ):
+                with patch(
+                    "ols.utils.config.ols_config.reference_content.product_docs_index_path",
+                    "./invalid_dir",
+                ):
+                    with patch(
```
onmete, line 373: Could you use patch decorators for the test functions instead of this huge nesting?

onmete (commented):
Thank you! I've got just some nits.

/lgtm

ols/app/endpoints/ols.py
xrajesh: Yes, I need it because we could get string from langchain and Response from llama_index

tisnik: are you trying to access tuple item by it's name?

ols/src/cache/redis_cache.py
```diff
@@ -70,14 +83,30 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
+        while retry_count < constants.REDIS_MAX_RETRY:
+            logger.debug("Updating redis cache ")
+            with self._lock:
+                if old_value:
+                    old_value.append(value)
+                    try:
+                        self.redis_client.set(key, old_value)
+                        break
+                    except Exception as redis_err:
+                        logger.info(f"Exception during updating the cache: {redis_err}")
```
tisnik, line 100: is it just info message?

ols/src/cache/redis_cache.py
tisnik: IMHO if you `return` instead o `break` in the `while` loop, there's no nerd to check `retry_count` -> would be possible to simply raise an exception

tisnik (commented):
in overall it looks ok, just some nitpicks

ols/src/cache/redis_cache.py
onmete: btw that `RedisMaxConnectionError` should be `RedisMaxRetryError`... my bad

ols/app/endpoints/ols.py
xrajesh: Resolved it by checking for string/None

tests/integration/test_ols.py
xrajesh: It can come handy. But no take on it. :) @onmete

tests/integration/test_ols.py
xrajesh: Other tests are following similar pattern. Leaving this for now @onmete

onmete: Just to be clear, I don't mind having `with patch` in tests. Just to follow the same pattern you already have in `test_doc_summarizer.py` when there is a need to patch multiple objects.

ols/app/endpoints/ols.py
```diff
@@ -96,7 +97,13 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
                 # See the comment in `summarize` method in `docs_summarizer.py`
                 # Because of that, we are ignoring some type checks when we
                 # are creating response model.
-                response = llm_response.response
+                if isinstance(llm_response, str):
```
bparees, line 100: we should not need this.  if docs_summarizer.summarize isn't consistently returning a single type, please fix it, don't work around it like this.

ols/app/endpoints/ols.py
```diff
@@ -113,10 +120,9 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
                 )
 
     if config.conversation_cache is not None:
+        conversation = Conversation(llm_request.query, str(response or ""))  # type: ignore
```
bparees, line 123: if the response is None then it doesn't seem like we should be inserting/updating anything in the cache.  but also if the response is None we should probably be returning some sort of error or placeholder response to the user.

ols/src/cache/redis_cache.py
```diff
@@ -70,14 +81,33 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
         old_value = self.get(user_id, conversation_id)
-        with self._lock:
-            if old_value:
-                self.redis_client.set(key, old_value + "\n" + value)
-            else:
-                self.redis_client.set(key, value)
+        retry_count = 0
+        while retry_count < constants.REDIS_MAX_RETRY:
+            logger.debug("Updating redis cache ")
+            with self._lock:
```
bparees, line 91: this is too broad a lock, to do this properly you should lock on the conversation id

ols/src/cache/redis_cache.py
```diff
@@ -70,14 +81,33 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
+                if old_value:
+                    old_value.append(value)
+                    try:
+                        self.redis_client.set(key, old_value)
+                        break
+                    except Exception as redis_err:
```
bparees, line 97: not all errors make sense to retry, do you know what kinds of exceptions to expect?

ols/src/cache/redis_cache.py
bparees: also you're locking after you retrieved the value from redis.....  which means you're not avoiding race conditions because another process could insert/update the value after you do your initial get.

And finally, as soon as we have multiple instances of OLS running, talking to the same shared redis cache, this local locking won't accomplish anything.  We need to be locking at the redis level, not the client level, if we want to ensure no race conditions.

ols/src/cache/redis_cache.py
```diff
@@ -70,14 +81,33 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
+                        logger.error(
+                            f"Exception during updating the cache: {redis_err}"
+                        )
+                        retry_count += 1
+                else:
+                    values = [value]
+                    try:
+                        self.redis_client.set(key, values)
+                        break
+                    except Exception as redis_err:
```
bparees, line 107: same here, not all exceptions make sense to retry.

ols/src/cache/redis_cache.py
```diff
@@ -70,14 +81,33 @@ def insert_or_append(self, user_id: str, conversation_id: str, value: str) -> No
...
+                        logger.error(
+                            f"Exception during creating the cache: {redis_err}"
+                        )
+                        retry_count += 1
+
+        raise RedisMaxRetryError("Updating cache failed after max retries")
```
bparees, line 113: this should also report the last exception/error we got

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional
+
+from langchain.schema import ChatMessage as langchainCM
+from llama_index.llms import ChatMessage, MessageRole
+
+from ols.src.cache.conversation import Conversation
+
+
+def get_llama_index_chat_history(
+    conversations: Optional[list[Conversation]],
```
bparees, line 12: why is this optional?  when would you ever call this with no conversations arg?

ols/src/query_helpers/chat_history.py
bparees: > I dont expect to be None , but even if we get a None we will return a empty History

seems like a weird way to define what ends up being a constructor.  I think you should just not accept None/empty lists.

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,46 @@
...
+) -> list[ChatMessage]:
+    """Produce list of ChatMessage object using list of Conversations for llama_index."""
+    chat_history: list[Conversation] = []
+    if conversations is None or len(conversations) < 1:
+        return chat_history
```
bparees, line 17: this isn't the right type.  The function returns a list[ChatMessage]. but chat_history is a list[Conversation]

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,46 @@
...
+from llama_index.llms import ChatMessage, MessageRole
+
+from ols.src.cache.conversation import Conversation
+
+
+def get_llama_index_chat_history(
+    conversations: Optional[list[Conversation]],
+) -> list[ChatMessage]:
+    """Produce list of ChatMessage object using list of Conversations for llama_index."""
+    chat_history: list[Conversation] = []
```
bparees, line 15: shouldn't this be list[ChatMessage] ?

ols/src/query_helpers/chat_history.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+        chat_history.append(user_chat_message)
+        assistant_chat_message = ChatMessage(
+            role=MessageRole.ASSISTANT, content=conversation.assistant_message
+        )
+        chat_history.append(assistant_chat_message)
+    return chat_history
+
+
+def get_langchain_chat_history(
```
bparees, line 30: all the same comments on this:

don't allow None as an arg and fix the chat_history type.

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -54,7 +62,7 @@ def summarize(
         query: str,
         history: Optional[str] = None,
         **kwargs: Any,
-    ) -> tuple[Response, str]:
+    ) -> tuple[Response | str, str]:
```
bparees, line 65: we should always return the same type.  Making callers examine the object they got back to understand the type is going to be error prone.

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -64,8 +72,161 @@ def summarize(
...
+        Returns:
+            A tuple containing the summary as string and referenced documents as a string.
         """
+        settings_string = (
+            f"conversation_id: {conversation_id}, "
+            f"query: {query}, "
+            f"history: {history}, "
+        )
+        logger.debug(f"{conversation_id} call settings: {settings_string}")
+        langchian_chat_engine, chat_history_no_reference = (
```
bparees, line 126:
```suggestion
        langchain_chat_engine, chat_history_no_reference = (
```

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -64,8 +72,161 @@ def summarize(
...
+            {"query": query, "chat_history": chat_history_no_reference}
+        )
+        response_text = response.get("text")
+        response_val = (
+            "The following response was generated without access to reference content:"
+            "\n\n"
+            # NOTE: The LLM returns AIMessage, but typing sees it as a plain str
+            f"{response_text}"  # type: ignore
+        )
+        return response_val, ""
```
bparees, line 139: why have a second return value if it's always empty string?

bparees (commented):
Rather than supporting both the langchain and llamaindex query engine, we should fix the rag flow to only use langchain (use llamaindex to retrieve docs, append them to the prompt, and submit the prompt to langchain).  If that means putting this work on hold while we do that, so be it, but the work you're doing here to support both is throwaway in light of that intended direction.

ols/app/endpoints/ols.py
xrajesh: @bparees yes - it is happening because we are trying to support both llama_index and langchain

xrajesh:
closing this PR as we need to make changes after RAG refactor.

PR #355, opened 2024-02-06
OLS-233: Make provider/model required inputs for LLMLoader

onmete:
## Description

This PR contains a responsibility shift.

The `QueryHelper` knows which provider/model (and parameters) wants to use. It is its responsibility to set the correct inputs for `LLMLoader`. 

Resolving if the combo is actually valid input is responsibility the `LLMLoader`. It gets what provider, model, and additional parameters for llm along with the config against which the values will be validated.

## Type of change

- [x] Refactor

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -24,20 +24,13 @@ from ols.utils import config
 
 logger = logging.getLogger(__name__)
 
+
 # workaround to disable UserWarning
 warnings.simplefilter("ignore", UserWarning)
 
 
 class LLMConfigurationError(Exception):
     """LLM configuration is wrong."""
-
-
-class MissingProviderError(LLMConfigurationError):
-    """Provider is not specified."""
-
-
-class MissingModelError(LLMConfigurationError):
-    """Model is not specified."""
 
 
 class UnknownProviderError(LLMConfigurationError):
@@ -62,47 +55,56 @@ class LLMLoader:
     Example:
         ```python
         # using the class and overriding specific parameters
-        params = {'temperature': 0.02, 'top_p': 0.95}
+        llm_params = {'temperature': 0.02, 'top_p': 0.95}
 
-        bare_llm = LLMLoader(provider="openai", model="gpt-3.5-turbo", params=params)
+        bare_llm = LLMLoader(provider="openai", model="gpt-3.5-turbo", llm_params=llm_params).llm
         llm_chain = LLMChain(llm=bare_llm, prompt=prompt)
         ```
     """
 
     def __init__(
         self,
-        provider: Optional[str] = None,
-        model: Optional[str] = None,
-        params: Optional[dict] = None,
+        provider: str,
+        model: str,
+        llm_params: Optional[dict] = None,
     ) -> None:
         """Initialize LLM loader.
 
         Args:
             provider: The LLM provider name.
             model: The LLM model name.
-            params: The parameters to override and pass to the LLM backend.
+            llm_params: The parameters to override and pass to the LLM backend.
 
         Raises:
-            MissingProviderError: When provider is not specified.
-            MissingModelError: When model is not specified.
-            UnsupportedProviderError: When provider is not supported or is unknown.
+            UnknownProviderError: When no configuration exists for the requested provider.
+            UnsupportedProviderError: When provider is unsupported.
             ModelConfigMissingError: When no configuration exists for the requested model name.
         """
-        if provider is None:
-            msg = "Missing provider"
-            logger.error(msg)
-            raise MissingProviderError(msg)
         self.provider = provider
-        if model is None:
-            msg = "Missing model"
-            logger.error(msg)
-            raise MissingModelError(msg)
         self.model = model
-        self.provider_config = self._get_provider_config()
+        self.llm_params = llm_params or {}
+        self.llm_providers = config.config.llm_providers
+        self.provider_config = self._resolve_provider_config()
 
-        # return empty dictionary if not defined
-        self.llm_params = params or {}
-        self.llm: LLM = self._llm_instance()
+        self.llm = self._llm_instance()
+
+    def _resolve_provider_config(self) -> ProviderConfig:
+        """Ensure the provided inputs (provider/model) are valid in config."""
+        if self.provider not in self.llm_providers.providers:
+            raise UnknownProviderError(
+                f"Provider '{self.provider}' is not a valid provider. "
+                f"Valid providers are: {list(self.llm_providers.providers.keys())}"
+            )
+
+        provider_conf = self.llm_providers.providers.get(self.provider)
+
+        if self.model not in provider_conf.models:
+            raise ModelConfigMissingError(
+                f"Model '{self.model}' is not a valid model for provider '{self.provider}'. "
+                f"Valid models are: {list(provider_conf.models.keys())}"
+            )
+
+        return provider_conf
 
     def _get_llm_url(self, default: str) -> str:
         return (
@@ -121,24 +123,6 @@ class LLMLoader:
             if self.provider_config.models[self.model].credentials is not None
             else self.provider_config.credentials
         )
-
-    # TODO: refactor after config implementation OLS-89
-    def _get_provider_config(self) -> ProviderConfig:
-        cfg = config.llm_config.providers.get(self.provider)
-        if not cfg:
-            msg = f"No configuration for LLM provider {self.provider}"
-            logger.error(msg)
-            raise UnknownProviderError(msg)
-
-        model = cfg.models.get(self.model)
-        if not model:
-            msg = (
-                f"No configuration provided for model {self.model} under "
-                f"LLM provider {self.provider}"
-            )
-            logger.error(msg)
-            raise ModelConfigMissingError(msg)
-        return cfg
 
     def _llm_instance(self) -> LLM:
         logger.debug(
--- a/ols/src/query_helpers/__init__.py
+++ b/ols/src/query_helpers/__init__.py
@@ -1,16 +1,29 @@
 """Question validators, statament classifiers, and response generators."""
 
+import logging
 from typing import Optional
 
+from ols.src.llms.llm_loader import LLMLoader
 from ols.utils import config
+
+logger = logging.getLogger(__name__)
 
 
 class QueryHelper:
     """Base class for query helpers."""
 
     def __init__(
-        self, provider: Optional[str] = None, model: Optional[str] = None
+        self,
+        provider: Optional[str] = None,
+        model: Optional[str] = None,
+        llm_params: Optional[dict] = None,
+        llm_loader: Optional[LLMLoader] = None,
     ) -> None:
         """Initialize query helper."""
+        # NOTE: As signature of this method is evaluated before the config,
+        # is loaded, we cannot use the config directly as defaults and we
+        # need to use those values in the init evaluation.
         self.provider = provider or config.ols_config.default_provider
         self.model = model or config.ols_config.default_model
+        self.llm_params = llm_params or {}
+        self.llm_loader = llm_loader or LLMLoader
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -12,7 +12,6 @@ from llama_index.response.schema import Response
 from llama_index.schema import NodeWithScore, QueryBundle
 
 from ols import constants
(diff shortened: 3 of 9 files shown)


onmete:
@bparees @tisnik Before I proceed and update tests, I would like to know if you think it's a good approach.

The follow-up would be to split the `LLMLoader` into separate modules (for a different provider).

So the end flow would be:
something requests a llm (query helper in our case) -> llm loader validates inputs -> internal llm initialization -> returns llm

tisnik (commented):
yup, this approach seems to be +- ok
+ the caching mechanism will need to be tested, but might work as written

onmete:
Tests updated.

Lowered unit-test coverage target, that is too aggressive. See the note about not covering LLMLoader with unit tests.

Removed `test test_post_question_on_noyaml_response_type`, it seems like a not really readable duplicate of `test_summarize_incorrect_directory`.

Some dependency injection was introduced, which allows a bit more straightforward unit testing without patches or mocks.

/assign @bparees

ols/src/llms/llm_loader.py
```diff
@@ -54,44 +46,55 @@ class LLMLoader:
...
         params: Optional[dict] = None,
+        llm_providers: Optional[LLMProviders] = None,
     ) -> None:
         """Initialize LLM loader.
 
         Args:
             provider: The LLM provider name.
             model: The LLM model name.
             params: The parameters to override and pass to the LLM backend.
+            llm_providers: The LLM providers configuration.
```
bparees, line 66: when would we expect a caller to pass this in, rather than llm_loader using the OLS config?

ols/src/llms/llm_loader.py
```diff
@@ -54,44 +46,55 @@ class LLMLoader:
...
-            raise MissingProviderError(msg)
         self.provider = provider
-        if model is None:
-            msg = "Missing model"
-            logger.error(msg)
-            raise MissingModelError(msg)
         self.model = model
-        self.provider_config = self._get_provider_config()
+        self.params = params
+        self.llm_providers = llm_providers or config.llm_config
```
bparees, line 76: when/why would we expect a caller to need to override the global config?

ols/src/query_helpers/question_validator.py
```diff
@@ -16,6 +15,14 @@
 class QuestionValidator(QueryHelper):
     """This class is responsible for validating questions and providing one-word responses."""
 
+    @property
+    def llm_params(self) -> dict:
+        """Return the parameters for the LLM."""
+        return {
+            "min_new_tokens": 1,
+            "max_new_tokens": 4,
+        }
```
bparees, line 24: i don't understand the choice to define a function to return what is effectively a constant.  Seems unnecessarily complicated.  The caller of llm_loader can simply pass the llm parameters they want which makes it much more obvious at call time what args are being passed.

ols/src/query_helpers/__init__.py
```diff
@@ -1,16 +1,52 @@
...
+        """Return the parameters for the LLM."""
+        logger.info(f"no LLM parameters are configured for {__class__.__name__}")
+        return {}
+
+    @lru_cache(maxsize=None)
+    def _get_llm(self, provider: str, model: str, **llm_params) -> LLM:
+        loaded_llm = self.llm_loader(provider, model, params=llm_params)
+        return loaded_llm.llm
+
+    def get_llm(self) -> LLM:
```
bparees, line 43: by doing this this way you've made it difficult or impossible for a single query helper to make use of more than one LLM, or use different parameters for the llm.

i don't know if we'll have that use case, but in general i'm not a fan of the way this has been restructured.  It blends too much responsibility between the llm loader and the query helper.

What is wrong with a single utility function on llm_loader that takes the model/provider/model_params as args returns the llm instance to the caller?  The return value can still be cached.

ols/src/query_helpers/question_validator.py
```diff
@@ -51,9 +58,7 @@ def validate_question(
 
         logger.info(f"{conversation} Validating query")
 
-        bare_llm = LLMLoader(
-            self.provider, self.model, params={"min_new_tokens": 1, "max_new_tokens": 4}
-        ).llm
+        bare_llm = self.get_llm()
```
bparees, line 61: summarizing my other comments:  i would rather see this function be provided by llmloader and take provider/model/params arguments so that:

1) it's obvious at point of use what model/provider/params are going to be used
2) it's possible for a query helper to get refs to multiple llms or multiple instances of one llm w/ different params
3) it's easier to change the parameters dynamically

bparees (commented):
the changes to llm loader (i.e. making the model/provider required) seem ok, but i'm not a fan of the way you've structured the query helper and how the query helper instantiates the llm.

Also the choice of making the provider+model required in llm_loader means taking a position that the query helper is responsible for resolving any defaulting.  Previously the idea was that if a caller didn't have a preferred model/provider, llm_loader could fallback to whatever default was defined in the ols config.

Now instead the caller needs to resolve the default.  I'm ok either way on that, i think there's a reasonable argument to be made that the caller needs to make a deliberate choice of what provider/model it wants to use(even if that choice is to use some default), but just wanted to call out that this is the choice implied by how you've changed the code.

bparees:
/hold

i think we need to talk some more about the way you are proposing structuring the query helper and the workflow of how a query helper configures+instantiates an llm through llm_loader.

ols/src/llms/llm_loader.py
onmete: It decouples components.
I agree that sometimes the only obvious benefit is having a clean unit test without patches, but at the same time, I think [dependency injection](https://python-dependency-injector.ets-labs.org/introduction/di_in_python.html#what-is-dependency-injection) is a good pattern to follow.

Do you have concerns with this? Could it confuse someone?

ols/src/query_helpers/__init__.py
onmete: Parameters are passed from the respective query helper.
Only single llm per query helper - true.

With the utility function, you mean something like this
```python
@lru_cache(maxsize=None)  
def get_llm(provide, model, params) -> LLM:
    llm_loader = LLMLoader(provider=provide, model=model, params=params)
    return llm_loader.llm
```
in `llm_loader.py`? And you would call it directly in query helper? That is against the dependency injection.

I think it can be fixed like this:

- I'll remove the `get_llm` and `_get_llm` methods from query helper.
  - looking at that now again, you are right, it "tights hands" too much
- Revert to the original way of initializing llm loader, eg. `bare_llm = self.llm_loader(provider, model, params=llm_params)`
- Have caching on the side of `LLMLoader`.

So with that, you will be able to create multiple instances of llm in the query helpers.
WDYT?

ols/src/llms/llm_loader.py
bparees: by allowing the config to be passed in, you force llm_loader to validate the config that was passed in(or trust the caller to have done so), instead of being able to rely on the config loading logic to have verified the config.

so that is the trade-off to consider, the centralization of validation logic.  

if the only expected use for this parameter is ease of writing tests, then i don't think it belongs here.  If we want to make it easier to manipulate the OLS config for testing purposes, we can do that.

ols/src/query_helpers/question_validator.py
bparees: yes i think that would be better

onmete: changed

ols/src/llms/llm_loader.py
onmete: > you force llm_loader to validate the config that was passed

Sure, there are some exceptions, but generally, I think it is the caller's responsibility. Otherwise, you would need to validate everything everywhere.

TBH this is not true DI as we are not injecting dependencies in the true meaning of the word (that is more complicated and requires other libs that do it for you - a bit similar how the fixtures works in tests).
So usually DI means making things just a bit more configurable. Often times, the only obvious benefit is better testability.
Here, we are decoupling `LLMLoader` and `LLMProviders`. It still depends on it, but it is not internal parameter.

But if you think it doesn't make sense, I can drop it.

ols/src/llms/llm_loader.py
```diff
@@ -62,47 +54,58 @@ class LLMLoader:
...
             model: The LLM model name.
-            params: The parameters to override and pass to the LLM backend.
+            llm_params: The parameters to override and pass to the LLM backend.
+            llm_providers: The LLM providers configuration.
 
         Raises:
-            MissingProviderError: When provider is not specified.
-            MissingModelError: When model is not specified.
-            UnsupportedProviderError: When provider is not supported or is unknown.
+            UnknownProviderError: When provider is not is unknown.
```
bparees, line 80:
```suggestion
            UnknownProviderError: When no configuration exists for the requested provider.
```

ols/src/llms/llm_loader.py
bparees: > Sure, there are some exceptions, but generally, I think it is the caller's responsibility. Otherwise, you would need to validate everything everywhere.

making the caller do it turns into "validate everything everywhere" because you will have many potential callers(places from which you call the function), only one callee.  Callers tend to just make the call and then expect to catch an exception if the call was invalid.  A function should pretty much never just trust the arguments it is passed w/o validating them.

> So usually DI means making things just a bit more configurable. Often times, the only obvious benefit is better testability.

and the downside is that functions that take more arguments need more logic to handle reconciling those arguments, especially when they are optional.  And then more tests to ensure the handling is done correctly.

You can leave this one as is, but there is definitely a fine line between structuring code to make it easy to test, and making the code more complicated than it needs to be and risking introducing errors as a result of that (or confusing callers of the code into doing things we don't really want them doing, because we only made it possible to make it easier to write tests, not because we wanted people actually doing it)

bparees (commented):
I have a half a mind to completely rewrite llm_loader to just be a static function that takes a model/provider/params, looks up the config, and returns the appropriately instantiated llm.  It feels very over-engineered in its current form.

but in terms of evolving it as it is, this is fine, just fix the one exception text description.

ols/src/llms/llm_loader.py
onmete: changed

onmete:
Rebased and removed the config parameter from llm loader.

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -92,6 +89,9 @@ def summarize(
 
         embed_model = "local:BAAI/bge-base-en"
 
+        bare_llm = self.llm_loader(
+            self.provider, self.model, llm_params=self.llm_params
```
bparees, line 93: this starts to raise an interesting question as to whether llm_params are more closely associated with the model you're invoking, or with the use case (in this case docs summarization).

nothing to do for now, but something to think about as we continue down the path of tweaking llm parameters

cc @asamal4 @williamcaban

ols/src/query_helpers/__init__.py
```diff
@@ -1,16 +1,28 @@
...
+        llm_loader: Optional[LLMLoader] = None,
     ) -> None:
         """Initialize query helper."""
+        # NOTE: As signature of this method is evaluated before the config,
+        # is loaded, we cannot use the config directly as defaults and we
+        # need to use those values in the init evaluation.
         self.provider = provider or config.ols_config.default_provider
         self.model = model or config.ols_config.default_model
+        self.llm_loader = llm_loader or LLMLoader
+        self.llm_params: dict = {}
```
bparees, line 28: why not make this an optional parameter of the init function?  I see later you added it as a parameter of the derived classes, but then each derived class is responsible for setting this value in its init function.

ols/src/query_helpers/docs_summarizer.py
williamcaban: The params that are exposed are a combination of model type + functionality (e.g. embedding vs completion/query) + provider. 

When using the models for "embedding" (e.g., /embeddings or /api/embeddings)  there are parameters like padding and embedding dimensions that are configurable for the model but not all providers or loaders support configuring them.

When the model is used for "generate" or "chat" (e.g. /completion, /generate, /api/generate, /chat, /api/chat or /v1/chat/completions), the parameters are different (e.g. top_k, temperature, etc.) and the provider may also recognized additional parameters (e.g. num_thread, seed, num_gpu).

If want to use default settings and not pass any params, just define an empty array `{}` for the llm_params argument.

ols/src/query_helpers/__init__.py
onmete: changed

tisnik (changes requested):

PR #361, opened 2024-02-07
Change conversation to conversation_id

onmete:
## Description

Change `conversation` to `conversation_id` as it is misleading and implies it is something more than id.

## Type of change

- [x] Refactor


tisnik (commented):
seems legit

PR #362, opened 2024-02-07
Remove obsolete tests

onmete:
## Description

I'm genuinely confused about this one. It fails on me locally, but it seems ok in CI.

Still, these tests seem to be obsolete.
There is no validation logic in `LLMLoader` regarding checking if there is an API key set.
Such validation should happen when the config is created, so it should fail on `config.init_config("tests/config/empty_openai_api_key.yaml")`, not when initializing `LLMLoader`.
Also, I can't find the place that actually raises an error with `"Did not find openai_api_key"`.

## Type of change

- [x] Refactor


tisnik (commented):
it looks ok

PR #363, opened 2024-02-07
OLS-233: Import all llm libs at once

onmete:
## Description

We don't have any specific install groups. All deps listed in `pyproject.toml` are available.

## Type of change

- [x] Refactor

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -5,7 +5,18 @@ import logging
 import warnings
 from typing import Optional
 
+from genai import Client, Credentials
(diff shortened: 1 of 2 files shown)


tisnik (commented):
does not it need much more resources (especially RAM) to load everything, even for customers with only one provider configured?

onmete:
From my understanding, those are just imports. An actual LLM initialization happens in respected methods of `LLMLoader`.

bparees:
/lgtm
/approve

i think this was a premature optimization and i support simplifying it for now

PR #364, opened 2024-02-07
Fix e2e tests

onmete:
## Description

@bparees Not sure how long it was broken.

## Type of change

- [x] Test fix

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -67,10 +67,10 @@ def test_valid_question() -> None:
     json_response["conversation_id"] == conversation_id
     # assuming the response will be consistent
     assert (
-        "Kubernetes is an open source container orchestration tool"
+        "The following response was generated without access to reference content"
         in json_response["response"]
     )
     assert (
-        "The following response was generated without access to reference content:"
-        not in json_response["response"]
+        "Kubernetes is an open-source container orchestration platform"
+        in json_response["response"]
     )
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -29,7 +29,6 @@ def test_raw_prompt():
         json={"conversation_id": conversation_id, "query": "say hello"},
         timeout=20,
     )
-    print(vars(r))
     response = r.json()
 
     assert r.status_code == requests.codes.ok
@@ -44,7 +43,6 @@ def test_invalid_question():
         json={"conversation_id": conversation_id, "query": "test query"},
         timeout=20,
     )
-    print(vars(response))
     assert response.status_code == requests.codes.ok
     expected_details = (
         "I can only answer questions about OpenShift and Kubernetes. "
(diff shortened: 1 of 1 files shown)


tests/e2e/test_api.py
```diff
@@ -64,16 +62,15 @@ def test_valid_question() -> None:
...
     json_response["conversation_id"] == conversation_id
     # assuming the response will be consistent
     assert (
-        "Kubernetes is an open source container orchestration tool"
+        "The following response was generated without access to reference content"
         in json_response["response"]
     )
     assert (
-        "The following response was generated without access to reference content:"
-        not in json_response["response"]
```
onmete, line 78: Ok so it seems that the index is available in e2e, while I don't have it locally. What would be the right approach here?

The second update tool->platform seems to be valid though.

vishnuchalla: Yes, we do have an [index](https://github.com/openshift/lightspeed-service/blob/main/scripts/test-e2e-standalone.sh#L23) deployed as a part of e2e test in PROW. But if we want to distinguish our test asserts based on local and PROW, we can use these [envs](https://docs.prow.k8s.io/docs/jobs/#job-environment-variables) probably the `JOB_TYPE`.

onmete: I guess it might not be a good thing to have different asserts based on env, but on the other side, e2e are bit special here. So I can't say I'm against it. :)

We need some consensus if this is the way we want to proceed with our e2e tests. Otherwise, we need to update the dev process to ensure the dev env is the same as in CI for the e2e test.

So @bparees @tisnik @xrajesh WDYT?

tests/e2e/test_api.py
```diff
@@ -29,7 +29,6 @@ def test_raw_prompt():
         json={"conversation_id": conversation_id, "query": "say hello"},
         timeout=20,
     )
-    print(vars(r))
```
bparees, line 32: this is used for debugging when the test fails, do you have a reason to remove it?

tests/e2e/test_api.py
bparees: The dev process should be to run `scripts/test-standalone-e2e.sh` which pulls down and sets up the RAG content.

In general going forward we should be assuming the RAG content is present during testing and devs will need to get used to having it (it's not that hard to setup once and reference from your config yaml)

bparees (commented):
Rather than this PR, it sounds like maybe we need some updated instructions about setting up for and running the e2e tests.

tests/e2e/test_api.py
onmete: I can debug the tests in the same way as every other test (put debug point on the line I want the execution to stop). The app just needs to be running. So this doesn't seem helpful to me.

Is there a debug scenario I don't understand?

bparees: When i submit a PR and a test job fails this gives me the output(in the job logs) to help me understand why the test failed w/o having to go through the steps you go to rerun the test w/ a debugger or add print statements.

In an ideal world you should not have to rerun a test to understand why it failed, the test should output all relevant information as part of the failure.  And generally just the "Assert" output is not enough, you know what didn't match the expected value, but not necessarily the context needed to understand why it didn't match.

onmete:
nothing to fix here, closing

PR #365, opened 2024-02-07
Separate routers

tisnik:
## Description

- Separate routers handling functions
- It is hard/impossible to call any code from `main.py` from unit or integration tests, so it's better to refactor all functions out of it

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/app/routers.py
+++ b/ols/app/routers.py
@@ -1,0 +1,18 @@
+"""REST API routers."""
+
+from fastapi import FastAPI
+
+from ols.app import metrics
+from ols.app.endpoints import feedback, health, ols
+
+
+def include_routers(app: FastAPI):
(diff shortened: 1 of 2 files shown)


ols/app/routers.py
```diff
@@ -0,0 +1,18 @@
+"""REST API routers utility functions."""
```
onmete, line 1: Just "REST API routers"?

onmete:
one nit

/lgtm

ols/app/routers.py
tisnik: ok, changed @onmete

PR #368, opened 2024-02-07
OLS-259: Refactor the YAML config implementation fully utilizing pydantic

m-danilow:
## Description

The aim of this PR is to develop a standard for yaml config and to determine its requirements. It also simplifies the current implementation and makes use of pydantic capabilities.

## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -1,5 +1,5 @@
 llm_providers:
-  - name: my_bam
+  bam:
     type: bam
     url: "https://bam-api.res.ibm.com"
     credentials_path: bam_api_key.txt
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -1,315 +1,148 @@
 """Config classes for the configuration structure."""
 
 import logging
-import os
 import re
-from typing import Any, Optional, Self
-from urllib.parse import urlparse
+from typing import Any, Literal, Optional
 
-from pydantic import BaseModel, PositiveInt, model_validator
+from pydantic import (
+    AnyHttpUrl,
+    BaseModel,
+    DirectoryPath,
+    FilePath,
+    PositiveInt,
+    field_validator,
+    model_validator,
+)
 
 from ols import constants
 
 
-def _is_valid_http_url(url: str) -> bool:
-    """Check is a string is a well-formed http or https URL."""
-    result = urlparse(url)
-    return all([result.scheme, result.netloc]) and result.scheme in {
-        "http",
-        "https",
-    }
-
-
-def _get_attribute_from_file(data: dict, file_name_key: str) -> Optional[str]:
-    """Retrieve value of an attribute from a file."""
-    file_path = data.get(file_name_key)
-    if file_path is not None:
-        with open(file_path, mode="r", encoding="utf-8") as f:
-            return f.read().rstrip()
-    return None
-
-
-def _dir_check(path: str, desc: str) -> None:
-    """Check that path is a readable directory."""
-    if not os.path.exists(path):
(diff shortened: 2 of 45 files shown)


ols/app/models/config.py
```diff
@@ -5,580 +5,213 @@
...
 
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator, PositiveInt
+from typing import Literal
 
 from ols import constants
 
 
+# Helper functions
 def _is_valid_http_url(url: str) -> bool:
```
onmete, line 15: Pydantic has a [validator](https://docs.pydantic.dev/latest/api/networks/) for URLs.

ols/app/models/config.py
```diff
@@ -5,580 +5,213 @@
...
-        """Compare two objects for equality."""
-        if isinstance(other, LLMProviders):
-            return self.providers == other.providers
-        return False
-
-    def validate_yaml(self) -> None:
-        """Validate LLM config."""
-        for v in self.providers.values():
-            v.validate_yaml()
+    def __eq__(self, other):
```
onmete, line 53: Pydantic implements equality magic on its own. It is not sufficient enough?
What is the use case anyway for comparing two models/configs?

examples/olsconfig.yaml
```diff
@@ -1,10 +1,10 @@
 llm_providers:
-  - name: bam
+  bam:
     url: "https://bam-api.res.ibm.com"
     credentials_path: bam_api_key.txt
     models:
       - name: ibm/granite-13b-chat-v2
-  - name: openai
+  openai:
```
onmete, line 7: For context, we are currently defining providers as a list in the config, but converting it to dict in the actual model (to be able to access provider by the key).
So this reflects the code we have to the config.

onmete (commented):
It is an improvement, but I'm unsure if it properly shows the end result.

Would it be possible to raise another PR with just a single model refactored? Eg. just `logging_config` or `dev_config` along with necessary code changes and tests? That may give a cleaner view on now/after.

bparees:
@m-danilow is this still something you have time and inclination to pursue?  If so I agree with @onmete's recommendation that we start with a PR that just reworks one of the models so we can get a feel for what the changes look like and whether we prefer it overall, and then we can go from there.

tisnik:
@m-danilow are you testing your changes locally? Seems like most issues can be caught during local development. Also please look at @onmete and @bparees suggestions to try to change on model in separate PR to be reviewed and merged. It would be much easier for everybody and the changeset will be minimal (probably just `config.py` and probably few tests will need change).

bparees:
/close

We agree w/ the direction this PR suggests but given the rapid pace of change in the OLS config logic it just isn't feasible for us to stop and focus on this refactor (and the in depth review/testing required for such a substantial change) for the time it would take given our other priorities right now.  Unfortunately I think I overestimated how quickly we thought we'd stabilize those apis.

We have started the process of adopting pydantic validation to the config as we have refactored and added new config schemas and we will continue that process.  It's also possible that once we get through GA and the config api + implementation logic stabilizes a bit (and some of the other higher priority work is complete), we will be in a better position to take on a full refactor of the remaining config logic, but it probably won't be worth attempting to rebase this PR at that point vs just starting fresh (though this PR will likely be a useful implementation reference).

@m-danilow i apologize that we were never able to land this directly, but it was still of value as it has influenced the direction of the OLS config framework and will continue to do so.  Thank you for pointing us in this direction.

PR #376, opened 2024-02-08
Replaces helm chart with an example set of YAML manifests.

thoraxe:
## Description
Replaces helm chart with an example set of YAML manifests.

## Type of change

- [X] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [X] Documentation Update
- [ ] Configuration Update

N/A

## Testing
I deployed this set of manifests into a cluster when I was testing insights operator integration, so it was enough to allow me to use the Gradio UI to validate that it was working.


examples/openshift-lightspeed.yaml
```diff
@@ -0,0 +1,161 @@
...
+    targetPort: 8080
+  selector:
+    app: openshift-lightspeed
+    deployment: lightspeed-w-rag
+  sessionAffinity: None
+  type: ClusterIP
+
+---
+apiVersion: route.openshift.io/v1
+kind: Route
```
bparees, line 90: until we have auth on by default we should probably not create a route.  This essentially creates a public facing openai proxy that uses our credentials/bill

the console won't need a route anyway

examples/openshift-lightspeed.yaml
```diff
@@ -0,0 +1,161 @@
...
+    app.kubernetes.io/component: lightspeed-w-rag
+    app.kubernetes.io/instance: lightspeed-w-rag
+    app.kubernetes.io/name: lightspeed-w-rag
+    app.kubernetes.io/part-of: lightspeed-w-rag-app
+  name: olsconfig
+
+---
+apiVersion: v1
+data:
+  openai_api_key.txt: <INSERT_KEY_HERE>
```
bparees, line 151: Did this work for you?  It needs to be the base64 encoded version of the key.

or you can use stringData instead which allows you to provide the value in string format and then will be converted to base64 and placed into the data field automatically.

README.md
```diff
@@ -93,6 +93,26 @@ key file(s) (eg: `credentials_path: config/openai_api_key.txt`)
...
+`openshift-lightspeed.yaml`. This includes all the resources necessary to get
+OpenShift Lightspeed running in a cluster. It is configured expecting to only
+use OpenAI as the inference endpoint, but you can easily modify these manifests,
+looking at the `olsconfig.yaml` to see how to alter it to work with BAM as the
+provider.
+
+Note that this creates an OpenShift Route with TLS Edge termination enabled by 
+default. Depending on your OpenShift environment, this may result in a
+self-signed certificate being associated with the endpoint, requiring you to
+change your use of curl/Postman/etc.
```
bparees, line 107: instead of defining the route, let's just give them the "oc expose" command they can run to expose OLS via a route if they want to.

bparees (commented):
couple small items

examples/openshift-lightspeed.yaml
thoraxe: How would you access OLS for testing without a route when it's running "over there" (in a cluster)?

Even if you are using OpenShift Local, you would still need a route.

Without a route, the only way to access OLS from outside the cluster for testing would be to spin up a pod inside the cluster that has curl and then use `oc rsh` or `oc exec` or similar.

I'm fine to remove the route, I just don't understand how you expect to access it -- can you explain?

examples/openshift-lightspeed.yaml
thoraxe: Actually you are right, it was Base64. I'll fix that in the instructions.

examples/openshift-lightspeed.yaml
thoraxe: ah, forgot about port-forward. good call.

examples/openshift-lightspeed.yaml
```diff
@@ -0,0 +1,161 @@
...
+    app.kubernetes.io/component: lightspeed-w-rag
+    app.kubernetes.io/instance: lightspeed-w-rag
+    app.kubernetes.io/name: lightspeed-w-rag
+    app.kubernetes.io/part-of: lightspeed-w-rag-app
+  name: olsconfig
+
+---
+apiVersion: v1
+data:
+  openai_api_key.txt: <BASE64_ENCODED_OPENAI_API_KEY>
```
bparees, line 151: using the stringData field would make this easier for people so they don't have to base64 encode.

image

bparees (commented):
switch it to use the stringData field (which k8s automatically base64 encodes and copies to the data field), otherwise lgtm

bparees:
/approve
/lgtm

@kyoto give this a try next time you need to deploy OLS. Note that you'll have to uncomment the route definition if you want to use the route to create a TLS endpoint that the console can talk to.

PR #379, opened 2024-02-09
Unit test for `include_routers` function + mocked App class

tisnik:
## Description

Unit test for `include_routers` function + mocked App class

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

## Testing
- `make test-unit`

### DIFF
--- a/tests/unit/app/test_routers.py
+++ b/tests/unit/app/test_routers.py
@@ -1,0 +1,22 @@
+"""Unit tests for routers."""
+
+from ols.app import routers
+from ols.app.endpoints import feedback, health, ols
+from tests.mock_classes.mock_app import MockApp
+
+
+def test_include_routers():
+    """Tests if function include_routers add all routers into FastAPI application."""
+    # Alternativelly it would be possible to use MagicMock and then check calls etc.,
+    # but it is a bit fragile and less readable in the end.
+    app = MockApp()
+    routers.include_routers(app)
+
+    assert len(app.mounts) == 1 and app.mounts[0] == "/metrics"
+    assert len(app.prefixes) == 3
+    assert "/v1" in app.prefixes
+
+    # we don't expect the calls to be ordered
+    assert ols.router in app.routers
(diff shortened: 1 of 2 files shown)


tests/unit/app/test_routers.py
```diff
@@ -0,0 +1,22 @@
+"""Unit tests for routers."""
+
+from ols.app import routers
+from ols.app.endpoints import feedback, health, ols
+from tests.mock_classes.mock_app import MockApp
+
+
+def test_include_routers():
+    """Tests if function include_routers add all routers into FastAPI application."""
```
bparees, line 9: do we really need a test for this?  if we are failing to add a router to fastapi, then i'd expect we will have integration+e2e tests that will fail when the router/endpoint in question is not available.

tisnik: personally I'd like to have it covered to have a pretty stable test pyramid. In theory and later yes - e2e might be able to catch this. But not currently.

bparees: my view is that this kind of test just adds a refactoring and maintenance headache while not providing a test that's likely to protect us from anything we would otherwise miss

tisnik: np up to you, you can close this PR

bparees: i hate to do it but i think i am going to close it.  Feel free to say "i told you so" if at some point in the future we break something that this test would have caught :)

I am a little worried about the extent of heavily mocked tests we currently have that may hamper adding new features and refactoring code as we proceed.  That's not intended to discourage adding tests generally, but we do need to balance the cost of additional tests (writing them, maintaining them) against the value they provide.

bparees:
/close

see discussion in https://github.com/openshift/lightspeed-service/pull/379#discussion_r1484616470

PR #380, opened 2024-02-09
Refactor: LLMConfigurationError is not thrown from DocsSummarizer

tisnik:
## Description

Refactor: LLMConfigurationError is not thrown from DocsSummarizer

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/integration/test_ols.py
+++ b/tests/integration/test_ols.py
@@ -8,7 +8,6 @@ from fastapi.testclient import TestClient
 
 from ols import constants
 from ols.app.models.config import ProviderConfig, ReferenceContent
-from ols.src.llms.llm_loader import LLMConfigurationError
 from ols.utils import config, suid
 from tests.mock_classes.llm_chain import mock_llm_chain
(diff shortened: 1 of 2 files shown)


tests/integration/test_ols.py
```diff
@@ -167,31 +166,6 @@ def test_post_question_on_generic_response_type_summarize_error():
             assert response.json() == expected_json
 
 
-def test_post_question_on_generic_response_llm_configuration_error():
```
bparees, line 170: it is concerning that we managed to write a test that successfully verified that our code was doing something that our code doesn't actually do.  This is the kind of thing that worries me w/ our heavily mocked tests.

thanks for cleaning it up

tisnik: yeah, it should be part of DocSummarizer refactoring - but PR with this refactoring did not touch tests for some reason

PR #381, opened 2024-02-09
History type for Gradio's Chat UI

tisnik:
## Description

History type for Gradio's Chat UI

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Trivial change

### DIFF
--- a/ols/src/ui/gradio_ui.py
+++ b/ols/src/ui/gradio_ui.py
@@ -39,7 +39,7 @@ class GradioUI:
     def chat_ui(
         self,
         prompt: str,
-        history,
+        history: list[list[Optional[str]]],
         use_history: Optional[bool] = None,
         provider: Optional[str] = None,
         model: Optional[str] = None,


ols/src/ui/gradio_ui.py
```diff
@@ -39,7 +39,7 @@ def __init__(
     def chat_ui(
         self,
         prompt: str,
-        history,
+        history: list[list[Optional[str]]],
```
bparees, line 42: this argument isn't even used, how did you conclude this should be its data type?

(I think we should just be deleting the argument)

tisnik: is not `prompt` and `history` used automatically by the Chat widget?

tisnik: and the conclusion - I looked into Gradio's sources https://github.com/gradio-app/gradio/blob/7302a6e151dac553c17833be64d4639ee4cf97aa/gradio/chat_interface.py

bparees: sorry, you're correct:

from the gradio docs:

```
Should accept two parameters: a string input message and list of two-element lists of the form [[user_message, bot_message], ...] representing the chat history
```

PR #383, opened 2024-02-10
OLS-264: Add config support for watsonx

bparees:
Also move us to the preferred langchain-openai library instead of
langchain-community(where the openai chat support is deprecated)

## Description
adds a projectid field to llm provider configurations, which is needed for watsonx
support.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


bparees:
/assign @tisnik 

ptal

bparees:
/assign @syedriko 

also since he did most of the work on the config subsystem

syedriko:
@bparees This obviously gets the job done, but it doesn't account for the fact that project_id only makes sense for WatsonX. Config validation is delayed until way after config is loaded, too.

tisnik:
> @bparees This obviously gets the job done, but it doesn't account for the fact that project_id only makes sense for WatsonX. Config validation is delayed until way after config is loaded, too.

project_id is optional and filled-in/checked/used only for WatsonX, so it seems legit (other option will be to introduce model-specific and provider-specific params). So yes, gets the job done.

tisnik (commented):
LGTM. Probably if project_id has specific format (UUID perhaps?) we might add a validation for it as well.

bparees:
> project_id is optional and filled-in/checked/used only for WatsonX, so it seems legit (other option will be to introduce model-specific and provider-specific params). So yes, gets the job done.

yeah i did not want to introduce provider-specific config schemas at this point.  Arguably this continues a pattern we established w/ `url` which is also not needed for all providers (e.g. openai doesn't have alternate or custom api urls currently).  

Somewhat related to this, we should probably consider refactoring the provider config to allow for multiple instances of a given provider.  That is, we can add a `type` field to the provider (bam, watsonx, openai, etc), while keeping the name field.  We'd use the type field in llm_loader (and any other validation logic).

but the validation could be moved up to init time if we did provider specific validation during the config load ("if provider type=watsonx, require project id").  I'll revisit it on monday.


> Probably if project_id has specific format (UUID perhaps?) we might add a validation for it as well.

It would be fairly easy but i'm not sure how much confidence we should have that the format won't potentially change in the future.  I'd hate for us to have a bug where customers can't configure OLS successfully because we are rejecting valid project ids.

syedriko:
> Somewhat related to this, we should probably consider refactoring the provider config to allow for multiple instances of 
>a given provider. That is, we can add a `type` field to the provider (bam, watsonx, openai, etc), while keeping the name 
>field. We'd use the type field in llm_loader (and any other validation logic).

@bparees PTAL at https://github.com/openshift/lightspeed-service/pull/368, it's going in the opposite direction.

> > Probably if project_id has specific format (UUID perhaps?) we might add a validation for it as well.
> 
> It would be fairly easy but i'm not sure how much confidence we should have that the format won't potentially change in the future. I'd hate for us to have a bug where customers can't configure OLS successfully because we are rejecting valid project ids.

Agreed 100%. This is a pattern that comes up often enough where well-meaning validations of, say, URLs, file paths, phone numbers, etc, are added up in the call chain, only to find out later that the validations are incorrect because those who wrote them didn't understand some nuances in the first place, or the definition of valid has changed. We shouldn't interpret such things, just pass them along, and let the code that directly deals with them fail if it needs to. All that's left for us to do is propagate the error back to the user.
For WatsonX project_id, the source of truth about it is at https://ibm.github.io/watson-machine-learning-sdk/model.html#ibm_watson_machine_learning.foundation_models.Model. It doesn't mention a particular format, just where to get it from in the WatsonX UI.
BTW, there's also a space_id there and a note to the effect that "One of these parameters is required: [‘project_id ‘, ‘space_id’]"

ols/app/models/config.py
```diff
@@ -74,6 +74,7 @@ class ProviderConfig(BaseModel):
     name: Optional[str] = None
     url: Optional[str] = None
     credentials: Optional[str] = None
+    project_id: Optional[str] = None
```
onmete, line 77: What about having an `extra` section within the provider for these provider-specific configurations?

onmete: Ok, I see the discussion above now (I didn't see the comments originally).

bparees:
@syedriko @tisnik ok i think this is ready for another look.  I do think there are two refactoring followups worth doing later:

1) we have two "test_config.py" modules, one under utils and one under models.  This seems like something ripe for combining so we don't have a bunch of overlapping coverage of our config schema
2) when it comes to validating the config, we do some validation in the model class init functions, and other validation in the "validate_yaml" functions.  We should try to standardize.  After my experience having to make all the configs "valid" for various tests, my inclination is to make the init functions not do any validation, and instead make validation be something you can choose to run against a config object after you've instantiated/loaded it.  But at a minimum i'm not a fan of the way currently we validate some things at init time and other things when validate_yaml is called.

@xrajesh  @raptorsun  fyi, you'll need to add the `type` field to the operator CRD and logic to populate it in the config.yaml.  But if `type` is not specified, then the `name` field is used as the type, so it's a backwards compatible change.

ols/app/models/config.py
```diff
@@ -82,8 +84,22 @@ def __init__(self, data: Optional[dict] = None) -> None:
         if data is None:
             return
         self.name = data.get("name", None)
+        # Default provider type to be the provider name, unless
+        # specified explicitly.
+        self.type = data.get("type", self.name)
+        if self.type not in constants.SUPPORTED_PROVIDER_TYPES:
```
syedriko, line 90: str(self.type).lower() , as in llm_loader.py? or be UNIXy and don't lower() in llm_loader.py?

syedriko: actually, with addition of this code ^^^, the lower() in llm_loader.py doesn't do anything, only lowercase self.type will pass validation.

ols/src/llms/llm_loader.py
```diff
@@ -145,7 +145,7 @@ def _llm_instance(self) -> LLM:
             f"[{inspect.stack()[0][3]}] Loading LLM {self.model} from {self.provider}"
         )
         # convert to string to handle None or False definitions
-        match str(self.provider).lower():
+        match str(self.provider_config.type).lower():
```
syedriko, line 148: no need for this lower()

syedriko:
> @syedriko @tisnik ok i think this is ready for another look. I do think there are two refactoring followups worth doing later:
> 
> 1. we have two "test_config.py" modules, one under utils and one under models.  This seems like something ripe for combining so we don't have a bunch of overlapping coverage of our config schema
> 2. when it comes to validating the config, we do some validation in the model class init functions, and other validation in the "validate_yaml" functions.  We should try to standardize.  After my experience having to make all the configs "valid" for various tests, my inclination is to make the init functions not do any validation, and instead make validation be something you can choose to run against a config object after you've instantiated/loaded it.  But at a minimum i'm not a fan of the way currently we validate some things at init time and other things when validate_yaml is called.

@bparees agreed on the two test_config.py modules. as far as validation, PTAL at https://github.com/openshift/lightspeed-service/pull/368, it's the pydantic way of doing things, with a completely different approach to validation.

bparees:
> PTAL at https://github.com/openshift/lightspeed-service/pull/368, it's the pydantic way of doing things, with a completely different approach to validation.

yeah i've been watching that one from a distance, waiting to see it shape up a bit more, but i agree i'd rather have the schema validation logic built into the class definition where possible.

ols/app/models/config.py
bparees: probably needs a follow up, we should be case-insensitive on matching the provider type.

PR #392, opened 2024-02-12
OLS-233: Nest dev_config llm_params

onmete:
## Description

As those serve for dev purposes, it is not necessary to do the validation of the actual values.
Also, this allows more configuration without the explicit need to handle dev params.

## Type of change

- [x] Refactor

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -35,6 +35,7 @@ ols_config:
   default_model: ibm/granite-13b-chat-v2
 dev_config:
   enable_dev_ui: true
-  # llm_temperature_override: 0
+  # llm_params:
+  #   temperature_override: 0
   # disable_question_validation: false
   # disable_auth: false
--- a/tests/config/singleprovider.e2e.template.config.yaml
+++ b/tests/config/singleprovider.e2e.template.config.yaml
@@ -18,4 +18,5 @@ ols_config:
   default_provider: $PROVIDER
   default_model: $MODEL
 dev_config:
-  llm_temperature_override: 0
+  llm_params:
+    temperature: 0
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -507,7 +507,7 @@ class DevConfig(BaseModel):
 
     enable_dev_ui: bool = False
     disable_question_validation: bool = False
-    llm_temperature_override: Optional[float] = None
+    llm_params: Optional[dict] = None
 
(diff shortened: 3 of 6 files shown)


onmete:
The original condition `if config.dev_config.llm_temperature_override` evaluates as False when `llm_temperature_override` set to 0. Which we had in the config. So no temperature was actually overridden in tests.

It is weird, but the result of that is that "...an open source..." is not in the actual response (Kubernetes is an open source container orchestration tool
). I've set it to 0.05 in e2e test config as that seems to be the "right" value.

examples/olsconfig.yaml
```diff
@@ -25,6 +25,7 @@ ols_config:
   default_model: ibm/granite-13b-chat-v2
 dev_config:
   enable_dev_ui: true
-  # llm_temperature_override: 0
+  # llm_params:
+  #   temperature_override: 0
```
bparees, line 29: the reason i implemented this as i originally did is because I anticipate us potentially exposing all llm parameters as part of the model-specific config that even user can provide (or at least that we hardcode into the ols config).

this is ok for now, but will need to be revisited as we evaluate what llm configuration needs to be:
1) controlled on a per model basis, not a global basis
2) controlled by the user vs being hardcoded into OLS

bparees:
> It is weird, but the result of that is that "...an open source..." is not in the actual response (Kubernetes is an open source container orchestration tool
). I've set it to 0.05 in e2e test config as that seems to be the "right" value.

the point of setting it to zero was to ensure consistent responses from the LLMs between e2e runs.  So if it wasn't getting properly set to zero, that's a bug to fix, but we should not be setting it to a non-zero value for e2e testing purposes, we want reproducible results.

if that means we have to change the expected response slightly, then we should do that

tests/config/singleprovider.e2e.template.config.yaml
```diff
@@ -17,4 +17,5 @@ ols_config:
   default_provider: $PROVIDER
   default_model: $MODEL
 dev_config:
-  llm_temperature_override: 0
+  llm_params:
+    temperature: 0.05
```
bparees, line 21: please put this back to 0, my guess is it'll just work since now you're merging the entire params dict instead of checking if the temperature field is set.  But if not, please add whatever other changes are needed so that a temperature of zero is respected + used.

onmete:
@bparees As you proposed, I updated e2e test to check just some key information instead of a literal response.

PR #401, opened 2024-02-13
OLS-254: Refactoring: retrieving conversation ID

tisnik:
## Description

Refactoring: retrieving conversation ID

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-254](https://issues.redhat.com//browse/OLS-254)

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -31,16 +31,14 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
     """
     # Initialize variables
     previous_input = None
-    conversation_id = llm_request.conversation_id
 
     # TODO: retrieve proper user ID from request
     user_id = "user1"
 
-    # Generate a new conversation ID if not provided
-    if not conversation_id:
-        conversation_id = suid.get_suid()
-        logger.info(f"{conversation_id} New conversation")
-    else:
+    conversation_id = retrieve_conversation_id(llm_request)
+
+    # TODO: will be refactored in following PR
+    if llm_request.conversation_id:
         previous_input = config.conversation_cache.get(user_id, conversation_id)
         logger.info(f"{conversation_id} Previous conversation input: {previous_input}")
 
@@ -115,6 +113,18 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
     return llm_response
 
 
+def retrieve_conversation_id(llm_request: LLMRequest) -> str:
+    """Retrieve conversation ID based on existing ID or on newly generated one."""
+    conversation_id = llm_request.conversation_id
+
+    # Generate a new conversation ID if not provided
+    if not conversation_id:
+        conversation_id = suid.get_suid()
+        logger.info(f"{conversation_id} New conversation")
+
+    return conversation_id
+
+
 @router.post("/debug/query")
 def conversation_request_debug_api(llm_request: LLMRequest) -> LLMResponse:
     """Handle requests for the base LLM completion endpoint.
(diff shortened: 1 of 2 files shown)


tests/unit/app/endpoints/test_ols.py
```diff
@@ -21,6 +21,21 @@ def load_config():
     config.init_config("tests/config/test_app_endpoints.yaml")
 
 
+def test_retrieve_conversation_new_id(load_config):
+    """Check the function to retrieve conversation ID."""
+    llm_request = LLMRequest(query="Tell me about Kubernetes", conversation_id=None)
+    new_id = ols.retrieve_conversation_id(llm_request)
+    assert suid.check_suid(new_id), "Improper organization ID generated"
```
onmete, line 28: organization ID?

ols/app/endpoints/ols.py
```diff
@@ -115,6 +113,18 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
     return llm_response
 
 
+def retrieve_conversation_id(llm_request: LLMRequest) -> str:
```
onmete, line 116: Should be defined at the top, but you'll be moving it around I guess.

ols/app/endpoints/ols.py
```diff
@@ -31,16 +31,14 @@ def conversation_request(llm_request: LLMRequest) -> LLMResponse:
...
 
-    # Generate a new conversation ID if not provided
-    if not conversation_id:
-        conversation_id = suid.get_suid()
-        logger.info(f"{conversation_id} New conversation")
-    else:
+    conversation_id = retrieve_conversation_id(llm_request)
+
+    # TODO: will be refactored in following PR
+    if llm_request.conversation_id:
```
onmete, line 41: I did a PR a while ago when I was adding the generation of uuid in the pydantic model itself (in case it is missing). But in that case, some other attr needs to be set to know if the id was provided or generated.

This is just FYI. It is probably better to not handle it model.

tests/unit/app/endpoints/test_ols.py
tisnik: oh yes, 4 years in CCX. Sorry

ols/app/endpoints/ols.py
tisnik: yes

tests/unit/app/endpoints/test_ols.py
tisnik: fixed @onmete

ols/app/endpoints/ols.py
bparees, line 116: @tisnik this function needs to return a second value to indicate whether the conversation id was found in the request, or generated new.  Because we don't want to attempt a conversation cache lookup if we know the conversation id was just generated.  We won't find a conversation if it's a new id and:

1) we don't want to waste the effort on making a request we know will not find anything
2) we don't want to conflate "expected conversation not found in the cache"(merits a warning because while it can happen, it means we've lost history and the user may need to know about it) with "we didn't find it in the cache because it's a new conversation" (not a concern at all, totally expected)

onmete: Hmm, so with this, it might make sense to do the id generation in the request model. If it would have a field like `new_conversation: bool`. WDYT?

bparees: meaning all request objects would have a conversationid, and then new_conversation would indicate whether we should expect to find it in the cache or not?

it certainly could be done that way.  i'm not seeing it as a big improvement over "if the request object has no conversation id, it's a new conversation" though.

PR #407, opened 2024-02-13
Fix exception message

tisnik:
## Description

Fix exception message

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/src/cache/cache.py
+++ b/ols/src/cache/cache.py
@@ -24,13 +24,13 @@ class Cache(ABC):
         """Check if given user ID is valid."""
         # TODO: needs to be updated when we know the format
         if Cache.COMPOUND_KEY_SEPARATOR in user_id:
-            raise ValueError("Incorrect user ID {user_id}")
+            raise ValueError(f"Invalid user ID {user_id}")
 
     @staticmethod
     def _check_conversation_id(conversation_id: str) -> None:
         """Check if given conversation ID is a valid UUID (including optional dashes)."""
         if not check_suid(conversation_id):
-            raise ValueError(f"Incorrect conversation ID {conversation_id}")
+            raise ValueError(f"Invalid conversation ID {conversation_id}")
 
     @staticmethod
     def construct_key(user_id: str, conversation_id: str) -> str:
--- a/tests/unit/cache/test_in_memory_cache.py
+++ b/tests/unit/cache/test_in_memory_cache.py
@@ -53,15 +53,15 @@ def test_get_nonexistent_user(cache):
 
 def test_get_improper_user_id(cache):
     """Test how improper user ID is handled."""
-    with pytest.raises(ValueError, match="Incorrect user ID"):
+    with pytest.raises(ValueError, match="Invalid user ID :"):
         cache.get(":", conversation_id)
-    with pytest.raises(ValueError, match="Incorrect user ID"):
+    with pytest.raises(ValueError, match="Invalid user ID foo:bar"):
         cache.get("foo:bar", conversation_id)
 
 
 def test_get_improper_conversation_id(cache):
     """Test how improper conversation ID is handled."""
-    with pytest.raises(ValueError, match="Incorrect conversation ID"):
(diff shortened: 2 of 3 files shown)


ols/src/cache/cache.py
```diff
@@ -24,7 +24,7 @@ def _check_user_id(user_id: str) -> None:
         """Check if given user ID is valid."""
         # TODO: needs to be updated when we know the format
         if Cache.COMPOUND_KEY_SEPARATOR in user_id:
-            raise ValueError("Incorrect user ID {user_id}")
+            raise ValueError(f"Incorrect user ID {user_id}")
```
bparees, line 27:
```suggestion
            raise ValueError(f"Invalid user ID {user_id}")
```

tests/unit/cache/test_in_memory_cache.py
```diff
@@ -53,9 +53,9 @@ def test_get_nonexistent_user(cache):
 
 def test_get_improper_user_id(cache):
     """Test how improper user ID is handled."""
-    with pytest.raises(ValueError, match="Incorrect user ID"):
+    with pytest.raises(ValueError, match="Incorrect user ID :"):
```
bparees, line 56:
```suggestion
    with pytest.raises(ValueError, match="Invalid user ID :"):
```

tests/unit/cache/test_in_memory_cache.py
```diff
@@ -53,9 +53,9 @@ def test_get_nonexistent_user(cache):
...
         cache.get(":", conversation_id)
-    with pytest.raises(ValueError, match="Incorrect user ID"):
+    with pytest.raises(ValueError, match="Incorrect user ID foo:bar"):
```
bparees, line 58:
```suggestion
    with pytest.raises(ValueError, match="Invalid user ID foo:bar"):
```

tests/unit/cache/test_redis_cache.py
```diff
@@ -45,9 +45,9 @@ def test_get_nonexistent_key(cache):
 
 def test_get_improper_user_id(cache):
     """Test how improper user ID is handled."""
-    with pytest.raises(ValueError, match="Incorrect user ID"):
+    with pytest.raises(ValueError, match="Incorrect user ID :"):
```
bparees, line 48:
```suggestion
    with pytest.raises(ValueError, match="Invalid user ID :"):
```

tests/unit/cache/test_redis_cache.py
```diff
@@ -45,9 +45,9 @@ def test_get_nonexistent_key(cache):
...
         cache.get(":", conversation_id)
-    with pytest.raises(ValueError, match="Incorrect user ID"):
+    with pytest.raises(ValueError, match="Incorrect user ID foo:bar"):
```
bparees, line 50:
```suggestion
    with pytest.raises(ValueError, match="Invalid user ID foo:bar"):
```

bparees (commented):
basically we should have 2 kinds of errors:

1) invalid - the formatting is not correct or the userid or conversationid is missing entirely
2) not found - we got a userid/conversationid header, but we couldn't find a matching conversation in the cache

for case (2) we should log it and also return some indicator in the Response api object so the UI can warn the user, but it should not fail the overall request.  It just means the generated response won't benefit from conversation history context.

tests/unit/cache/test_in_memory_cache.py
tisnik: fixed

tests/unit/cache/test_redis_cache.py
tisnik: fixed

tests/unit/cache/test_redis_cache.py
tisnik: fixed

tisnik: fixed all @bparees

PR #413, opened 2024-02-14
OLS-254: Refactored: debug endpoint handler

tisnik:
## Description

- Refactored: debug endpoint handler
- Added unit tests (100% coverage)

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

- Related Issue #[OLS-254](https://issues.redhat.com//browse/OLS-254)

### DIFF
--- a/tests/unit/app/endpoints/test_ols.py
+++ b/tests/unit/app/endpoints/test_ols.py
@@ -222,15 +222,21 @@ def fake_llm_chain_call(self, **kwargs):
 
 @patch("ols.app.endpoints.ols.LLMChain", new=mock_llm_chain({"text": "llm response"}))
 @patch("ols.app.endpoints.ols.LLMLoader", new=mock_llm_loader(None))
-def test_conversation_request_debug_api(load_config):
+def test_conversation_request_debug_api_no_conversation_id(load_config):
     """Test conversation request debug API endpoint."""
-    # no conversation id
     llm_request = LLMRequest(query="Tell me about Kubernetes")
     response = ols.conversation_request_debug_api(llm_request)
     assert response.response == "llm response"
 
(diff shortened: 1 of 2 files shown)


tests/unit/app/endpoints/test_ols.py
```diff
@@ -322,3 +328,24 @@ def test_generate_response_unknown_validation_result(load_config):
...
+def test_generate_bare_response_no_conversation_id(load_config):
+    """Test the generation of bare response for debug endpoint."""
+    llm_request = LLMRequest(query="Tell me about Kubernetes", conversation_id=None)
+    response = ols.generate_bare_response(None, llm_request)
+    assert response == "llm response"
+
+
+@patch("ols.app.endpoints.ols.LLMChain", new=mock_llm_chain({"text": "llm response"}))
+@patch("ols.app.endpoints.ols.LLMLoader", new=mock_llm_loader(None))
+def test_generate_bare_response_with_conversation_id(load_config):
```
bparees, line 344: is there a meaningful difference between these two tests?  one passes a conversation id and one doesn't, but they seem to expect the same behavior, so in what sense is the converation id arg actually being tested or relevant?

i'm wary of tests that imply that they are testing more than they really are, which is why i bring it up.

tisnik: no, at this moment these tests check the same thing. I was thinking that in future, the conversation_id will be handled in tested function, but you are right, now it is not needed.

tisnik: updated @bparees

PR #419, opened 2024-02-14
OLS-118: Add Auth Support

nemerna:
## Description
Enable Authentication support in OLS using Openshift/K8S SSAR.
the provided implementation doesnt make use of any centralized k8s client that makes token review for users,
instead we use SSAR with the User token for that.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #267 
- Closes #267 

## Testing

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -40,9 +40,14 @@ ols_config:
   #   - name: bar_filter
   #     pattern: '\b(?:bar)\b'
   #     replace_with: "openshift"
+  authentication_config:
+    k8s_cluster_api: "https://api.example.com:6443"
+    k8s_ca_cert_path: "/Users/home/ca.crt"
+    skip_tls_verification: false
 dev_config:
   enable_dev_ui: true
   # llm_params:
   #   temperature_override: 0
   # disable_question_validation: false
   # disable_auth: false
+  # k8s_auth_token: optional_token_when_no_available_kube_config
--- a/ols/utils/auth_dependency.py
+++ b/ols/utils/auth_dependency.py
@@ -1,0 +1,214 @@
+"""Manage authentication flow for FastAPI endpoints with K8S/OCP."""
+
+import logging
+
+import kubernetes.client
+from fastapi import HTTPException, Request
+from kubernetes.client.rest import ApiException
+from kubernetes.config import ConfigException
+
+from ols.utils import config
+
+logger = logging.getLogger(__name__)
+
+
+class K8sClientSingleton:
+    """Return the Kubernetes client instances.
+
+    Ensures we initialize the k8s client only once per application life cycle.
+    manage the initialization and config loading.
+    """
+
+    _instance = None
+    _api_client = None
+    _authn_api = None
+    _authz_api = None
+
+    def __new__(cls):
+        """Create a new instance of the singleton, or returns the existing instance.
+
+        This method initializes the Kubernetes API clients the first time it is called.
+        and ensures that subsequent calls return the same instance.
+        """
+        if cls._instance is None:
+            cls._instance = super(K8sClientSingleton, cls).__new__(cls)
+            configuration = kubernetes.client.Configuration()
+
+            try:
+                if config.ols_config.authentication_config.k8s_cluster_api not in [
+                    None,
+                    "None",
+                    "",
+                ] and config.dev_config.k8s_auth_token not in [None, "None", ""]:
+                    logger.info("loading kubeconfig from app Config config")
+                    configuration.api_key["authorization"] = (
+                        config.dev_config.k8s_auth_token
+                    )
+                    configuration.api_key_prefix["authorization"] = "Bearer"
+                else:
+                    logger.debug(
+                        "no Auth Token Override was provided,\
+                            procceeding with in-cluster config load"
+                    )
+                    try:
+                        logger.info("loading in-cluster config")
+                        kubernetes.config.load_incluster_config(
+                            client_configuration=configuration
+                        )
+                    except ConfigException as e:
+                        logger.debug(f"unable to load in-cluster config: {e}")
+                        try:
+                            logger.info("loading config from kube-config file")
+                            kubernetes.config.load_kube_config(
+                                client_configuration=configuration
+                            )
+                        except ConfigException as e:
+                            logger.error(
+                                f"failed to load kubeconfig, in-cluster config\
+                                  and no override token was provided: {e}"
+                            )
+
+                configuration.host = (
+                    config.ols_config.authentication_config.k8s_cluster_api
+                    if config.ols_config.authentication_config.k8s_cluster_api
+                    not in [None, "None", ""]
+                    else configuration.host
+                )
+                configuration.verify_ssl = (
+                    not config.ols_config.authentication_config.skip_tls_verification
+                )
+                configuration.ssl_ca_cert = (
+                    config.ols_config.authentication_config.k8s_ca_cert_path
+                    if config.ols_config.authentication_config.k8s_ca_cert_path
+                    not in [None, "None", ""]
+                    else configuration.ssl_ca_cert
+                )
+                api_client = kubernetes.client.ApiClient(configuration)
+                cls._api_client = api_client
+                cls._authn_api = kubernetes.client.AuthenticationV1Api(api_client)
+                cls._authz_api = kubernetes.client.AuthorizationV1Api(api_client)
+            except Exception as e:
+                logger.info(f"Failed to initialize Kubernetes client: {e}")
+                raise
+        return cls._instance
+
+    @classmethod
+    def get_authn_api(cls):
+        """Return the Authentication API client instance.
+
+        Ensures the singleton is initialized before returning the Authentication API client.
+        """
+        if cls._instance is None or cls._authn_api is None:
+            cls()
+        return cls._authn_api
+
+    @classmethod
+    def get_authz_api(cls):
+        """Return the Authorization API client instance.
+
+        Ensures the singleton is initialized before returning the Authorization API client.
+        """
+        if cls._instance is None or cls._authz_api is None:
+            cls()
+        return cls._authz_api
+
+
+def get_user_info(token):
+    """Perform a Kubernetes TokenReview to validate a given token.
+
+    Parameters:
+        token (str): The bearer token to be validated.
+
+    Returns:
+        The user information if the token is valid, None otherwise.
+    """
+    auth_api = K8sClientSingleton.get_authn_api()
+    token_review = kubernetes.client.V1TokenReview(
+        spec=kubernetes.client.V1TokenReviewSpec(token=token)
+    )
+    try:
+        response = auth_api.create_token_review(token_review)
+        if response.status.authenticated:
+            return response.status
+        else:
+            return None
+    except ApiException as e:
+        logger.error(f"API exception during TokenReview: {e}")
+        return None
+    except Exception as e:
+        logger.error(f"Unexpected error during TokenReview - Unauthorized: {e}")
+        raise HTTPException(status_code=500, detail="Forbidden: Unable to Review Token")
+
+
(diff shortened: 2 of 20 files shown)


tests/unit/utils/test_auth_dependency.py
```diff
@@ -0,0 +1,29 @@
+"""Unit tests for the auth_dependency utility function.
+
+Currently it does not make real tests, the tests here are meant to increase coverage only.
```
bparees, line 3: We should not be doing this.  Please write unit tests that use a mock k8s client to test the auth logic you added

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,101 @@
...
+        configuration.verify_ssl = True
+
+        if config.ols_config.authentication_config.k8s_ca_cert_path:
+            configuration.ssl_ca_cert = (
+                config.ols_config.authentication_config.k8s_ca_cert_path
+            )
+        if config.ols_config.authentication_config.skip_tls_verification:
+            configuration.verify_ssl = False
+
+        k8s_client = kubernetes.client.ApiClient(configuration)
```
bparees, line 70: instead of having to create a new client w/ the configuration every time when the only thing that is changing is the authorization header, we should be able to create a single client with our own OLS token which has permission to do SubjectAccessReviews(SAR).

That single client can then be called repeatedly, each time performing a SAR, passing the user's access token associated w/ the request and asking k8s if that subject(token) has access to the /ols resource.

nemerna: the New implementation recently pushed, has the centralized client.
but in case of using centralized client, we can make SAR instead of SSAR
for SAR, we make use of user identity
for that i also had to implement fetching user identity feature as a prerequisite.

tests/unit/utils/test_auth_dependency.py
nemerna: I have fixed that, i hace provided real tests that still meet the Project coverage requirement

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,210 @@
...
+                        )
+                        logger.warn("Trying to load k8s config from app config")
+                        if (
+                            config.ols_config.authentication_config.k8s_cluster_api
+                            and config.dev_config.k8s_auth_token
+                        ):
+                            configuration.api_key["authorization"] = (
+                                config.dev_config.k8s_auth_token
+                            )
+                            configuration.api_key_prefix["authorization"] = "Bearer"
```
bparees, line 61: i think this order is backwards.  If the ols config supplied a k8s token, that is what we should prefer to use.  If that is not provided in the ols config, then we should attempt to load the in cluster config from the pod (which is what i'd expect to end up happening most of the time, the ols config would only be for override or dev/debug purposes)

bparees: also it's not clear to me that regardless of what exception is thrown while loading the k8s config, that setting this field and proceeding is an appropriate action to to take.  If you think it is, please comment this code explaining why it's appropriate

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,210 @@
...
+                    try:
+                        logger.warn(
+                            "kubeconfig file not found, trying in-cluster configuration"
+                        )
+                        kubernetes.config.load_incluster_config(
+                            client_configuration=configuration
+                        )
+                    except Exception:
+                        logger.warn(
+                            "Failed to load in-cluster Kubernetes configuration"
```
bparees, line 51: we should report the exception text

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,210 @@
...
+                cls._authorization_api = kubernetes.client.AuthorizationV1Api(
+                    api_client
+                )
+            except Exception as e:
+                logger.info(f"Failed to initialize Kubernetes client: {e}")
+                raise
+        return cls._instance
+
+    @classmethod
+    def get_auth_api(cls):
```
bparees, line 92:
```suggestion
    def get_authentication_api(cls):
```

or authn and then change get_authorization_api to get_authz_api

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,210 @@
...
+                    not config.ols_config.authentication_config.skip_tls_verification
+                )
+                configuration.ssl_ca_cert = (
+                    config.ols_config.authentication_config.k8s_ca_cert_path
+                    if config.ols_config.authentication_config.k8s_ca_cert_path
+                    else configuration.ssl_ca_cert
+                )
+                api_client = kubernetes.client.ApiClient(configuration)
+                cls._api_client = api_client
+                cls._auth_api = kubernetes.client.AuthenticationV1Api(api_client)
```
bparees, line 82: `authentication` or `authn` if you want to keep it short (and then authz for authorization).  But just `auth` is ambiguous.

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,210 @@
...
+        """Return the Authorization API client instance.
+
+        Ensures the singleton is initialized before returning the Authorization API client.
+        """
+        if cls._instance is None or cls._authorization_api is None:
+            cls()
+        return cls._authorization_api
+
+
+def perform_token_review(token):
```
bparees, line 112:
```suggestion
def get_user_info(token):
```

?

tests/unit/utils/test_auth_dependency.py
```diff
@@ -0,0 +1,80 @@
...
+    mock_token_review_response,
+)
+
+
+@patch.dict(os.environ, {"OLS_CONFIG_FILE": "tests/config/auth_config.yaml"})
+def setup():
+    """Setups and load config."""
+    global client
+    cfg_file = os.environ.get("OLS_CONFIG_FILE", "tests/config/auth_config.yaml")
+    config.init_config(cfg_file)
```
bparees, line 23: why do you need these 2 lines?  the other tests don't need them to setup the config yaml path

tests/unit/utils/test_auth_dependency.py
```diff
@@ -0,0 +1,80 @@
...
+    from ols.app.main import app
+
+    client = TestClient(app)
+
+
+@patch.dict(os.environ, {"OLS_CONFIG_FILE": "tests/config/auth_config.yaml"})
```
bparees, line 29: should not need this?

tests/unit/utils/test_auth_dependency.py
```diff
@@ -0,0 +1,80 @@
...
+@pytest.mark.asyncio
+@patch("ols.utils.auth_dependency.K8sClientSingleton.get_auth_api")
+@patch("ols.utils.auth_dependency.K8sClientSingleton.get_authorization_api")
+@patch.dict(os.environ, {"OLS_CONFIG_FILE": "tests/config/auth_config.yaml"})
```
bparees, line 33: shouldn't need this?

tests/unit/utils/test_auth_dependency.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Check if the correct user info has been returned
+    assert user_uid == "valid-uid"
+    assert username == "valid-user"
+
+
+@pytest.mark.asyncio
+@patch("ols.utils.auth_dependency.K8sClientSingleton.get_auth_api")
+@patch("ols.utils.auth_dependency.K8sClientSingleton.get_authorization_api")
+@patch.dict(os.environ, {"OLS_CONFIG_FILE": "tests/config/auth_config.yaml"})
```
bparees, line 59: shouldn't need this

bparees (commented):
looks pretty close, a few small comments

tests/unit/utils/test_auth_dependency.py
nemerna: Resolved

tests/unit/utils/test_auth_dependency.py
nemerna: Resolved

tests/unit/utils/test_auth_dependency.py
nemerna: Resolved

tests/unit/utils/test_auth_dependency.py
nemerna: Resolved removed additional unneeded line

ols/utils/auth_dependency.py
nemerna: patched

ols/utils/auth_dependency.py
nemerna: changed to authn for authentication and authz for authorization

ols/utils/auth_dependency.py
nemerna: changed to authn for authentication and authz for authorization

ols/utils/auth_dependency.py
nemerna: added the exception message to error logs

ols/utils/auth_dependency.py
nemerna: changed the order and added the exact expected exception

ols/utils/auth_dependency.py
```diff
@@ -0,0 +1,214 @@
...
+                    )
+                    try:
+                        logger.info("loading in-cluster config")
+                        kubernetes.config.load_incluster_config(
+                            client_configuration=configuration
+                        )
+                    except ConfigException as e:
+                        logger.debug(f"unable to load in-cluster config: {e}")
+                        try:
+                            logger.info("loading config from kubec-config file")
```
vishnuchalla, line 61: probably a typo here.

examples/olsconfig.yaml
```diff
@@ -32,8 +32,13 @@ ols_config:
...
   default_model: ibm/granite-13b-chat-v2
+  authentication_config:
+    k8s_cluster_api: "https://api.example.com:6443"
+    k8s_ca_cert_path: "/Users/home/ca.crt"
+    skip_tls_verification: false
 dev_config:
   enable_dev_ui: true
   # llm_temperature_override: 0
   # disable_question_validation: false
   # disable_auth: false
```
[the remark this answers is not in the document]
bparees: @nemerna  this comment still needs to be addressed

bparees (commented):
/approve

needs an update to the readme(for setting auth enabled/disabled), and addressing @vishnuchalla's latest comments and then i think it's good.

bparees:
@kyoto heads up when this lands it's going to require the console invoke OLS w/ a bearer token for a user who has permission to call OLS (rbac permission).

The easiest way to avoid being blocked by the auth requirement will be to set the dev config field `disable_auth: true` in the ols config yaml.

tests/unit/utils/test_auth_dependency.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Expect an HTTPException for invalid tokens
+    with pytest.raises(HTTPException) as exc_info:
+        await auth_dependency(request)
+
+    # Check if the correct status code is returned for unauthorized access
+    assert exc_info.value.status_code == 403
+
+
+@pytest.mark.asyncio
```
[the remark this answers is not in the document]
nemerna: @vishnuchalla  Indeed you are right i've missed that!

Resolved and removed the async from method signature
 and removed the annotation

ols/utils/auth_dependency.py
nemerna: Correct, Resolved

examples/olsconfig.yaml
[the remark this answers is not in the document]
nemerna: Documented

bparees:
> We would need operator create the roles needed for SSAR for the operand before this PR is merged. @nemerna @raptorsun FYI.

@xrajesh we don't need to wait for that, we can continue to run OLS w/ auth disabled in environment where we haven't setup the roles/bindings/etc.

PR #425, opened 2024-02-15
OLS-58: Truncating flag to be returned in response

tisnik:
## Description

Truncating flag to be returned in response.
This need changes in model, changes in DocsSummarizer etc.
The trunsation itself is not implemented there. This is 1st step to avoid having huge PR that will need to be constantly rebased.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-58](https://issues.redhat.com//browse/OLS-58)

### DIFF
--- a/ols/app/models/models.py
+++ b/ols/app/models/models.py
@@ -61,11 +61,13 @@ class LLMResponse(BaseModel):
         conversation_id: The optional conversation ID (UUID).
         response: The optional response.
         referenced_documents: The optional URLs for the documents used to generate the response.
+        truncated: Set to True if conversation history was truncated to be within context window.
     """
 
     conversation_id: str
     response: str
     referenced_documents: list[str]
+    truncated: bool
 
     # provides examples for /docs endpoint
     model_config = {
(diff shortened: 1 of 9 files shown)


vishnuchalla:
Considering this as a pre-flight PR for trunsation, looks good to me.

ols/app/models/models.py
```diff
@@ -61,11 +61,13 @@ class LLMResponse(BaseModel):
         conversation_id: The optional conversation ID (UUID).
         response: The optional response.
         referenced_documents: The optional URLs for the documents used to generate the response.
+        truncated: Set to True if conversation history was truncated to be within context window.
```
bparees, line 64: @tisnik i would have give this a more explicit name, like `history_truncated` so that it's clearer what it refers to, and so that if we have other "truncation" related metadata we want to include in the response, the field names won't conflcit.  APIs are forever(once we ship), so names matter.

tests/e2e/test_api.py
```diff
@@ -54,6 +54,7 @@ def test_invalid_question():
         "conversation_id": conversation_id,
         "response": expected_details,
         "referenced_documents": [],
+        "truncated": False,
```
bparees, line 57: any reason you only did this on one of the e2e tests?

we don't need to test every field in every e2e test, but at least checking for it in a `valid` response flow and not just a rejected one seems to have value.

PR #426, opened 2024-02-15
 OLS-105: Add ability to filter terms in user query

xrajesh:
## Description
Adds ability to specify list of Regex and redact string in olsconfig yaml. This will be used to redact questions/queries from user to avoid sensitive information from queries.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue [#](https://issues.redhat.com/browse/OLS-105)

## Testing


examples/olsconfig.yaml
```diff
@@ -32,6 +32,13 @@ ols_config:
     lib_log_level: warning
   default_provider: my_bam
   default_model: ibm/granite-13b-chat-v2
+# question_filters:
```
tisnik, line 35: is the idea to have it commented to disable filtering?
Won't it be better to use some enable/disable option?

ols/app/models/config.py
```diff
@@ -301,6 +302,57 @@ def validate_yaml(self) -> None:
...
+    regular_expression: Optional[str] = None
+    replace_with_string: Optional[str] = None
+
+    def __init__(self, data: Optional[dict] = None) -> None:
+        """Initialize configuration and perform basic validation."""
+        super().__init__()
+        if data is None:
+            return
+        try:
+            self.name = data.get("name", None)
```
tisnik, line 318: None is default value, not have to be specified (but it's used elsewhere anyway)

ols/app/models/config.py
```diff
@@ -301,6 +302,57 @@ def validate_yaml(self) -> None:
...
+            self.replace_with_string = data.get("replace_with_string", None)
+            if (
+                self.name is None
+                or self.regular_expression is None
+                or self.replace_with_string is None
+            ):
+                raise ValueError
+        except ValueError:
+            raise InvalidConfigurationError(
+                "name,regular_expression and replace_with_string needs to be a string"
```
tisnik, line 329: Oxford comma? :)

tisnik: "need to be specified" perhaps

tisnik (commented):
looks ok in overall! just some nitpicks

ols/utils/question_filter.py
```diff
@@ -0,0 +1,51 @@
...
+    def setup(cls) -> "QuestionFilter":
+        """Set up  the class instance using question_filter found in Config."""
+        logger.debug(f"Question filters : {config.ols_config.question_filters}")
+        regex_filters: list[RegexFilter] = []
+        if not config.ols_config.question_filters:
+            return cls(regex_filters)
+        for filter in config.ols_config.question_filters:
+            try:
+                pattern = re.compile(
+                    str(filter.regular_expression), flags=re.IGNORECASE
```
syedriko, line 32: we're not compiling with `re.IGNORECASE` in config validation

tests/integration/test_ols.py
```diff
@@ -249,3 +249,56 @@ def test_post_question_on_noyaml_response_type() -> None:
...
+    config.ols_config.reference_content = ReferenceContent(None)
+    config.ols_config.reference_content.product_docs_index_path = "./invalid_dir"
+    config.ols_config.reference_content.product_docs_index_id = "product"
+    question_filters = [
+        Filter(
+            {
+                "name": "test",
+                "regular_expression": r"(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})",
+                "replace_with_string": "redacted_ip",
+            }
```
syedriko, line 266: also a matter of taste, `redacted_ip` could be XXX.XXX.XXX.XXX

examples/olsconfig.yaml
```diff
@@ -32,6 +32,13 @@ ols_config:
...
+# - name: ip_filter
+#   regular_expression: '(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'
```
tisnik, line 37: strictly speaking it is not IP (can match 999.999.999.999), but yes, it's not the point of this PR

syedriko:
@xrajesh What I'm wondering, is the user going to see the redacted bits in the replies from the LLM or are we looking to redact things between the OLS and the LLM and unredact them between the user and the OLS.

examples/olsconfig.yaml
syedriko: strictly speaking, this doesn't cover IPv6 addresses either, but that's not the point ;)

tests/integration/test_ols.py
xrajesh: user can choose what they want to replace with. This is only a test.

examples/olsconfig.yaml
xrajesh: This is a example file. We dont know what the real users regex and replacement words will be.

ols/app/models/config.py
```diff
@@ -301,6 +302,57 @@ def validate_yaml(self) -> None:
         """Validate memory cache config."""
 
 
+class Filter(BaseModel):
+    """Filter configuration."""
+
+    name: Optional[str] = None
+    regular_expression: Optional[str] = None
+    replace_with_string: Optional[str] = None
+
```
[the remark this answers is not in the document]
xrajesh: I am not good with naming things. Will change to "pattern" and "replace_with" @syedriko

xrajesh:
@syedriko The real user will not be able to see the actual questions sent to LLM. But the admins will use this to configure their filter words . They will be able to see the actual questions, ONLY in the logs :)

syedriko:
> @syedriko The real user will not be able to see the actual questions sent to LLM. But the admins will use this to configure their filter words . They will be able to see the actual questions, ONLY in the logs :)

Let's say a user asks (a contrived example) "How can I schedule a pod on a node with IP 1.2.3.4?" and the OLS is configured to replace IPs with `redacted_ip`, isn't the user going to get a reply with `redacted_ip` in it, like

"First apply a label to the node:
$ oc label nodes <node-name> mynodeip=redacted_ip
"
?

xrajesh:
@syedriko The user can even replace his/her real ip with ""(empty str) or give a fake ip like 9.9.9.9 - Goal is to mask the users identity.

ols/app/endpoints/ols.py
```diff
@@ -142,6 +143,20 @@ def store_conversation_history(
 
 def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
     """Validate user question, raise HTTPException in case of any problem."""
+    # Redact the input question
+    try:
+        redacted_query = QuestionFilter.setup().redact_question(llm_request.query)
```
bparees, line 148: i don't think this should be called from within validate_question, validating questions and filtering/redacting questions should be considered peer steps in the query flow.  This current implementation couples them together in an undesirable way

bparees: also the fact that setup() is being called on every request is not ideal, we should only need to call it once during OLS initialization.

tests/integration/test_ols.py
```diff
@@ -249,3 +249,56 @@ def test_post_question_on_noyaml_response_type() -> None:
...
                         in response.json()["response"]
                     )
+
+
+def test_post_query_with_question_filter_response_type() -> None:
+    """Check the REST API /v1/query with POST HTTP method with question filter."""
+    config.init_empty_config()
+    config.ols_config.reference_content = ReferenceContent(None)
+    config.ols_config.reference_content.product_docs_index_path = "./invalid_dir"
+    config.ols_config.reference_content.product_docs_index_id = "product"
```
bparees, line 259: setting up an invalid path for reference content seems strange.  Why not just not provide a reference content config at all?  OLS is supposed to still respond to queries even if rag is unavailable

tests/mock_classes/mock_question_filter.py
```diff
@@ -0,0 +1,18 @@
+"""Mocked QuestionFilter."""
+
+
+class MockQuestionFilter:
```
bparees, line 4: why do we need a mock for this?  Seems pretty straightforward to create the actual object

tests/unit/app/models/test_config.py
```diff
@@ -1086,3 +1086,157 @@ def test_config_improper_model():
...
+                },
+            }
+        ).validate_yaml()
+
+
+def test_config_with_multiple_question_filter():
+    """Test the Config model with multiple question filter is set."""
+    with pytest.raises(
+        InvalidConfigurationError,
+        match="default_model specifies an unknown model test_default_model",
```
bparees, line 1146: this doesn't seem like the right expected error.  Is it supposed to be an error to have 2 filters with the same name?  if so this should expect that error.

If the filters defined here are supposed to be considered valid, then fix the test so that the config is overall valid and no exception is expected.

tests/unit/app/models/test_config.py
```diff
@@ -1086,3 +1086,157 @@ def test_config_improper_model():
...
+                            "replace_with": "test_replace_with",
+                        },
+                    ],
+                },
+            }
+        ).validate_yaml()
+
+
+def test_config_invalid_regex_question_filter():
+    """Test the Config model with improper question filter (no name) is set."""
```
bparees, line 1194: test description is wrong

bparees (commented):
can you try to add an e2e test?  one way i'd approach it would be to do something like:

1) ask a question like "what is a `foo` in `bar`"
2) setup replacement filters in the e2e config yaml to replace `foo` and `bar` with `deployment` and `openshift`, respectively)
3) confirm you get a reasonable answer back(answer contains some expected keywords) from the LLM

in general i'd like people to start thinking about how to add e2e tests for the functionality we're building.

ols/app/endpoints/ols.py
xrajesh: @bparees I was expecting this comment. When I looked it , we need to make use of fastapi lifespan to do this . https://fastapi.tiangolo.com/advanced/events/#async-context-manager . I would like this to be separate from this PR . as I feel that , I will touch a lot of files.

bparees: doesn't seem like we have to do it that way.  We load/init our config only a single time and then share/reuse it.  Even the docs you linked to say:

```
You could load it at the top level of the module/file, but that would also mean that it would load the model even if you are just running a simple automated test, then that test would be slow because it would have to wait for the model to load before being able to run an independent part of the code.
```

the drawbacks don't seem to me to apply to this init logic, nor do we care about cleaning it up on shutdown, so the lifecycle events seem like a nice way to handle this, but certainly not the only way and i'm not convinced it justifies a refactor that would touch a lot of files.


but i am glad to hear you anticipated the comment and had thought about how to address it!

xrajesh: Loaded part of init function.

tests/unit/app/endpoints/test_ols.py
```diff
@@ -139,17 +139,32 @@ def test_validate_question_on_validation_error(validate_question_mock, load_conf
...
 def test_conversation_request(
+    mock_question_redactor,
     mock_conversation_cache_get,
     mock_summarize,
     mock_validate_question,
-    load_config,
 ):
     """Test conversation request API endpoint."""
+    # redact question
+    mock_question_redactor.return_value = "Tell me about Kubernetes"
```
bparees, line 166: the test doesn't seem to rely on this value so why do we need to have a mock return it?

not clear why we care about setting up a question filter mock at all for this test.

tests/unit/app/endpoints/test_ols.py
```diff
@@ -193,12 +208,15 @@ def test_conversation_request(
 
 @patch("ols.src.query_helpers.question_validator.QuestionValidator.validate_question")
 @patch("ols.utils.config.conversation_cache.get")
+@patch("ols.utils.config.QuestionFilter.redact_question")
```
bparees, line 211: same question here.....why do we need this?

tests/e2e/test_api.py
```diff
@@ -104,3 +104,22 @@ def test_rag_question() -> None:
         "The following response was generated without access to reference content:"
         not in json_response["response"]
     )
+
+
+def test_question_filter() -> None:
+    """Ensure responses does not include filterd words."""
```
bparees, line 110:
```suggestion
    """Ensure responses does not include filtered words."""
```

tests/unit/app/endpoints/test_ols.py
xrajesh: It does not need it . Fixed it.

tests/unit/app/endpoints/test_ols.py
xrajesh: It does not need it . Fixed it.

ols/utils/question_filter.py
```diff
@@ -0,0 +1,51 @@
...
+    @classmethod
+    def setup(cls) -> "QuestionFilter":
+        """Set up  the class instance using question_filter found in Config."""
+        logger.debug(f"Question filters : {config.ols_config.question_filters}")
+        regex_filters: list[RegexFilter] = []
+        if not config.ols_config.question_filters:
+            return cls(regex_filters)
+        for filter in config.ols_config.question_filters:
+            try:
+                pattern = re.compile(str(filter.pattern), flags=re.IGNORECASE)
```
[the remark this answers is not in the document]
xrajesh: Will remove the Exception handling here .

ols/app/endpoints/ols.py
```diff
@@ -172,8 +176,30 @@ def store_conversation_history(
...
+    try:
+        logger.debug(f"Redacting query for conversation {conversation_id}")
+        if not config.question_redactor:
+            logger.debug("question_redactor not found")
+            return llm_request
+        redacted_question = config.question_redactor.redact_question(
+            conversation_id, llm_request.query
+        )
+        llm_request.query = redacted_question
+        return llm_request
```
syedriko, line 190: llm_request.query = config.question_redactor.redact_question(
            conversation_id, llm_request.query
        )

ols/app/endpoints/ols.py
```diff
@@ -172,8 +176,30 @@ def store_conversation_history(
...
+    except Exception as redactor_error:
+        logger.error("Error while redacting query")
+        logger.error(redactor_error)
+        raise HTTPException(
```
syedriko, line 194: can we combine these two log statements?

ols/app/models/config.py
```diff
@@ -311,6 +312,53 @@ def validate_yaml(self) -> None:
         """Validate memory cache config."""
 
 
+class Filter(BaseModel):
+    """Filter configuration."""
```
syedriko, line 316: QueryFilter?

ols/app/models/config.py
```diff
@@ -467,6 +515,7 @@ class OLSConfig(BaseModel):
 
     default_provider: Optional[str] = None
     default_model: Optional[str] = None
+    question_filters: Optional[list[Filter]] = None
```
syedriko, line 518: query_filters?

ols/utils/question_filter.py
```diff
@@ -0,0 +1,51 @@
...
+
+class QuestionFilter:
+    """Redact the question based on the regex filters provided in the config file."""
+
+    def __init__(self, regex_filters: Optional[list[RegexFilter]]) -> None:
+        """Initialize the class instance."""
+        self.regex_filters = regex_filters
+
+    @classmethod
+    def setup(cls) -> "QuestionFilter":
```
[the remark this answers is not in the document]
xrajesh: We could have it in __init__ . Thought that its only one place that we will create that instance, hence created a Setup method . But __init__ would do . updated the code .

ols/app/endpoints/ols.py
```diff
@@ -172,8 +176,30 @@ def store_conversation_history(
...
+        )
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail={"response": f"Error while redacting query '{redactor_error}'"},
+        )
+
+
 def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
     """Validate user question, raise HTTPException in case of any problem."""
+    # Validate the query
```
bparees, line 202: ?

seems unrelated to your changes.

bparees:
/approve

one last nit that i see, but will let the other reviewers take another pass also to ensure their comments are satisfactorily addressed

tests/unit/utils/test_config.py
```diff
@@ -825,3 +826,40 @@ def test_config_file_without_logging_config():
...
+                name="foo_filter",
+                replace_with="openshift",
+            ),
+            RegexFilter(
+                pattern=re.compile(r"\b(?:bar)\b"),
+                name="bar_filter",
+                replace_with="kubernetes",
+            ),
+        ]
+    except Exception:
```
tisnik, line 863: dtto

tisnik (commented):
LGTM, just two nitpicks/questions...

ols/utils/query_filter.py
```diff
@@ -0,0 +1,42 @@
...
+from collections import namedtuple
+
+from ols.utils import config
+
+logger = logging.getLogger(__name__)
+
+RegexFilter = namedtuple("RegexFilter", "pattern, name, replace_with")
+
+
+class QueryFilter:
```
syedriko, line 14: Nit: What we call a QueryFilter in config is a RegexFilter here.

xrajesh: @syedriko I think this is okay, thats the name of the namedtuple. Would leave as it is.

PR #429, opened 2024-02-16
OLS-114: Labelled metric

tisnik:
## Description

Labelled metric

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #OLS-114

### DIFF
--- a/ols/app/main.py
+++ b/ols/app/main.py
@@ -41,12 +41,13 @@ else:
 @app.middleware("")
 async def rest_api_counter(request: Request, call_next):
     """Middleware with REST API counter update logic."""
+    response = await call_next(request)
     # ignore /metrics endpoint that will be called periodically
     path = request.url.path
     if not path.endswith("/metrics/"):
         # just update metrics
-        metrics.rest_api_calls.inc()
-    return await call_next(request)
+        metrics.rest_api_calls_total.labels(response.status_code).inc()
+    return response
 
 
 routers.include_routers(app)
--- a/ols/app/metrics.py
+++ b/ols/app/metrics.py
@@ -2,7 +2,10 @@
 
 from prometheus_client import Counter, make_asgi_app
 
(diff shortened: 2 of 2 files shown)


ols/app/main.py
```diff
@@ -41,12 +41,13 @@
...
 async def rest_api_counter(request: Request, call_next):
     """Middleware with REST API counter update logic."""
+    response = await call_next(request)
     # ignore /metrics endpoint that will be called periodically
     path = request.url.path
     if not path.endswith("/metrics/"):
         # just update metrics
-        metrics.rest_api_calls.inc()
-    return await call_next(request)
+        metrics.rest_api_calls_total.labels(response.status_code).inc()
```
bparees, line 49: seems like we should included the path in the labels, so we can differentiate the types of api calls being made.

tisnik: yup that would be super easy. I'll create PR later today

raptorsun: A reminder that we should put the path leading to the endpoint function without query parameters, for example: 
we label with  `/version/endpoint`  instead of `/version/endpoint?arg1=val1...` or `/version/endpoint/arg1/val1`
A distinct value in the path label is an additional cardinality to the metric, making the Prometheus consumes more storage and computation.

tisnik: @bparees https://github.com/openshift/lightspeed-service/pull/432

tisnik: > A reminder that we should put the path leading to the endpoint function without query parameters, for example: we label with `/version/endpoint` instead of `/version/endpoint?arg1=val1...` or `/version/endpoint/arg1/val1` A distinct value in the path label is an additional cardinality to the metric, making the Prometheus consumes more storage and computation.

path does not contain query params

raptorsun: great, that simplifies our work :)

PR #431, opened 2024-02-16
OLS-246: Use llamaindex only as retriever

asamal4:
## Description

- Use llamaindex only for retrieving index and matching data based on query.
- Use langchain for LLM call.
- Integrate with token truncation.
- Some refactoring.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-246](https://issues.redhat.com//browse/OLS-246)
- Closes #

## Testing

### DIFF
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -5,28 +5,22 @@ import os
 from typing import Any, Optional
 
 import llama_index
+from langchain.chains import LLMChain
 from llama_index import ServiceContext, StorageContext, load_index_from_storage
-from llama_index.postprocessor import BaseNodePostprocessor
-from llama_index.prompts import PromptTemplate
-from llama_index.response.schema import Response
-from llama_index.schema import NodeWithScore, QueryBundle
+from llama_index.indices.vector_store.base import VectorStoreIndex
 
 from ols import constants
+from ols.src.prompts.prompts import CHAT_PROMPT
 from ols.src.query_helpers import QueryHelper
 from ols.utils import config
+from ols.utils.token_handler import (
+    # TODO: Use constants from config
+    CONTEXT_WINDOW_LIMIT,
+    RESPONSE_WINDOW_LIMIT,
+    TokenHandler,
+)
 
 logger = logging.getLogger(__name__)
-
-
-class _NoLLMMetadataPostprocessor(BaseNodePostprocessor):
-    """A LlamaIndex PostProcessor that prevents node metadata from reaching the LLM."""
-
-    def _postprocess_nodes(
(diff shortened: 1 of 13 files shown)


ols/src/query_helpers/docs_summarizer.py
```diff
@@ -5,30 +5,23 @@
...
-from llama_index.response.schema import Response
-from llama_index.schema import NodeWithScore, QueryBundle
+from llama_index.indices.vector_store.base import VectorStoreIndex
 
 from ols import constants
+from ols.src.prompts.prompts import CHAT_PROMPT
 from ols.src.query_helpers import QueryHelper
 from ols.utils import config
+from ols.utils.token_handler import (
+    CONTEXT_WINDOW_LIMIT,
```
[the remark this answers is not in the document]
tisnik: sure, yes, we can merge, and work on integration next week

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -127,56 +157,41 @@ def summarize(
...
 
-        if use_llm_without_reference_content:
+        rag_context, referenced_documents = self._format_rag_data(rag_context_data)
+
+        chat_engine = LLMChain(
+            llm=bare_llm,
+            prompt=CHAT_PROMPT,
+            verbose=verbose,
+        )
+        summary = chat_engine.invoke({"context": rag_context, "query": query})
```
bparees, line 176: it seems worthy of note that you're using the same prompt template regardless of whether there actually is any context data to include or not.

while real installs will always have a rag db available, it's not hard to imagine that in the case of some questions all the returned RAG content is so low scoring that we don't want to include any of it. So should we have a separate prompt for cases where there is no rag results to inject?

bparees: did you test this with both BAM and OpenAI providers?  i'm a little surprised the same code works for both, given the TODO statement/hack below that you're removing.

bparees: > If content is not available then it will try to answer from the data it was trained on (if that is valid/relevant)..

I understand, but the prompt is going to say something like `using the information below, answer the user's question`, except there won't actually be any "information below" if we don't have a RAG db configured or there's no retrieved content we want to include.  So wouldn't it be better to have a different prompt when we are not adding retrieved context?

bparees:
follow up items on this:
1) using a different prompt template when we are not injecting any rag documents (https://github.com/openshift/lightspeed-service/pull/431/files#r1494851545)
2) potentially pivoting to langchain from llama index if we adopt FAISS for our embedding storage format
3) we may want to move the rag functions out to their own module rather than putting them in "docs_summarizer"

/approve
/lgtm

PR #441, opened 2024-02-19
Use plain old teste function instead of methods where appropriate

tisnik:
## Description

Use plain old teste function instead of methods where appropriate

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/unit/app/models/test_config.py
+++ b/tests/unit/app/models/test_config.py
@@ -563,41 +563,39 @@ def test_llm_providers_equality():
     assert provider_config_1 != other_value
 
 
-class TestLoggingConfig:
-    """Test the LoggingConfig model."""
+def test_valid_values():
+    """Test valid values."""
+    # test default values
+    logging_config = LoggingConfig({})
+    assert logging_config.app_log_level == logging.INFO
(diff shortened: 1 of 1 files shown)


onmete:
I see you don't like the classes there. :)

Could you just remove the class and have the methods as plain test functions instead?

tisnik:
> I see you don't like the classes there. :)
> 
> Could you just remove the class and have the methods as plain test functions instead?

Well both are correct. But ATM when there are no shared state, it does not make much sense to use instance methods, because static methods are ok/sufficient/more optional.

tisnik:
> @tisnik i'm pretty sure i've seen people creating PRs going the other way (wrapping test functions into test classes) too, so let's make sure we're not fighting each other on these patterns.

So should we update contributing guide?

bparees:
> So should we update contributing guide?

i don't know if it will help :)  for now i'll settle for having raised awareness.  sorry, wish i could point to the other PR i am thinking of.

PR #445, opened 2024-02-19
OLS-311: Ability to start service from Python script

tisnik:
## Description

Ability to start service from Python script

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-311](https://issues.redhat.com//browse/OLS-311)

## Testing
- Simply start the new script: `python runner.py`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ install-deps-test: install-tools ## Install all required dependencies needed to 
 	pdm install --dev
 
 run: ## Run the service locally
-	uvicorn ols.app.main:app --reload --port 8080
+	python runner.py
 
 test: test-unit test-integration test-e2e ## Run all tests
 
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
 ### Run the server
 in order to run the API service
 ```sh
-uvicorn ols.app.main:app --reload --port 8080
+make run
 ```
 
 ### Optionally run with podman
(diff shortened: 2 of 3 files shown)


onmete:
You can also use the VSCode launcher for FastAPI. It looks like this
```
{
    "name": "Python: FastAPI",
    "type": "python",
    "request": "launch",
    "module": "uvicorn",
    "args": [
        "ols.app.main:app",
        "--reload",
        "--port",
        "8080"
    ],
    "jinja": true,
    "justMyCode": true
}
```

Can we consider this as a dev feature and maybe just add it to contrib docs and gitignore? Instead of having dev related code in root?

bparees:
> Can we consider this as a dev feature and maybe just add it to contrib docs and gitignore? Instead of having dev related code in root?

@onmete  this isn't dev related code, we're going to pivot to launching OLS this way.  Do you have concerns about that?

The primary motivation is so that we have a proper place to put our env var initialization for env vars we need to set before OLS starts up.  I believe @syedriko had some items to be put there (see https://issues.redhat.com/browse/OLS-345)

@tisnik the code looks fine but are you intending to update our scripts/makefiles/docs to reference this approach?

onmete:
> we're going to pivot to launching OLS this way

no objections than

/lgtm

tisnik:
> @tisnik the code looks fine but are you intending to update our scripts/makefiles/docs to reference this approach?

Yup, I just wanted to know if it's the right direction. I'll continue working on it knowing it's ok-ish :)

tisnik:
@bparees updated Makefile + doc (e2e test script use `make run` already)

PR #446, opened 2024-02-19
OLS-233: Llm loader modularity

onmete:
## Description

Split LLMLoader into modules + ability to easily define new providers.

There are various ways how to achieve plugins in python. I didn't want to hardcode a switch for each provider or import those explicitly, so I used dynamic loading.
This creates some extra complexity, but the benefit is that whoever wants to experiment with a new provider, he/she can just create a new module eg. `ols.src.llms.providers.my_provider.py` and it will be auto registered without the need to change anything else elsewhere. So this "dev" change won't affect the rest of our codebase. It should be easier for the person to play in the area without a constant need of conflicts.

LLM params evaluation order is dev config params > params provided to LLMLoader > default provider params. But I guess this is a subject of broader params discussion/change.

## Type of change

- [x] Refactor
- [x] New feature


ols/src/llms/providers/openai.py
```diff
@@ -0,0 +1,48 @@
...
+from langchain_openai import ChatOpenAI
+
+from ols import constants
+from ols.src.llms.providers import LLMProvider, register_llm_provider_as
+
+logger = logging.getLogger(__name__)
+
+
+@register_llm_provider_as(constants.PROVIDER_OPENAI)
+class OpenAI(LLMProvider):
```
onmete, line 15: The main purpose of having the `LLMProvider` class now is to process the params. Currently, we don't know if and how we want to do that, so the provider definition can be a simple function too.

pyproject.toml
```diff
@@ -15,7 +15,7 @@ flake8-pytest-style.fixture-parentheses = false
 
 [tool.coverage.report]
 # unit tests fails if the total coverage measurement is under this threshold value
-fail_under = 90
+fail_under = 85
```
onmete, line 18: leftover

ols/src/llms/llm_loader.py
```diff
@@ -67,13 +55,15 @@ def __init__(
         provider: str,
         model: str,
         llm_params: Optional[dict] = None,
+        _llm_providers: Optional[LLMProviders] = None,
+        _llm_providers_reg: Optional[LLMProvidersRegistry] = None,
```
onmete, line 59: I put those as options, but regarding our last discussion @bparees, I can just set those as defaults in init.

onmete (commented):
@bparees @tisnik Please validate if this would be the way to go.

If so, I update docs and tests too.

ols/src/llms/llm_loader.py
```diff
@@ -33,86 +33,42 @@ class ModelConfigMissingError(LLMConfigurationError):
...
+        )
 
-        bare_llm = LLMLoader(provider="openai", model="gpt-3.5-turbo", llm_params=llm_params).llm
-        llm_chain = LLMChain(llm=bare_llm, prompt=prompt)
-        ```
-    """
+    return provider_config
+
+
+def load_llm(provider: str, model: str, llm_params: Optional[dict]) -> LLM:
```
onmete, line 58: Replaced LLMLoader with func (needs polishing). WDYT?

tisnik: it's IMHO much better solution

ols/src/llms/providers/openai.py
```diff
@@ -0,0 +1,48 @@
...
+    """OpenAI provider."""
+
+    @property
+    def default_params(self):
```
onmete, line 19: It is out of the scope of this PR, but regarding the llm parameters. We can probably use these `default_params` (or rename it to `allowed_params` with defaults) to validate incoming parameters and drop those that are not present in these defaults/allowed.
If we decide the whitelist is how we want to handle this problem.

ols/src/llms/providers/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @abc.abstractmethod
+    def load(self) -> LLM:
+        """Load and langchain `LLM` instance and return it."""
+
+
+class LLMProvider(AbstractLLMProvider):
+    """LLM provider."""
+
+    def __init__(self, model: str, params: dict, provider_config: ProviderConfig):
```
bparees, line 68:
```suggestion
    def __init__(self, model: str, params: Optional[dict], provider_config: ProviderConfig):
```

?

tests/unit/llms/test_llm_loader.py
```diff
@@ -36,25 +36,9 @@ def test_constructor_unknown_provider():
...
-        }
-    )
-    config.config.llm_providers.providers[test_provider].type = test_provider_type
-    config.llm_config = config.config.llm_providers
+    providers = LLMProviders([{"name": "some-provider", "models": [{"name": "model"}]}])
+    config.init_empty_config()
+    config.config.llm_providers = providers
 
     with pytest.raises(UnsupportedProviderError):
         LLMLoader(provider=test_provider, model=constants.GRANITE_13B_CHAT_V1)
```
bparees, line 44:
```suggestion
    load_llm(provider=test_provider, model=constants.GRANITE_13B_CHAT_V1)
```

?

bparees: a few other spots too

bparees (commented):
this looks much better, thank you!  just a few small items

ols/src/llms/providers/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+from langchain.llms.base import LLM
+
+from ols.app.models.config import ProviderConfig
+from ols.utils import config
+
+logger = logging.getLogger(__name__)
+
+
+class LLMProvidersRegistry:
```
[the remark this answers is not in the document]
onmete: didn't noticed that one, I'll fix this

tests/unit/llms/test_llm_loader.py
onmete: Yup, I should have mentioned that pushed just the code for approval (without tests update).

ols/src/llms/llm_loader.py
```diff
@@ -1,32 +1,19 @@
...
 from ols.app.models.config import ProviderConfig
+from ols.src.llms.providers.registry import LLMProvidersRegistry
 from ols.utils import config
 
 logger = logging.getLogger(__name__)
 
 
 # workaround to disable UserWarning
-warnings.simplefilter("ignore", UserWarning)
+# warnings.simplefilter("ignore", UserWarning)
```
onmete, line 16: I didn't find a difference when set and when commented, so I would be for removing.
Do you know what is the reason behind?

tisnik:
> @tisnik are there any other type-related issues? Why those aren't discovered by mypy?

it's ok

ols/src/llms/llm_loader.py
tisnik: I'd remove it, and also remove the `import warnings`

ols/src/llms/llm_loader.py
```diff
@@ -45,292 +28,62 @@ class ModelConfigMissingError(LLMConfigurationError):
...
-            self.provider_config.models[self.model].url
-            if self.provider_config.models[self.model].url is not None
-            else (
-                self.provider_config.url
-                if self.provider_config.url is not None
-                else default
-            )
+    if provider not in providers_config.providers:
+        raise UnknownProviderError(
+            f"Provider '{provider}' is not a valid provider! "
```
bparees, line 38: we should not be using exclamation marks in error messages, either periods or no punctuation.  It's already an error, there's no reason to further alarm or accuse the user.

bparees:
@onmete seems there is an issue w/ these changes working w/ watsonx: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/446/pull-ci-openshift-lightspeed-service-main-e2e-ols-watsonx-granite13b-chat-v2/1760373276355334144

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/446/pull-ci-openshift-lightspeed-service-main-e2e-ols-watsonx-granite13b-chat-v2/1760373276355334144/artifacts/e2e-ols-watsonx-granite13b-chat-v2/e2e/artifacts/ols.log

looks like maybe the api url isn't getting passed through correctly.

onmete:
It was the project_id. It was looking for that from params instead of config.

/test e2e-ols-watsonx-granite13b-chat-v2

onmete:
And yes, the I removed the default URLs. Seems to work for openai and bam, but watson is probably not defined in some ci config. I've re-enabled it in the last commit. 

/test e2e-ols-watsonx-granite13b-chat-v2

PR #459, opened 2024-02-21
Explicit return from `try` block

tisnik:
## Description

The `try`-`except` statement has an `else` clause for code that should
run _only_ if no exceptions were raised. Using the `else` clause is more
explicit than using a `return` statement inside of a `try` block.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -186,7 +186,6 @@ def redact_query(conversation_id: str, llm_request: LLMRequest) -> LLMRequest:
         llm_request.query = config.query_redactor.redact_query(
             conversation_id, llm_request.query
         )
-        return llm_request
     except Exception as redactor_error:
         logger.error(
             f"Error while redacting query {redactor_error} for conversation {conversation_id}"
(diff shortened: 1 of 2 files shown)


onmete:
I can't decide if it is better or worse in terms of code readability. I'll let others to lgtm this one.

bparees:
currently the expected flow of the code would be:

1) do some work
2) return some value
3) if an exception happened in (1), handle it.  This should be unusual.

changing it to:

1) do some work
2) check if an exception happened + handle it
3) return some value from (1)

imho makes it less readable because the return statement is separated from the main logic that setup/computed the return value.  For someone who's reading the code to understand the happy path, they need to recognize that they need to go look at the "else" statement to see what additional steps are going to be executed.

personally when i'm trying to understand code i usually stop reading when i get to the exception handling portion (assuming i don't have a reason to care about it).

all in all it's a bikeshed, but i'm a -1 on this structure.

bparees:
I guess i'll also add that this change seems to violate the principal of "earliest possible return":
https://www.linkedin.com/pulse/early-return-dip-ghosh/

but again, that's to some degree a personal preference, certainly not a hard+fast rule.

bparees:
@tisnik do you want to make a case for this or can we close it?

tisnik:
We can close it, won't be relevant anymore.

PR #466, opened 2024-02-21
Unit test for reduct_query function for no filters, one filter, and two filters

tisnik:
## Description

Unit test for `reduct_query` function (happy path that was not covered)

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

### DIFF
--- a/tests/unit/app/endpoints/test_ols.py
+++ b/tests/unit/app/endpoints/test_ols.py
@@ -140,8 +140,20 @@ def test_validate_question_on_validation_error(validate_question_mock, load_conf
 
 
 @patch("ols.utils.config.query_redactor")
+def test_query_filter_on_correct_reduct(mock_redact_query, load_config):
+    """Test the function to redact query."""
+    conversation_id = suid.get_suid()
+    query = "Tell me about Kubernetes"
+    llm_request = LLMRequest(query=query, conversation_id=conversation_id)
+    mock_redact_query.redact_query.return_value = "Mocked query"
+    result = ols.redact_query(conversation_id, llm_request)
+    assert result is not None
+    assert result.query == "Mocked query"
+
+
+@patch("ols.utils.config.query_redactor")
 def test_query_filter_on_redact_error(mock_redact_query, load_config):
-    """Test conversation request API endpoint."""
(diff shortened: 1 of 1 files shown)


tests/unit/app/endpoints/test_ols.py
```diff
@@ -139,9 +139,21 @@ def test_validate_question_on_validation_error(validate_question_mock, load_conf
         ols.validate_question(conversation_id, llm_request)
 
 
+@patch("ols.utils.config.query_redactor")
+def test_query_filter_on_correct_reduct(mock_redact_query, load_config):
```
bparees, line 143:
```suggestion
def test_query_filter_on_correct_redact(mock_redact_query, load_config):
```

tests/unit/app/endpoints/test_ols.py
```diff
@@ -139,9 +139,21 @@ def test_validate_question_on_validation_error(validate_question_mock, load_conf
...
+    """Test the function to redact query."""
+    conversation_id = suid.get_suid()
+    query = "Tell me about Kubernetes"
+    llm_request = LLMRequest(query=query, conversation_id=conversation_id)
+    mock_redact_query.redact_query.return_value = "Mocked query"
+    result = ols.redact_query(conversation_id, llm_request)
+    assert result is not None
+    assert result.query == "Mocked query"
```
bparees, line 151: this is another case of "i don't really see the point of this test".  It seems to be just checking the return type of ols.redact_query.

Can't we have a test that:
1) actually verifies that the redaction happens
2) *and* confirms the return type

together?  This test alone hardly seems worth having+maintaining.  If ols.redact_query starts returning a different object, i'd expect us to quickly notice that when everything else starts failing.

tisnik: ok that sounds a bit like a plan :)

tisnik: now it might be better @bparees , WDYT?

PR #467, opened 2024-02-21
Disable Gradio telemetry as soon as possible

tisnik:
## Description

Disable Gradio telemetry as soon as possible

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -1,6 +1,12 @@
 """Script that is able to start Uvicorn-based REST API service."""
 
+import os
+
 import uvicorn
 
 if __name__ == "__main__":
+    # TODO syedriko consolidate this env var into a central location as per OLS-345.
+    # disable Gradio analytics, which calls home to https://api.gradio.app
+    os.environ["GRADIO_ANALYTICS_ENABLED"] = "false"
+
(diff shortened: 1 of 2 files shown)


runner.py
```diff
@@ -1,6 +1,12 @@
 """Script that is able to start Uvicorn-based REST API service."""
 
+import os
+
 import uvicorn
 
 if __name__ == "__main__":
+    # TODO syedriko consolidate this env var into a central location as per OLS-345.
+    # disable Gradio analytics, which calls home to https://api.gradio.app
+    os.environ["GRADIO_ANALYTICS_ENABLED"] = "false"
```
bparees, line 10: @syedriko wasn't there at least one other env var we wanted to set related to some phone-home behavior?

syedriko: There're two more at, they affect HuggingFace libraries
https://github.com/openshift/lightspeed-service/blob/b23c5940e8b5cd7afa6c24060496385b9bfffeb4/ols/src/query_helpers/docs_summarizer.py#L214

We can move TRANSFORMERS_OFFLINE=1 now, but the other one depends on config.

PR #469, opened 2024-02-21
OLS-159: add metrics counting sent/received LLM tokens

raptorsun:
## Description

This PR adds 2 metrics:
- `llm_token_sent_total` with 2 labels `provider` and `model`
- `llm_token_received_total` with 2 labels `provider` and `model`

It also adds the  label `provider` and `model` to metric `llm_calls_total`, so that the token counts are seperated for each provider-model combo.

## Type of change

- [ ] Refactor
- [x ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes [OLS-159](https://issues.redhat.com/browse/OLS-159)

## Testing


ols/app/metrics.py
```diff
@@ -10,11 +10,18 @@
...
 )
 
-llm_calls_total = Counter("llm_calls_total", "LLM calls counter")
+llm_calls_total = Counter("llm_calls_total", "LLM calls counter", ["provider", "model"])
 llm_calls_failures_total = Counter("llm_calls_failures_total", "LLM calls failures")
 llm_calls_validation_errors_total = Counter(
     "llm_validation_errors_total", "LLM validation errors"
 )
 
+llm_token_sent_total = Counter(
```
tisnik, line 19: is Counter the best option? Probably histogram might be better fit?

ols/src/llms/utils.py
```diff
@@ -0,0 +1,42 @@
+from langchain.callbacks.base import BaseCallbackHandler
```
tisnik, line 1: I'd rather not to use `utils.py  again please. It contains implementation of `GenericTokenCounter`

tisnik (commented):
left comments. So far looks good...

ols/src/llms/utils.py
```diff
@@ -0,0 +1,42 @@
...
+                # typical token_usage: {'prompt_tokens': 252, 'completion_tokens': 4, 'total': 256}
+                if 'prompt_tokens' in token_usage:
+                    input_tokens_count += token_usage['prompt_tokens']
+
+                if  'completion_tokens' in token_usage:
+                    self.output_tokens += token_usage['completion_tokens']
+                else:
+                    self.output_tokens += self.llm.get_num_tokens(r.generations[0][0].text)
+
+                continue
```
tisnik, line 35: interesting nested blocks, will be a bit tricky to test it

onmete: Maybe just
```python
....
else:
   # fallback to counting
   self.output_tokens += self.llm.get_num_tokens(r.generations[0][0].text)
```
instead of `continue` would be a bit more readable.

ols/src/llms/utils.py
onmete: I would propose something like
```
ols
  /metrics
     metrics.py   # moved from ols/app
     token_counter.py
```

ols/app/endpoints/ols.py
```diff
@@ -234,7 +233,22 @@ def generate_bare_response(conversation_id: str, llm_request: LLMRequest) -> str
 
     prompt = PromptTemplate.from_template("{query}")
     llm_chain = LLMChain(llm=bare_llm, prompt=prompt, verbose=True)
-    response = llm_chain(inputs={"query": llm_request.query})
+    token_counter = GenericTokenCounter(bare_llm)
```
onmete, line 236: This code is more or less repeated in 3 places. It would change the implementation, but what about having it as a context manager? Something like:
```python
with metrics.token_counter(bare_llm):
   llm_chain = LLMChain(llm=bare_llm, prompt=prompt, verbose=True)   
   response = llm_chain(inputs={"query": llm_request.query})
```

ols/src/llms/utils.py
```diff
@@ -0,0 +1,42 @@
+from langchain.callbacks.base import BaseCallbackHandler
+from langchain_core.outputs.llm_result import LLMResult
+from typing import Any, Dict, List
+import logging
+
+logger = logging.getLogger(__name__)
+
+class GenericTokenCounter(BaseCallbackHandler):
```
onmete, line 8: Do you plan to add a docstring here? Example usage might be useful to see here.

ols/src/llms/utils.py
raptorsun: will move to `ols/metrics/token_counter.py`

ols/app/metrics.py
raptorsun: Couter is the cheapest option to meet the needs for [OLS-159](https://issues.redhat.com/browse/OLS-159) :) 
Of course the histogram give us more insights on the token count distribution in its `_buckets` metrics. 
Do we need these details about token count? @bparees ?
Available metrics types: https://prometheus.io/docs/tutorials/understanding_metric_types/

ols/src/llms/utils.py
raptorsun: yes, i should have added a docstring here. will add in new commit

ols/src/llms/utils.py
raptorsun: Let me add a unit test about these callbacks in new commit. 
I tend to use less indentation cause I like put editors side by side on my 4:3 screen ;) I will use `else` block in the update.

ols/app/endpoints/ols.py
raptorsun: I will change GenericTokenCounter that it updates the metrics by the end of the callback function.

ols/app/metrics.py
raptorsun: as discussed in standup meeting, we keep using a simple Counter for the moment.

tisnik (commented):
it looks good!

bparees:
/approve

as a follow up i'd like to see an e2e test that:

1) checks the metrics endpoint to see the metrics exist w/ counts of 0
1) invokes a query
2) checks the metrics endpoint to confirm the metrics are updated (you don't have to check for exact values, that will be too fragile, especially when we swap providers, but you can confirm the values were updated and fall within some reasonable range of values)

PR #473, opened 2024-02-22
Refactor: simpler return statement, not need to use intermediate vari…

tisnik:
## Description

Refactor: simpler `return` statement, not need to use intermediate variable

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -84,6 +84,4 @@ def load_llm(provider: str, model: str, llm_params: Optional[dict] = None) -> LL
     logger.debug(f"loading LLM '{model}' from '{provider}'")
 
     llm_provider = llm_providers_reg.llm_providers[provider_config.type]
-    llm = llm_provider(model, provider_config, llm_params or {}).load()
-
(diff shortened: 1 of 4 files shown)


onmete:
I'm doing that on purpose. With that, I can set the debug point on return and see what it is returning. With this, I would set the point on return and need to evaluate the statement in the debugger.

But I don't have any argument besides that. 🤷‍♂️

tisnik:
> I'm doing that on purpose. With that, I can set the debug point on return and see what it is returning. With this, I would set the point on return and need to evaluate the statement in the debugger.
> 
> But I don't have any argument besides that. 🤷‍♂️

if you put breakpoint at return statement, you will see the computed return value too

tisnik:
> where? image

fff

onmete:
Well, yea, but you need to be one level above.

As I said, it's my personal thing. I'll try to remember it from now on.

/lgtm

PR #484, opened 2024-02-23
Adding TLS and removing username from RedisConfig

vishnuchalla:
## Description

Removing username from redis configuration. By default redis username is `default` until unless we need some ACLs (Access control lists) which is not required for our usecases. Also adding support for TLS certs.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Closes:
  * https://issues.redhat.com/browse/OLS-221
  * https://issues.redhat.com/browse/OLS-289
  * https://issues.redhat.com/browse/OLS-288

## Testing
Tested this change integration with operator.
sample query:
```
[vchalla@vchalla-thinkpadp1gen2 lightspeed-service]$ oc rsh pod/lightspeed-app-server-5474c9b97-pqp2d
(venv) sh-5.1$ curl -X 'POST' 'http://0.0.0.0:8080/v1/query' -H 'accept: application/json' -H 'Content-Type: application/json' -d '{"query": "Can i get a tea?"}'
{"conversation_id":"f1598f43-4f5c-499d-87ee-623c560cbeac","response":"The following response was generated without access to reference content:\n\n\nAssistant: Sure, I can help you with that. What kind of tea would you like?\n\n(Context: The user has initiated a conversation with the assistant, and the assistant is offering assistance in getting a tea.)","referenced_documents":[],"truncated":false}(venv) sh-5.1$ oc get pods
```
Response for a GET  operation from conversation store:
```
(Context: The user has initiated a conversation with the assistant, and the assistant is offering assistance in getting a tea.) Stored response from conversation cache
```


vishnuchalla:
Make verify works in local but says 1 file needs to formatted in CI. Looks like it needs some investigation to figure that out.

tisnik:
@vishnuchalla please make sure you have installed the rigtht version of Black

vishnuchalla:
> @vishnuchalla please make sure you have installed the rigtht version of Black

Yes, I remember installing black previously outside OLS. That version might be having a different functionality than ours. Will take a look again.

ols/app/models/config.py
```diff
@@ -219,31 +219,35 @@ def validate_yaml(self) -> None:
...
     def __init__(self, data: Optional[dict] = None) -> None:
-        """Initialize redis credentials."""
+        """Initialize redis TLS certificates."""
         super().__init__()
         if not isinstance(data, dict):
             return
-        self.username = _get_attribute_from_file(data, "username_path")
-        self.password = _get_attribute_from_file(data, "password_path")
+        self.cert_path = data.get("cert_path")
+        self.key_path = data.get("key_path")
```
bparees, line 234: why does OLS need the redis tls key?  it should only need to be able to verify the cert, and it should be able to verify the cert just based on it being signed by the OCP CA.  So it shouldn't need either of these things to be able to make a trusted TLS connection to Redis.

bparees (commented):
i don't understand why OLS needs this TLS config info

ols/app/models/config.py
vishnuchalla: Sorry this was one my `no-auth` experiments that i have accidentally committed. Yes we don't require TLS config in OLS. OLS as a client just needs to have a CA cert with it to trust the identity of the redis server and talk to it. Redis server is configured with TLS cert, key and CA cert. Certificate verification has been made mandatory to make sure the communication is encrypted between them.

ols/src/cache/redis_cache.py
```diff
@@ -35,7 +35,7 @@ def initialize_redis(self, config: RedisConfig) -> None:
         if config.password is not None:
             kwargs["password"] = config.password
         if config.ca_cert_path is not None:
-            kwargs["ssl"] = True
+            kwargs["ssl"] = True  # type: ignore
```
bparees, line 38: what's the issue here?  should we be setting types on kwargs instead?  (type Any?)

cc @tisnik

vishnuchalla: I am running into this error in one of the verify steps
```
mypy --explicit-package-bases --disallow-untyped-calls --disallow-untyped-defs --disallow-incomplete-defs ols/
ols/src/cache/redis_cache.py:38: error: Incompatible types in assignment (expression has type "bool", target has type "str")  [assignment]
```
So I cross checked the redis dependency version and source code: https://redis.readthedocs.io/en/latest/_modules/redis/client.html, it says `bool` but I am not sure why our CI verify step is expecting it to be `str`. Same thing doesn't happen in my local again. I tried aligning the dependency versions with mainstream, failed in them and finally ended up putting an ignore comment beside it.

bparees: I assume it's because you're also putting string content into the dict, so mypy expects the all the types to be the same.  you probably need to give type hints when you construct the dict to indicate the type is str OR bool.

again @tisnik can probably help better than me here

bparees: i think it would be something like `a: dict[str, str|bool] = {}`

tisnik: yes exactly. Mypy is doing type inference (like in ML, OCaml, F# and in less ideal way in Go). So in this case you need to specify explicitly that kwargs is `dict[str, Any]` or `dict[str, str | bool]`

tisnik: ah @bparees was faster ;)

vishnuchalla: Updated. Thanks

PR #503, opened 2024-02-28
Unit test code coverage set to 95%

tisnik:
## Description

Unit test code coverage set to 95%

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update


bparees:
/hold

let's discuss this.  I have concerns about the cost+value setting the coverage expectations this high vs other places the team can invest effort.  In particular i'm concerned that we are not getting enough effort spent on higher level testing(e2e tests that confirm the stack actually works for the the functionality users will be exposed to)

cc @joshuawilson

PR #504, opened 2024-02-28
Rebuiling redis image with rpm packages

vishnuchalla:
## Description

Rebuilding redis with rpm packages.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Closes # https://issues.redhat.com/browse/OLS-312

## Testing
Tested integrated with this PR: https://github.com/openshift/lightspeed-operator/pull/23


Containerfile.redis
```diff
@@ -1,33 +1,17 @@
 # Use Red Hat Universal Base Image (UBI) 9 as the base image
 FROM registry.access.redhat.com/ubi9/ubi-minimal
 
-# Install necessary dependencies (gcc, make, etc.) for building Redis from source
-RUN microdnf install -y gcc make wget tar gzip && \
-    microdnf clean all
+# Install necessary dependencies for downloading and installing Redis
+RUN microdnf install -y wget yum-utils logrotate systemd && \
+    wget https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/redis-7.0.12-1.module_el9+499+353c08ce.x86_64.rpm && \
```
bparees, line 6: better.  wish we could just dnf install the rpm from the repo directly but it doesn't seem to show up for some reason.

bparees: (even when i tried using a centos streams base image and dnf installing into that).  maybe @williamcaban has an idea

Containerfile.redis
```diff
@@ -1,33 +1,17 @@
 # Use Red Hat Universal Base Image (UBI) 9 as the base image
 FROM registry.access.redhat.com/ubi9/ubi-minimal
 
-# Install necessary dependencies (gcc, make, etc.) for building Redis from source
-RUN microdnf install -y gcc make wget tar gzip && \
-    microdnf clean all
+# Install necessary dependencies for downloading and installing Redis
+RUN microdnf install -y wget yum-utils logrotate systemd && \
```
syedriko, line 5: Why are we installing logrotate? Redis logs should be going into stdout/stderr.

Containerfile.redis
```diff
@@ -1,33 +1,17 @@
...
 
-# Build Redis from source
-RUN make && \
-    make install
-
-# Set protected-mode to no and bind to all interfaces (0.0.0.0) in redis.conf
-RUN sed -i 's/^protected-mode yes/protected-mode no/' redis.conf
-RUN sed -i 's/^bind 127.0.0.1 -::1/bind 0.0.0.0/' redis.conf
+# Grant necessary permissions
+RUN chmod -R 777 /var/lib/redis /var/log/redis /etc/redis
```
syedriko, line 14: It would be better to move these to the RUN command above so as not to create an extra image layer. Also, the perms seem too permissive. On the cluster, with the restricted-v2 SCC uid is whatever but gid is 0, would 770 be enough?

Containerfile.redis
```diff
@@ -1,33 +1,17 @@
...
 FROM registry.access.redhat.com/ubi9/ubi-minimal
 
-# Install necessary dependencies (gcc, make, etc.) for building Redis from source
-RUN microdnf install -y gcc make wget tar gzip && \
-    microdnf clean all
+# Install necessary dependencies for downloading and installing Redis
+RUN microdnf install -y wget yum-utils logrotate systemd && \
+    wget https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/redis-7.0.12-1.module_el9+499+353c08ce.x86_64.rpm && \
+    rpm -i redis-7.0.12-1.module_el9+499+353c08ce.x86_64.rpm && \
+    rm -f redis-7.0.12-1.module_el9+499+353c08ce.x86_64.rpm
```
syedriko, line 8: Could uninstall wget here.

Containerfile.redis
tisnik: `/etc/redis` might be just `+r`, not `+w` nor `+x`

Containerfile.redis
vishnuchalla: Yes. But redis itself requires logrotate as a dependency. If we try to build this image without it, we will run into this error
```
error: Failed dependencies:
	logrotate is needed by redis-6.2.7-1.el9.x86_64
```

Containerfile.redis
vishnuchalla: Added cleanup commands too. Thanks

Containerfile.redis
vishnuchalla: Acknowledged. Combined these with the above RUN command and granted only the required perms.

Containerfile.redis
```diff
@@ -1,39 +1,18 @@
...
-    tar xzf redis-${REDIS_VERSION}.tar.gz -C /etc/redis --strip-components=1 && \
-    rm redis-${REDIS_VERSION}.tar.gz && \
-    chmod 777 /var/run
-
-# Change working directory to the extracted Redis source code directory
+# Install necessary dependencies for downloading and installing Redis
+RUN microdnf install -y wget yum-utils logrotate systemd && \
+    wget https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/redis-6.2.7-1.el9.x86_64.rpm && \
+    rpm -i redis-6.2.7-1.el9.x86_64.rpm && \
+    rm -f redis-6.2.7-1.el9.x86_64.rpm && \
```
bparees, line 8: if we're deliberately switching to redis6 (and i'm not sure why we would, since rhel offers redis 7 and v7 will be in support longer), it can definitely be installed on centos using a centos base image, with just a dnf install.

which we should probably doing regardless (using a centos base image, if we're going to add centos content..i don't see a good reason not to):

https://quay.io/repository/centos/centos?tab=tags

bparees:
/hold

i want to talk about why we're switching to redis6

vishnuchalla:
> /hold
> 
> i want to talk about why we're switching to redis6

redis 7 disappeared all of a sudden from the official centos stream dowloads: https://pkgs.org/download/redis (i.e. which makes me think if its reliable). So had to switch to redis 6 which also has TLS enabled. But if we are okay with building from quay's centos, I can take a look at it.

Containerfile.redis
```diff
@@ -1,39 +1,21 @@
...
-
-# Set protected-mode to no and bind to all interfaces (0.0.0.0) in redis.conf
-RUN sed -i 's/^protected-mode yes/protected-mode no/' redis.conf
-RUN sed -i 's/^bind 127.0.0.1 -::1/bind 0.0.0.0/' redis.conf
-RUN sed -i 's/^port 6379/port 0/' redis.conf
-RUN sed -i 's/^# tls-port 6379/tls-port 6379/' redis.conf
-RUN sed -i 's/^# tls-cert-file redis.crt/tls-cert-file \/etc\/certs\/tls.crt/' redis.conf
-RUN sed -i 's/^# tls-key-file redis.key/tls-key-file \/etc\/certs\/tls.key/' redis.conf
-RUN sed -i 's/^# tls-ca-cert-file ca.crt/tls-ca-cert-file \/etc\/certs\/cm-olsredisca\/service-ca.crt/' redis.conf
-RUN sed -i 's/^# tls-auth-clients optional/tls-auth-clients optional/' redis.conf
```
bparees, line 36: you don't need any of these config file tweaks any more?

vishnuchalla: Those do not look good here, So I am passing them through cli [command](https://github.com/openshift/lightspeed-operator/pull/25/files#diff-0bba25c929831058ca39ea8a7599b3ee823c1eee0798ae2a1a1228665fd33e7dR53).

bparees:
@vishnuchalla i think this makes our redis containerfile better, but:

1) the commits say it uses rpms, but that's no longer true so please fix the commit text (i already fixed the PR title)
2) this does not actually implement the "install from an rpm" acceptance criteria of OLS-312, so i think we need to put OLS-312 back into our backlog until we resolve the issue w/ getting the centos rpm.

vishnuchalla:
> @vishnuchalla i think this makes our redis containerfile better, but:
> 
> 1. the commits say it uses rpms, but that's no longer true so please fix the commit text (i already fixed the PR title)
> 2. this does not actually implement the "install from an rpm" acceptance criteria of [OLS-312](https://issues.redhat.com//browse/OLS-312), so i think we need to put [OLS-312](https://issues.redhat.com//browse/OLS-312) back into our backlog until we resolve the issue w/ getting the centos rpm.

Done. Updated as suggested.

PR #507, opened 2024-02-29
Basic metrics endpoint e2e test

tisnik:
## Description

Basic metrics endpoint e2e test

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] e2e test

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -123,3 +123,28 @@ def test_query_filter() -> None:
     assert "deployment" in json_response["response"].lower()
     assert "foo" not in json_response["response"]
     assert "bar" not in json_response["response"]
+
+
+def test_metrics() -> None:
+    """Check if service provides metrics endpoint with expected metrics."""
+    response = client.get("/metrics/")
+    assert response.status_code == requests.codes.ok
+    assert response.text is not None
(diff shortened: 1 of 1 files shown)


tests/e2e/test_api.py
```diff
@@ -123,3 +123,28 @@ def test_query_filter() -> None:
     assert "deployment" in json_response["response"].lower()
     assert "foo" not in json_response["response"]
     assert "bar" not in json_response["response"]
+
+
+def test_metrics() -> None:
```
bparees, line 128: this is a nice start but what is really needed is a test that:

1) fetches the metrics
2) submits a query
3) fetches the metrics again
4) ensures the metric values increased as expected (checking for exact values is probably not feasible since things like token counts will vary, but seeing that they increased by some reasonable amount should work)

otherwise all this test does is confirm that we are reporting metrics w/ these names, but not that the values actually reflect the what they are supposed to be measuring.

tisnik: yup, exactly this is on my TODO list ;)

PR #508, opened 2024-02-29
Possible fix: RAG installation

tisnik:
## Description

Possible fix: RAG installation

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


bparees:
https://github.com/openshift/lightspeed-service/pull/506 fixes this.  Also we shouldn't need to install python again because it's extending the base ols image which already installs it: https://github.com/openshift/lightspeed-service/blob/e096673a2f18b56f9edbfd117ce1490f699629ea/Containerfile#L6

PR #512, opened 2024-03-01
Use `object` type instead of `Any`

tisnik:
## Description

The Python documentation recommends the use of `object` to "indicate that a
value could be any type in a typesafe manner", while `Any` should be used to
"indicate that a value is dynamically typed."

The semantics of `__eq__` and `__ne__` are such that the `obj` parameter
should be any type, as opposed to a dynamically typed value. Therefore, the
`object` type annotation is more appropriate.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -3,7 +3,7 @@
 import logging
 import os
 import re
-from typing import Any, Optional
+from typing import Optional
 from urllib.parse import urlparse
 
 from pydantic import BaseModel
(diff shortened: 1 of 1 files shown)


syedriko:
@tisnik Some are going even further:
-  https://stackoverflow.com/questions/37557411/why-does-defining-the-argument-types-for-eq-throw-a-mypy-type-error
- https://jcalderone.livejournal.com/32837.html

PR #513, opened 2024-03-01
OLS-345: configuration in runner.py

onmete:
## Description

Envs setting moved to runner.py

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -1,12 +1,34 @@
-"""Script that is able to start Uvicorn-based REST API service."""
+"""Main app entrypoint. Starts Uvicorn-based REST API service."""
 
 import os
 
 import uvicorn
 
+
+def configure_hugging_face_envs() -> None:
+    """Configure HuggingFace library environment variables."""
+    # NOTE: We import config here to avoid triggering import of anything
+    # else via our code before other envs are set (mainly the gradio).
+    from ols.utils import config
+
+    cfg_file = os.environ.get("OLS_CONFIG_FILE", "olsconfig.yaml")
+    config.init_config(cfg_file)
+    if (
+        config.ols_config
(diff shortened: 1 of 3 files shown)


runner.py
```diff
@@ -1,12 +1,39 @@
...
+if __name__ == "__main__":
+
+    cfg_file = os.environ.get("OLS_CONFIG_FILE", "olsconfig.yaml")
+    config.init_config(cfg_file)
+
+    configure_logging(config.ols_config.logging_config)
+    logger = logging.getLogger(__name__)
+    logger.info(f"Config loaded from {Path(cfg_file).resolve()}")
+
+    configure_3rd_party_envs()
```
syedriko, line 34: It is safer to set these as soon as possible, before we do anything else. For example, Gradio looks at GRADIO_ANALYTICS_ENABLED at package initialization time:
https://github.com/gradio-app/gradio/blob/24626ea362a35606d344f341a308556c52b34cc0/gradio/strings.py#L39

tisnik (commented):
it looks good

runner.py
```diff
@@ -1,12 +1,34 @@
...
+        )
+        os.environ["TRANSFORMERS_OFFLINE"] = "1"
+
+
 if __name__ == "__main__":
-    # TODO syedriko consolidate this env var into a central location as per OLS-345.
     # disable Gradio analytics, which calls home to https://api.gradio.app
     os.environ["GRADIO_ANALYTICS_ENABLED"] = "false"
 
+    configure_hugging_face_envs()
```
bparees, line 32: i would expect config initialization to be called from here, not from within configure_hugging_face_envs.  config initialization shouldn't be a side effect of setting the hugging face env vars.

bparees: @onmete ^^^^

PR #522, opened 2024-03-04
OLS-378: Check for proper user UUID

tisnik:
## Description

Now we know that user ID is proper UUID. It means that compound key to Redis can be checked more thorougly.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-378](https://issues.redhat.com//browse/OLS-378)

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -40,7 +40,7 @@ def conversation_request(
     referenced_documents: list[str] = []
 
     # TODO: retrieve proper user ID from request
-    user_id = "user1"
+    user_id = constants.DEFAULT_USER_UID
 
     conversation_id = retrieve_conversation_id(llm_request)
     previous_input = retrieve_previous_input(user_id, llm_request)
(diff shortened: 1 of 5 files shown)


ols/app/endpoints/ols.py
```diff
@@ -40,7 +40,7 @@ def conversation_request(
     referenced_documents: list[str] = []
 
     # TODO: retrieve proper user ID from request
-    user_id = "user1"
+    user_id = constants.DEFAULT_USER_UID
```
bparees, line 43: it is odd to call this a "default" user id.  There is no default user id for OLS.  this is a sample value, at best.  I would not expect any OLS code outside of test logic, to make use of this value.  After all, there is no such thing as a valid "default" user.

when we get into production with auth enabled and have requests coming in w/ no user id associated, we do not want to fallback to some default value(which will then fail, but in a less obvious way).  We want to fail because the user id is a required parameter.

So if the purpose of this logic is to allow OLS to not fail when we don't have user id information being passed in, we need to think a little more about how we want to handle that during dev/testing, without impacting how OLS behaves in production.

tisnik: yup I know, there's OLS-378 that are pretty easy to get done (talked about it with Nimer couple hours ago).
So it will go away (no default user at all).

PR #524, opened 2024-03-04
OLS-241 , OLS-297- enable TLS

nemerna:
## Description

1) added support for TLS in Uvicorn and app config.
2) moved the config initialization to the runner.py
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


runner.py
```diff
@@ -1,12 +1,61 @@
...
+        if (
+            config.ols_config.tls_config.ssl_certfile
+            and config.ols_config.tls_config.ssl_keyfile
+        ):
+            uvicorn.run(
+                "ols.app.main:app",
+                host=host,
+                port=8443,
+                workers=1,
+                **config.ols_config.tls_config.__dict__,
```
tisnik, line 43: ah nice trick!

runner.py
```diff
@@ -1,12 +1,61 @@
...
+
+    config.init_query_filter()
+    config.init_vector_index()
+
+    # ruff: noqa: S104
+    host = "127.0.0.1" if config.dev_config.run_on_localhost else "0.0.0.0"
+
+    if config.dev_config.tls_disabled:
+        # TLS is disabled, run without SSL configuration
+        uvicorn.run("ols.app.main:app", host=host, port=8080, workers=1)
```
tisnik, line 32: I'd probably add a comment why `workers=1` does the trick

scripts/generate-certs.sh
```diff
@@ -0,0 +1,17 @@
...
+CERT_DIR="./certs"
+PRIVATE_KEY="$CERT_DIR/private.key"
+CERTIFICATE="$CERT_DIR/certificate.crt"
+DAYS_VALID=365
+
+# Create directory for certificates if it doesn't exist
+mkdir -p "$CERT_DIR"
+
+# Generate a private key and self-signed certificate
+openssl req -x509 -newkey rsa:4096 -sha256 -days "$DAYS_VALID" -nodes \
```
tisnik, line 13: dunno if we need to install `openssl` for e2e-tests or if it's in the base image already

tests/config/singleprovider.e2e.template.config.yaml
```diff
@@ -27,4 +27,5 @@ ols_config:
 dev_config:
   llm_params:
     temperature: 0
-  disable_auth: true
\ No newline at end of file
+  disable_auth: true
+  tls_disabled: true
```
tisnik, line 31: just nitpick (or hard read for me): there are two "disable-X" configuration options, each one spelled differently ("disable-a" and "b-disabled")

runner.py
```diff
@@ -1,12 +1,61 @@
...
+            uvicorn.run(
+                "ols.app.main:app",
+                host=host,
+                port=8443,
+                workers=1,
+                **config.ols_config.tls_config.__dict__,
+            )
+        else:
+            logger.error("TLS is enabled, But no certificate/key was provided")
+    # an alternative from directly using the config __dict__
```
tisnik, line 47: please remove commented code before ack

README.md
```diff
@@ -103,7 +103,54 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
...
+
+
+   1. Enabling and Disabling TLS
+   
+      By default, TLS is enabled in OLS. To disable TLS, adjust the `dev_config` in your configuration file as shown below:
+
+      ```yaml
+         dev_config:
+            tls_disabled: false
+      ```
```
syedriko, line 118: `To disable TLS` it needs to be
```
    tls_disabled: true
```

README.md
```diff
@@ -103,7 +103,54 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
...
+
+   2. Configuring TLS in local Environmenets:
```
syedriko, line 120: environments (spelling)

README.md
```diff
@@ -103,7 +103,54 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
...
+
+      1. Generate Self-Signed Certificates: To generate self-signed certificates, run the following command from the project's root directory:
+         ```bash
+            ./scripts/generate-certs.sh
+         ``` 
+      2. Update OLS Configuration: Modify your config.yaml to include the paths to your certificates:
```
syedriko, line 126: there's only one certificate.  "paths to your certificate and its private key"?

README.md
```diff
@@ -103,7 +103,54 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
...
+         ```yaml
+            ols_config:
+               tls_config:
+                  ssl_certfile: /full/path/to/certs/cert.pem
+                  ssl_keyfile: /full/path/to/certs/key.pem
+         ```
+      3. Launch OLS with HTTPS: After applying the above configurations, OLS will run over HTTPS.
+   
+   3. Configuring OLS in Opnehsift:
```
syedriko, line 135: OpenShift (spelling)

ols/app/models/config.py
```diff
@@ -109,6 +109,56 @@ def validate_yaml(self) -> None:
...
+
+    ssl_certfile: Optional[str] = None
+    ssl_keyfile: Optional[str] = None
+    ssl_keyfile_password: Optional[str] = None
+
+    def __init__(self, data: Optional[dict] = None) -> None:
+        """Initialize configuration and perform basic validation."""
+        super().__init__()
+        if data is not None:
+            self.ssl_certfile = data.get("ssl_certfile", self.ssl_certfile)
```
syedriko, line 123: This is IMO confusing, why not just say None for the default value?

README.md
syedriko: Should we call it "disable_tls" to mirror "disable_auth"?

ols/app/models/config.py
```diff
@@ -109,6 +109,56 @@ def validate_yaml(self) -> None:
...
+            self.ssl_keyfile = data.get("ssl_keyfile", self.ssl_keyfile)
+            self.ssl_keyfile_password = data.get(
+                "ssl_keyfile_password", self.ssl_keyfile_password
+            )
+
+    def validate_yaml(self) -> None:
```
syedriko, line 129: There're 8 combinations to consider. It doesn't make sense to not have a cert file, so we need to check the remaining 4.

```
def _file_check(path: str, desc: str) -> None:
    """Check that path is a readable regular file."""
    if not os.path.exists(path):
        raise InvalidConfigurationError(f"{desc} '{path}' does not exist")
    if not os.path.isfile(path):
        raise InvalidConfigurationError(f"{desc} '{path}' is not a directory")
    if not os.access(path, os.R_OK):
        raise InvalidConfigurationError(f"{desc} '{path}' is not readable")


def validate_yaml(self) -> None:
    """Validate TLS config."""
    if not self.ssl_certfile:
        raise InvalidConfigurationError(
            "Can not enable TLS without ols_config.tls_config.ssl_certfile"
        )
    else:
        _file_check(self.ssl_certfile, "OLS server certificate")
        if self.ssl_keyfile:
            _file_check(self.ssl_keyfile, "OLS server certificate private key")
            if self.ssl_keyfile_password:
                _file_check(self.ssl_keyfile_password, "OLS server certificate private key password")
        else:
            if self.ssl_keyfile_password:
                raise InvalidConfigurationError(
                    "Can not have ols_config.tls_config.ssl_certfile_password without ols_config.tls_config.ssl_keyfile"
                )
```

ols/app/models/config.py
```diff
@@ -581,6 +634,8 @@ def validate_yaml(self) -> None:
             self.reference_content.validate_yaml()
         if self.authentication_config:
             self.authentication_config.validate_yaml()
+        if self.tls_config:
+            self.tls_config.validate_yaml()
```
syedriko, line 638: Should we be validating TLS config if it is disabled?

runner.py
```diff
@@ -1,12 +1,61 @@
...
+
+    configure_logging(config.ols_config.logging_config)
+    logger = logging.getLogger(__name__)
+    logger.info(f"Config loaded from {Path(cfg_file).resolve()}")
+
+    config.init_query_filter()
+    config.init_vector_index()
+
+    # ruff: noqa: S104
+    host = "127.0.0.1" if config.dev_config.run_on_localhost else "0.0.0.0"
```
syedriko, line 28: This is IPv4-only. According to https://github.com/encode/uvicorn/discussions/1529
```
host = "localhost" if config.dev_config.run_on_localhost else None
```
should cover all three combination of IP stacks, but this needs to be tested.

scripts/generate-certs.sh
```diff
@@ -0,0 +1,17 @@
...
+  -keyout "$PRIVATE_KEY" -out "$CERTIFICATE" -subj "/CN=localhost" \
+  -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
```
syedriko, line 15: Please add ",IP:::1"

runner.py
nemerna: yep thats why i commented the Alternative

scripts/generate-certs.sh
nemerna: i expect it exists, if no we will need a separate task to include it or pr-generate certs ?

ols/app/models/config.py
nemerna: simply we disable TLS using dev_config, for that we cant know if dev_config.disable_tls is tru or false while initialiing/validating the model

ols/app/models/config.py
nemerna: we need only to check that if we provide a cert/key that they exist because they get used as a pth in the runtime

for passwod, i have modified it to be read from a file, instead of directly seting the pass in the config.yaml

i followed same logic as we do with the OPENAI_KEY config

ols/app/models/config.py
nemerna: set to if data

README.md
nemerna: Resolved

README.md
nemerna: resolved

README.md
nemerna: resolved

runner.py
nemerna: resolved

tests/config/singleprovider.e2e.template.config.yaml
nemerna: resolved

runner.py
nemerna: resolved

runner.py
nemerna: it works! Resolved

scripts/generate-certs.sh
nemerna: Resolved

scripts/generate-certs.sh
```diff
@@ -0,0 +1,27 @@
...
+      -keyout "$PRIVATE_KEY" -out "$CERTIFICATE" -subj "/CN=localhost" \
+      -addext "subjectAltName=DNS:localhost,IP:127.0.0.1,IP:::1"
+else
+    # Password provided, generate an encrypted private key and self-signed certificate
+    openssl req -x509 -newkey rsa:4096 -sha256 -days "$DAYS_VALID" \
+      -keyout "$PRIVATE_KEY" -out "$CERTIFICATE" -subj "/CN=localhost" \
+      -passout pass:"$PASSWORD" \
+      -addext "subjectAltName=DNS:localhost,IP:127.0.0.1,IP:::1"
+fi
+
```
syedriko, line 26: This is asking for deduplication:
```
if [ -z "$PASSWORD" ]; then
    PASSWORD_PARAM=(-noenc)
else
    PASSWORD_PARAM=(-passout "pass:$PASSWORD")
fi

openssl req -x509 -newkey rsa:4096 -sha256 -days "$DAYS_VALID" "${PASSWORD_PARAM[@]}" \
    -keyout "$PRIVATE_KEY" -out "$CERTIFICATE" -subj "/CN=localhost" \
    -addext "subjectAltName=DNS:localhost,IP:127.0.0.1,IP:::1"
```
it handles no password as well as passwords with and without spaces.

ols/app/models/config.py
syedriko: I meant
```
self.ssl_certfile = data.get("ssl_certfile", None)
```

ols/app/models/config.py
syedriko: There's no question about that. My comment was more about checking for valid combinations of ssl_certfile, ssl_keyfile and ssl_keyfile_password.

ols/app/models/config.py
syedriko: We can still do it by passing a disable_tls flag into OLSConfig.validate_yaml(). What's the UX we're looking for here? I'd imagine if I set disable_tls: true I won't need to generate cert/key.

nemerna: this will require changing the entire way we are validating config in the ols.utils.config
i thought about it multiple times

the only thing i reached to is that the OLSConfig should be aware of enablement/disablement of tls
and that can be done if we move the enable/disable entry to the ols_config instead of dev_config

tests/config/cluster_install/ols_configmap.yaml
```diff
@@ -41,3 +41,4 @@ data:
     dev_config:
       enable_dev_ui: true
       disable_auth: false
+      disable_tls: true
```
bparees, line 44: why disable tls?  we are already setting it up today:
https://github.com/openshift/lightspeed-service/blob/f7f3f29415053ad696d77c7bb8bbeadb8554e7f4/tests/config/cluster_install/ols_manifests.yaml#L49-L52

so you just need to add the paths to the tls cert/key files to the config.  The same ones we were passing as args previously.

ols/app/models/config.py
bparees: the proper check for the cert/key value is that if one value is provided, the other value must also be provided.

we have similar validation checks for other coupled fields.

(either both are None, or both have a valid filepath specified, or it's an error)

and yes it's also an error if they don't provide TLS key/cert values, but TLS is not disabled.  But i see that check is already being done during startup, which is fine.

ols/app/models/config.py
bparees: what you've done (checking at startup) is fine.  In general the goal is to fail as early as possible.  We want to fail before we get a request from a client, which you have achieved.

the key goal is that:
1) we are validating the configuration is complete/correct
2) we do it before the server starts accepting requests

tests/config/cluster_install/ols_configmap.yaml
bparees: also you need to fix the deployment in ols_manifests.yaml since it's not invoking runner.py today and it's passing tls values as arguments.

README.md
```diff
@@ -103,7 +103,54 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
...
+   
+   3. Configuring OLS in OpenShift:
+
+      For deploying in OpenShift, Service-Served Certificates can be utilized. Update your ols-config.yaml as shown below, based on the example provided in the examples directory:
+
+      ```yaml
+         ols_config:
+            tls_config:
+               ssl_certfile: /app-root/certs/cert.pem
+               ssl_keyfile: /app-root/certs/key.pem
```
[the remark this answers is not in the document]
nemerna: yes, but this is the term being used by uvicorn
we can replace it with tls, but this will require us to change the usage of kwargs (i used same name of the parameters names being used by uvicorn)


            uvicorn.run(
                "ols.app.main:app",
                host=host,
                port=8443,
                workers=1,
                **config.ols_config.tls_config.__dict__,
            )

README.md
```diff
@@ -103,7 +103,54 @@ OpenShift LightSpeed (OLS) is an AI powered assistant that runs on OpenShift and
...
+      ```
+   4. Using a Private Key with a Password
+      If your private key is encrypted with a password, specify the password in your ols-config.yaml as follows:
+      ```yaml
+         ols_config:
+            tls_config:
+               ssl_keyfile_password: yourpassword
```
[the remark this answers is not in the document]
raptorsun: +1 for the `_path` at the end of the property name 👍

ols/app/models/config.py
```diff
@@ -39,6 +39,16 @@ def _dir_check(path: str, desc: str) -> None:
         raise InvalidConfigurationError(f"{desc} '{path}' is not readable")
 
 
+def _file_check(path: str, desc: str) -> None:
+    """Check that path is a readable regular file."""
+    if not os.path.exists(path):
```
raptorsun, line 44: I think this check is unneccessary for checking whether the path points at a readable file.

README.md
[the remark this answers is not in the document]
bparees: uvicorn is an implementation detail, so our config should not be tied to how uvicorn expects config arguments.

ols/app/models/config.py
nemerna: Resolved

README.md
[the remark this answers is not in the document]
nemerna: Resolved

README.md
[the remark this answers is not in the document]
nemerna: Resolved

scripts/generate-certs.sh
nemerna: Resolved

nemerna:
@bparees im unable to recognize why tests are failing 
any thoughts?

runner.py
```diff
@@ -34,15 +39,38 @@ def configure_gradio_ui_envs() -> None:
...
-    uvicorn.run("ols.app.main:app", host="127.0.0.1", port=8080, reload=True)
+    # use workers=1 so config loaded can be accessed from other modules
+    if config.dev_config.disable_tls:
+        # TLS is disabled, run without SSL configuration
+        uvicorn.run("ols.app.main:app", host=host, port=8080, workers=1)
+    else:
+        uvicorn.run(
+            "ols.app.main:app",
+            host=host,
+            port=8443,
```
bparees, line 71: @xrajesh @raptorsun making sure you're aware of this....  with this PR, OLS is going to start listening on 8443 by default, unless tls is explicilty disabled in the config, so the OLS service needs to change accordingly.

@nemerna i think the best thing to do is first submit a PR to the ols operator that sets tls disabled=true.  Then we can merge this PR.  Then we can update the operator to setup the TLS/8443 service, certs, and set tls disabled=false.

raptorsun: When this change is made, we need update[ the container port  here ](https://github.com/openshift/lightspeed-operator/blob/main/internal/controller/constants.go#L37). 
The service port below is for exposing the endpoint to other applications such as console.

bparees: for now https://github.com/openshift/lightspeed-operator/pull/37 has the operator define the ols config w/ TLS disabled.

After this PR merges, the operator can be updated to properly configure TLS.

PR #527, opened 2024-03-04
OLS-364: Fixed all return type hints, so thich check will be possible to enabl…

tisnik:
## Description

[OLS-364](https://issues.redhat.com//browse/OLS-364): Fixed all return type hints, so thich check will be possible to enabl…

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Type hints

- Related Issue #[OLS-364](https://issues.redhat.com//browse/OLS-364)

### DIFF
--- a/ols/src/llms/llm_loader.py
+++ b/ols/src/llms/llm_loader.py
@@ -49,7 +49,7 @@ def _resolve_provider_config(
             f"Valid models are: {list(provider_config.models.keys())}"
         )
 
-    return provider_config
+    return provider_config  # type: ignore [return-value]
 
 
 def load_llm(provider: str, model: str, llm_params: Optional[dict] = None) -> LLM:
--- a/ols/src/llms/providers/openai.py
+++ b/ols/src/llms/providers/openai.py
@@ -47,4 +47,4 @@ class OpenAI(LLMProvider):
         if "max_new_tokens" in self.params:
             self.params.pop("max_new_tokens")
 
-        return ChatOpenAI(**self.params)
+        return ChatOpenAI(**self.params)  # type: ignore [return-value]
--- a/ols/src/llms/providers/registry.py
+++ b/ols/src/llms/providers/registry.py
@@ -35,7 +35,7 @@ def register_llm_provider_as(provider_type: str) -> Callable:
     ```
     """
 
-    def decorator(cls: LLMProvider) -> Callable:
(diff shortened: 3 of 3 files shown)


ols/src/llms/llm_loader.py
```diff
@@ -49,7 +49,7 @@ def _resolve_provider_config(
             f"Valid models are: {list(provider_config.models.keys())}"
         )
 
-    return provider_config
+    return provider_config  # type: ignore [return-value]
```
bparees, line 52: why do we need to ignore the type?  is ` providers_config.providers.get(provider)` not setting a return type?

basically same question on all of these.  Why do we have to ignore the return type? doesn't that imply we're not setting types correctly somewhere else?

bparees: providers is:
```
providers: dict[str, ProviderConfig] = {}
```

so why would providers.get() not be expected to return an object of type ProviderConfig? (or perhaps `ProviderConfig | None`)?

tisnik: `ProviderConfig | None` is the problem there, because `get` might return `None`. So either the get will be replaced by direct selector or we ignore return value type.

bparees: or we change this:
```
def _resolve_provider_config(
    provider: str, model: str, providers_config: LLMProviders
) -> ProviderConfig:
```

to
```
def _resolve_provider_config(
    provider: str, model: str, providers_config: LLMProviders
) -> ProviderConfig|None:
```

(or whatever the syntax is), right?

(even though we never really return none because we throw an exception if providerconfig is none)

tisnik: yes, but transitionally it will "force" us to check for None in every piece of code that use provider.

bparees: got it

bparees: i am increasingly getting the impression that exhaustive type checking is going to be:

1) difficult to implement
2) difficult to maintain
3) have significant limitations such as this one that will compromise the value it can provide or create confusion about when/why it's ok to ignore types

bparees: fair enough

PR #538, opened 2024-03-05
OLS-135: feedback storage

onmete:
## Description

Implementation of the feedback storage system.

## Type of change

- [x] New feature

## Related issues

[OLS-397](https://issues.redhat.com/browse/OLS-397)
[OLS-387](https://issues.redhat.com/browse/OLS-387)

## TODO

- [ ] - https://github.com/openshift/lightspeed-service/pull/595

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -90,14 +90,36 @@
                 }
             }
         },
+        "/v1/feedback/list": {
+            "get": {
+                "tags": [
+                    "feedback"
+                ],
+                "summary": "Get User Feedbacks",
+                "description": "Handle feedback listing requests.\n\nArgs:\n    auth: The Authentication handler (FastAPI Depends) that will handle authentication Logic.\n\nReturns:\n    Response containing the list of feedbacks.",
+                "operationId": "get_user_feedbacks_v1_feedback_list_get",
+                "responses": {
+                    "200": {
+                        "description": "Successful Response",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/FeedbacksListResponse"
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        },
         "/v1/feedback": {
             "post": {
                 "tags": [
                     "feedback"
                 ],
-                "summary": "User Feedback",
-                "description": "Handle feedback requests.\n\nArgs:\n    feedback_request: The request containing feedback information.\n    auth: The Authentication handler (FastAPI Depends) that will handle authentication Logic.\n\nReturns:\n    Response indicating the status of the feedback.",
-                "operationId": "user_feedback_v1_feedback_post",
+                "summary": "Store User Feedback",
+                "description": "Handle feedback requests.\n\nArgs:\n    feedback_request: The request containing feedback information.\n    auth: The Authentication handler (FastAPI Depends) that will handle authentication Logic.\n\nReturns:\n    Response indicating the status of the feedback storage request.",
+                "operationId": "store_user_feedback_v1_feedback_post",
                 "requestBody": {
                     "content": {
                         "application/json": {
@@ -108,6 +130,49 @@
                     },
                     "required": true
                 },
+                "responses": {
+                    "200": {
+                        "description": "Successful Response",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/FeedbackResponse"
+                                }
+                            }
+                        }
+                    },
+                    "422": {
+                        "description": "Validation Error",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/HTTPValidationError"
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        "/v1/feedback/{feedback_id}": {
+            "delete": {
+                "tags": [
+                    "feedback"
+                ],
+                "summary": "Remove User Feedback",
+                "description": "Handle feedback removal requests.\n\nArgs:\n    feedback_id: The feedback ID (UUID) to be removed.\n    auth: The Authentication handler (FastAPI Depends) that will handle authentication Logic.\n\nReturns:\n    Response indicating the status of the feedback removal.",
+                "operationId": "remove_user_feedback_v1_feedback__feedback_id__delete",
+                "parameters": [
+                    {
+                        "name": "feedback_id",
+                        "in": "path",
+                        "required": true,
+                        "schema": {
+                            "type": "string",
+                            "title": "Feedback Id"
+                        }
+                    }
+                ],
                 "responses": {
                     "200": {
                         "description": "Successful Response",
@@ -288,18 +353,45 @@
                         "type": "string",
                         "title": "Conversation Id"
                     },
-                    "feedback_object": {
-                        "type": "object",
-                        "title": "Feedback Object"
+                    "user_question": {
+                        "type": "string",
+                        "title": "User Question"
+                    },
+                    "llm_response": {
+                        "type": "string",
+                        "title": "Llm Response"
+                    },
+                    "sentiment": {
+                        "anyOf": [
+                            {
+                                "type": "integer"
+                            },
+                            {
+                                "type": "null"
+                            }
+                        ],
+                        "title": "Sentiment"
+                    },
+                    "user_feedback": {
+                        "anyOf": [
+                            {
+                                "type": "string"
+                            },
+                            {
+                                "type": "null"
+                            }
+                        ],
+                        "title": "User Feedback"
                     }
                 },
                 "type": "object",
                 "required": [
                     "conversation_id",
-                    "feedback_object"
+                    "user_question",
+                    "llm_response"
                 ],
                 "title": "FeedbackRequest",
-                "description": "Model representing a feedback request.\n\nAttributes:\n    conversation_id: The required conversation ID (UUID).\n    feedback_object: The JSON blob representing feedback.\n\nExample:\n    ```python\n    feedback_request = FeedbackRequest(\n        conversation_id=\"12345678-abcd-0000-0123-456789abcdef\",\n        feedback_object={\"rating\": 5, \"comment\": \"Great service!\"}\n    )\n    ```",
+                "description": "Model representing a feedback request.\n\nAttributes:\n    conversation_id: The required conversation ID (UUID).\n    user_question: The required user question.\n    llm_response: The required LLM response.\n    sentiment: The optional sentiment.\n    user_feedback: The optional user feedback.\n\nExample:\n    ```python\n    feedback_request = FeedbackRequest(\n        conversation_id=\"12345678-abcd-0000-0123-456789abcdef\",\n        user_question=\"what are you doing?\",\n        llm_response=\"I don't know\",\n        sentiment=-1,\n    )\n    ```",
                 "examples": [
                     {
                         "conversation_id": "12345678-abcd-0000-0123-456789abcdef",
@@ -326,6 +418,30 @@
                 "examples": [
                     {
                         "response": "feedback received"
+                    }
+                ]
+            },
+            "FeedbacksListResponse": {
+                "properties": {
+                    "feedbacks": {
+                        "items": {
+                            "type": "string"
+                        },
+                        "type": "array",
+                        "title": "Feedbacks"
+                    }
+                },
+                "type": "object",
+                "required": [
+                    "feedbacks"
+                ],
+                "title": "FeedbacksListResponse",
+                "description": "Model representing a response to a feedback list request.\n\nAttributes:\n    feedbacks: The list of feedback IDs.\n\nExample:\n    ```python\n    feedbacks_list_response = FeedbacksListResponse(\n        feedbacks=[\"12345678-abcd-0000-0123-456789abcdef\"]\n    )\n    ```",
+                "examples": [
+                    {
+                        "feedbacks": [
+                            "12345678-abcd-0000-0123-456789abcdef"
+                        ]
                     }
                 ]
             },
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -118,7 +118,7 @@
                     "feedback"
                 ],
                 "summary": "Get User Feedbacks",
-                "description": "Handle feedback listing requests.\n\nArgs:\n    auth: The Authentication handler (FastAPI Depends) that will handle authentication Logic.\n\nReturns:\n    Response containing the list of feedbacks.",
+                "description": "Handle feedback listing requests.\n\nArgs:\n    feedback_is_enabled: The feedback handler (FastAPI Depends) that\n        will handle feedback status checks.\n    auth: The Authentication handler (FastAPI Depends) that will\n        handle authentication Logic.\n\nReturns:\n    Response containing the list of feedbacks.",
                 "operationId": "get_user_feedbacks_v1_feedback_list_get",
                 "responses": {
                     "200": {
@@ -140,7 +140,7 @@
                     "feedback"
                 ],
                 "summary": "Store User Feedback",
-                "description": "Handle feedback requests.\n\nArgs:\n    feedback_request: The request containing feedback information.\n    auth: The Authentication handler (FastAPI Depends) that will handle authentication Logic.\n\nReturns:\n    Response indicating the status of the feedback storage request.",
+                "description": "Handle feedback requests.\n\nArgs:\n    feedback_request: The request containing feedback information.\n    feedback_is_enabled: The feedback handler (FastAPI Depends) that\n        will handle feedback status checks.\n    auth: The Authentication handler (FastAPI Depends) that will\n        handle authentication Logic.\n\nReturns:\n    Response indicating the status of the feedback storage request.",
                 "operationId": "store_user_feedback_v1_feedback_post",
                 "requestBody": {
                     "content": {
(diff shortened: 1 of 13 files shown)


ols/app/endpoints/feedback.py
```diff
@@ -1,19 +1,58 @@
...
+    """
+    # ensures storage path exists
+    storage_path = Path(config.ols_config.feedback_storage_location)
+    if not storage_path.exists():
+        storage_path.mkdir(parents=True)
+
+    # stores feedback in a file named after the conversation id
+    feedback_file_path = storage_path / f"{feedback_request.conversation_id}.json"
+    with open(feedback_file_path, "w", encoding="utf-8") as feedback_file:
+        json.dump(feedback_request.feedback_object, feedback_file)
```
onmete, line 32: Do we need to store the conversation_id in the json too?

onmete (commented):
@tisnik please review

tisnik (commented):
I'd add support for list of feedback files or something similar.

ols/app/endpoints/feedback.py
```diff
@@ -1,19 +1,58 @@
...
+    Args:
+        feedback_request: The request containing feedback information.
+    """
+    # ensures storage path exists
+    storage_path = Path(config.ols_config.feedback_storage_location)
+    if not storage_path.exists():
+        storage_path.mkdir(parents=True)
+
+    # stores feedback in a file named after the conversation id
+    feedback_file_path = storage_path / f"{feedback_request.conversation_id}.json"
```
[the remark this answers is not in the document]
tisnik: it might be part of the JSON itself, but can be added into the filename too if needed.

ols/app/endpoints/feedback.py
tisnik: I guess so, as the filename is not strictly part in the context.

ols/app/endpoints/feedback.py
```diff
@@ -1,19 +1,91 @@
...
+def remove_feedback(feedback_request: FeedbackRequest) -> None:
+    """Remove feedback from the local filesystem.
+
+    Args:
+        feedback_request: The request containing feedback information.
+    """
+    storage_path = Path(config.ols_config.feedback_storage_location)
+    feedback_file = storage_path / f"{feedback_request.conversation_id}.json"
+    if feedback_file.exists():
+        feedback_file.unlink()
```
tisnik, line 62: some check there perhaps?

tisnik (commented):
some comments/nitpicks

tests/integration/test_feedback.py
```diff
@@ -69,8 +75,44 @@ def test_feedback_wrong_conversation_id():
...
+    # store some feedback first
+    conversation_id = suid.get_suid()
+    response = client.post(
+        "/v1/feedback",
+        json={"conversation_id": conversation_id, "feedback_object": {"some": "data"}},
+    )
+    assert response.status_code == requests.codes.ok
+
+    # The TestClient.delete() method from Starlette's TestClient does
+    # not accept a json argument - workaround it by using `requests`
```
onmete, line 112: This suggests defining the endpoint in this manner `/v1/feedback/{feedback_id}`. As it is unsure that the endpoint will stay in place, I'm not sure if it makes sense to further invest in it. WDYT?

tests/integration/test_feedback.py
```diff
@@ -69,8 +75,44 @@ def test_feedback_wrong_conversation_id():
...
+    response = client.request(
+        "DELETE", "/v1/feedback/", json={"feedback_id": conversation_id}
+    )
+
+    assert response.status_code == requests.codes.ok
+    assert response.json() == {"response": "feedback removed"}
```
onmete, line 118: This comment is not related to the line, I just putting it here to have the "threading" ability on the comment.

I didn't implement the tests like `test_feedback_no_payload_send`. Those are basically testing the FastAPI/pydantic behavior. Unless there is a special logic for handling/validating the input parameters, I'm not sure what is the actual benefit of such tests. What we can do if it starts failing?

Up to discussion.

tests/unit/app/endpoints/test_feedback.py
```diff
@@ -1,17 +1,91 @@
...
+    """Fixture sets feedback location to tmpdir and return the path."""
+    config.init_empty_config()
+    config.ols_config.feedback_storage_location = tmpdir.strpath
+    return tmpdir.strpath
+
+
+def store_fake_feedback(path, filename, data):
+    """Store feedback data."""
+    with open(f"{path}/{filename}.json", "w") as f:
+        f.write(json.dumps(data))
```
tisnik, line 23: some cleanup mechanism?

tests/unit/app/endpoints/test_feedback.py
```diff
@@ -1,17 +1,91 @@
...
+    feedback.store_feedback(conversation_id, user_id, feedback_data)
+
+    stored_data = load_fake_feedback(conversation_id)
+    assert stored_data == {
+        "metadata": {
+            "conversation_id": "12345678-abcd-0000-0123-456789abcdef",
+            "user_id": "12345678-abcd-0000-0123-456789abcdef",
+        },
+        "feedback": {"something": "interesting"},
+    }
```
tisnik, line 58: cleanup

tisnik (commented):
it looks ok in overall. Just if tests fails for any reason, they will leave files in tempdir. That might affect new test run (and tests needs to be idempotent anyway). I'd suggest to use cleanup mechanism

tests/unit/app/endpoints/test_feedback.py
onmete: Are you sure about that? Docs about `tmpdir` fixture says

> Return a temporary directory path object which is unique to each test function invocation...

tisnik: yup, but don't the files remain on FS?

tisnik: if not, I'll LGTM ASAP

onmete: I believe so.

I've created another test (before the one that lists the dir) that I raised after storing and it seems to be passing.

tisnik: then it looks good, thank you!

onmete:
@bparees ready for your review :)

ols/app/endpoints/feedback.py
```diff
@@ -1,19 +1,111 @@
...
+    data_to_store = {
+        "metadata": {
+            "conversation_id": conversation_id,
+            "user_id": user_id,
+        },
+        "feedback": feedback,
+    }
+
+    # stores feedback in a file named after the conversation id
+    feedback_file_path = storage_path / f"{conversation_id}.json"
```
bparees, line 58: this isn't sufficiently unique to support cases where a user provides feedback on multiple interactions within a single conversation.  

or cases where someone submits feedback multiple times for the same interaction (less likely, but no particular reason we shouldn't allow it).

so i think this requires another identifier, just a uuid generated for each feedback submission.

ols/app/endpoints/feedback.py
```diff
@@ -1,19 +1,111 @@
...
+        else:
+            logger.error(f"feedback file '{feedback_file}' failed to remove")
+    else:
+        logger.error(f"feedback file '{feedback_file}' not found")
+
+
+# TODO: OLS-136 implements the collection mechanism - revisit the need
+# of this endpoint
+@router.get("/list")
+def get_user_feedbacks(auth: Any = Depends(auth_dependency)) -> FeedbackResponse:
```
bparees, line 90: yeah i would prefer we not have this until we decide we need it(if ever).

also if we did want it, the auth protection needs to be different from the auth requirement for sending queries to OLS.  We wouldn't want a regular OLS user to be able to retrieve all the feedback objects.  So that is something to keep in mind as you're adding new endpoints, that different endpoints have different permission requirements.

ols/app/endpoints/feedback.py
```diff
@@ -27,7 +119,33 @@ def user_feedback(
...
 
     return FeedbackResponse(response="feedback received")
+
+
+# TODO: OLS-136 implements the collection mechanism - revisit the need
+# of this endpoint
+@router.delete("")
+def remove_user_feedback(
+    feedback_request: RemoveFeedbackRequest, auth: Any = Depends(auth_dependency)
+) -> FeedbackResponse:
```
bparees, line 138: same comments as the list endpoint.  let's remove this until we need it, and if we need it, the auth permissions need to be different.

ols/app/models/models.py
```diff
@@ -103,7 +103,7 @@ class FeedbackRequest(BaseModel):
     """
 
     conversation_id: str
-    feedback_object: dict  # TODO: proper object (OLS-77)
+    feedback_object: dict  # TODO: proper object will be defined by OLS-138
```
bparees, line 106: [OLS-138](https://issues.redhat.com/browse/OLS-138) is concluded, the information we want to collect:
* sentiment(thumbs up/down)
* a freeform text string
* the user query/question that was submitted
* the llm response to the question
* the conversation id

@thoraxe sound right?

(we want the user query+response separately despite having the conversation id(with which we could go look up the entire conversation via our transcript recording feature) because we want to see the exact part of the conversation they are providing feedback on, in the case of a conversation that spans multiple follow up queries.  It's also possible some customers will enable feedback reporting, but will disallow the general collection of transcripts.  In that case, this will allow us to still have at least some context for the feedback)

tests/integration/test_feedback.py
bparees: i agree it's not worth investing in

tests/integration/test_feedback.py
bparees: > I didn't implement the tests like test_feedback_no_payload_send. Those are basically testing the FastAPI/pydantic behavior. Unless there is a special logic for handling/validating the input parameters, I'm not sure what is the actual benefit of such tests.

there are two benefits to writing such tests, imho:

1) writing the test forces us to write down(in the test) what we expect the failure behavior to be, which can lead to a useful discussion about whether that failure behavior is what we want it to be(e.g. is the error code we want? is there a useful error message associated that will help someone understand what went wrong?).  In general tests(especially integration and e2e tests) are a way for us to express the expected behavior of the system

2) if we accidentally break things such that an empty payload does *not* result in failure, that would be bad.  So the point of the test is to ensure that we fail when we expect to fail, because if we don't fail when we should have, that's a problem.

I am not going to claim they are the most important tests, but i think they do provide value.

bparees (commented):
i'd like to see an e2e test for this.  We want to ensure the feedback submission endpoint is properly auth-secured, for one thing.

Having a test that actually validates that the feedback was recorded is going to be a challenge when we are running on a cluster.  That might be a good enough reason to keep the "retrieve" feedback apis around, but as i noted in my comments, they will need to have a different auth permission enforcement.

ols/app/endpoints/feedback.py
onmete: replaced for uuid

ols/app/models/models.py
onmete: If I get it right, those are basically two different feedbacks. One is +1/-1 to a response and the other is generic. So maybe accept two models by endpoint (to avoid allowing posting "nothing")?
```python
# thumbs feedback
conversation_id: str
user_question: str
llm_response: str
sentiment: int

# generic feedback
user_feedback: str
```
Or maybe better to split the endpoint into two.
WDYT?

bparees: > Everything is optional because the freeform feedback is not bound to any conversation right? 

Feedback is only given in reaction to an exchange in a conversation.  Even the "Freeform" text is freeform text about the interaction.

they aren't two different kinds of feedback object.

```
conversationid: required str
user_question: required str
llm_response: required str
sentiment: required int
freeform_text: optional str
```

onmete:
In the latest changes, I've updated the e2e script to export the storage path via env. This is then picked in the e2e test.
Can't say I like it, but it works. 
The e2e-ols-cluster fails as it runs with a different script/config? I need to check. But I believe the same trick can be done.
With that, I can remove the list/del endpoints.

The other way would be to use the list endpoint in the e2e test and update the permissions of list/del endpoints (and add more integration tests to those).

Up to you @bparees.

ols/app/models/models.py
onmete: Hmm, ok. So the user can either provide thumbs or text feedback for response. The `conversation_id`, `user_question` and `llm_response` are mandatory. And `sentiment` or `freeform_text` will be provided (are optional from the model point of view). Correct?

bparees: > Hmm, ok. So the user can either provide thumbs or text feedback for response. The conversation_id, user_question and llm_response are mandatory. And sentiment or freeform_text will be provided (are optional from the model point of view). Correct?

well the way i wrote it above, thumbs up/down would be mandatory.  But thinking about it some more, i'm fine with it being optional.  (but if they don't specify sentiment and they don't specify freeform, that's an invalid feedback submission.  We need at least one of them)

bparees:
> Cluster e2e test failed with I see PermissionError: [Errno 13] Permission denied: 'insights', I'm tried to fix that with SCC, but it still fails with the [same](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/538/pull-ci-openshift-lightspeed-service-main-e2e-ols-cluster/1766035865886265344/artifacts/e2e-ols-cluster/e2e/artifacts/cluster/podlogs/ols-54bf664cd5-vshbr.log).
I need some help to figure this out @bparees.

it's not an scc problem, it's just a basic linux file permission issue, see:
https://redhat-internal.slack.com/archives/C068JAU4Y0P/p1709913851565069

onmete:
@bparees I'll do also https://issues.redhat.com/browse/OLS-387 as part of this PR. Otherwise, I would need to update all configs (tls, ...) with the `feedback_storage_location` to ensure it is not failing in some cases.

ols/app/endpoints/feedback.py
```diff
@@ -1,33 +1,197 @@
...
-        Response indicating the status of the feedback.
+        Response indicating the status of the feedback storage request.
     """
-    logger.info(
-        f"feed back received for conversation '{feedback_request.conversation_id}', "
-        f"feedback blob: {feedback_request.feedback_object}"
-    )
+    logger.info(f"feedback received {feedback_request}")
+
+    user_id = auth[0]
```
tisnik, line 167: it's used on several places. Probably time to refactor it as `auth[0]` is a bit wacky

ols/app/endpoints/feedback.py
```diff
@@ -1,33 +1,197 @@
...
+
+    feedbacks = list_feedbacks()
+
+    return FeedbacksListResponse(feedbacks=feedbacks)
+
+
 @router.post("")
-def user_feedback(
-    feedback_request: FeedbackRequest, auth: Any = Depends(auth_dependency)
+def store_user_feedback(
```
tisnik, line 148: should it return 201 Created?

tisnik (commented):
Also it looks like some HTTP states are not documented. Most new endpoints should return 200/201/404

docs/openapi.json
```diff
@@ -132,6 +176,49 @@
...
+                        "description": "Successful Response",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/FeedbackResponse"
+                                }
+                            }
+                        }
+                    },
+                    "422": {
```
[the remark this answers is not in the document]
onmete: No endpoint has 404 documented. If it should be there, I think we need to tweak the schema generator.

ols/app/endpoints/feedback.py
onmete: Do you have a story in mind?

ols/app/endpoints/feedback.py
onmete: I believe that this question is related to https://github.com/openshift/lightspeed-service/pull/538#discussion_r1530896191
It comes from our schema generator

tests/scripts/test-e2e-standalone.sh
```diff
@@ -43,6 +43,9 @@ mkdir -p $RAG_INDEX_DIR
 wget $RAG_INDEX
 unzip -j local.zip -d $RAG_INDEX_DIR
 rm -f local.zip
+# configure feedback storage location
+export FEEDBACK_STORAGE_LOCATION="$ARTIFACT_DIR/insights/feedback"
```
bparees, line 47: does this path get created by OLS if it does not exist?

bparees (changes requested):
thank you for including an e2e test :)

ols/app/endpoints/feedback.py
```diff
@@ -1,33 +1,197 @@
...
+    Args:
+        request (Request): The FastAPI request object.
+
+    Returns:
+        True if feedback is enabled, False otherwise.
+
+    Raises:
+        HTTPException: If feedback is disabled.
+    """
+    feedback_enabled = not config.ols_config.user_data_collection.feedback_disabled
```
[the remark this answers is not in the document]
onmete: I assume you mean `get_feedback_status`. I'll update.

ols/app/endpoints/feedback.py
```diff
@@ -1,33 +1,197 @@
...
+    if not feedback_enabled:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Feedback is currently disabled.",
+        )
+    return feedback_enabled
+
+
+def get_feedback_status() -> bool:
```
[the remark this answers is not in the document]
onmete: Nope, that is alright. And Pavel [should be](https://github.com/openshift/lightspeed-service/pull/538#discussion_r1530698077) happy too. :)

ols/app/endpoints/feedback.py
tisnik: schema is generated from our sources, not vice versa;)

tests/e2e/test_api.py
```diff
@@ -577,3 +577,55 @@ def test_improper_token():
         headers={"Authorization": "Bearer wrong-token"},
     )
     assert response.status_code == requests.codes.forbidden
+
+
+def test_feedback() -> None:
```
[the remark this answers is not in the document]
onmete: added

tests/e2e/test_api.py
```diff
@@ -577,3 +577,75 @@ def test_improper_token():
...
+
+    # check the feedback store is empty again
+    removed_feedback = client.get("/v1/feedback/list", timeout=BASIC_ENDPOINTS_TIMEOUT)
+    assert removed_feedback.status_code == requests.codes.ok
+    assert "feedbacks" in removed_feedback.json()
+    assert len(removed_feedback.json()["feedbacks"]) == 0
+
+
+@pytest.mark.cluster
+def test_feedback_can_post_with_wrong_token():
```
bparees, line 635:
```suggestion
def test_feedback_cannot_post_with_wrong_token():
```

bparees (commented):
one test naming nit and then lgtm

PR #549, opened 2024-03-06
OLS-165: Response sanity check

asamal4:
Validate any deviation in LLM response.
A stand alone python script to check responses for different scenario / different models. The json file must contain question/answer pair for given scenario & model.

- Load pre-defined question/answer pair from json file.
- Call API with question/query
- Check similarity between answer from file & llm response.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-165](https://issues.redhat.com//browse/OLS-165)
- Closes #

## Testing

### DIFF
--- a/scripts/validate_response.py
+++ b/scripts/validate_response.py
@@ -1,0 +1,147 @@
+"""Response validation using pre-defined question/answer pair."""
+
+import argparse
+import json
+import os
+import sys
+from collections import defaultdict
+
+import requests
+from llama_index.embeddings.huggingface import HuggingFaceEmbedding
+from pandas import DataFrame
+from scipy.spatial.distance import cosine, euclidean
+
+from ols.constants import NO_RAG_CONTENT_RESP
+
+
+# TODO: Add more question/answer pair
+def _args_parser(args):
+    """Arguments parser."""
+    parser = argparse.ArgumentParser(description="Response validation module.")
+    parser.add_argument(
+        "-s",
+        "--scenario",
+        choices=["with_rag", "without_rag"],
+        default="with_rag",
+        help="Scenario for which responses will be evaluated.",
+    )
+    parser.add_argument(
+        "-m",
+        "--model",
+        choices=["gpt", "granite"],
+        default="gpt",
+        help="Model for which responses will be evaluated.",
+    )
+    parser.add_argument(
+        "-t",
+        "--threshold",
+        type=lambda val: float(val),
+        default=0.25,
+        help="Threshold value to be used for similarity score.",
+    )
+    parser.add_argument(
+        "-n",
+        "--num_questions",
+        type=lambda val: int(val),
+        default=2,
+        help="Number of questions to be validated.",
+    )
+    return parser.parse_args(args)
+
+
+class ResponseValidation:
+    """Validate LLM response."""
+
+    def __init__(self):
+        """Initialize."""
+        self._embedding_model = HuggingFaceEmbedding("BAAI/bge-base-en")
+
+    def get_similarity_score(self, response, answer):
+        """Calculate similarity score between two strings."""
+        response = response.replace(NO_RAG_CONTENT_RESP, "")
+        res_vec = self._embedding_model.get_text_embedding(response)
+        ans_vec = self._embedding_model.get_text_embedding(answer)
+
+        # Distance score
+        cos_score = cosine(res_vec, ans_vec)
+        euc_score = euclidean(res_vec, ans_vec)
+
+        # Naive length consideration with reduced weightage.
+        len_res, len_ans = len(response), len(answer)
+        len_score = (abs(len_res - len_ans) / (len_res + len_ans)) * 0.1
+
+        score = len_score + (cos_score + euc_score) / 2
+        # TODO: Consider both contextual/non-contextual embedding.
+
+        print(
+            f"cos_score: {cos_score}, "
+            f"euc_score: {euc_score}, "
+            f"len_score: {len_score}\n"
+            f"final_score: {score}"
+        )
+        return score
+
+    def get_response_quality(self, args, qa_pairs):
+        """Get response quality."""
+        result_dict = defaultdict(list)
+
+        for idx in range(min(len(qa_pairs), args.num_questions)):
+            question = qa_pairs[idx]["question"]
+            answer = qa_pairs[idx]["answer"]
+
+            response = requests.post(
+                # API question validator can be disabled.
+                "http://localhost:8080/v1/query",
+                json={"query": question},
+                timeout=90,
+            ).json()["response"]
+
+            print(f"Calculating score for query: {question}")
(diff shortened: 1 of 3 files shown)


scripts/question_answer_pair.json
```diff
@@ -0,0 +1,16 @@
+{
+    "gpt": {
+        "rag": [
+            {
+                "que": "what is openshift virtualization?",
```
tisnik, line 5: why not to use full attribute names?

scripts/validate_response.py
```diff
@@ -0,0 +1,97 @@
...
+#         "no_rag": [
+#             {
+#                 "que": Q1,
+#                 "ans": A2,
+#             },
+#         ]
+#     },
+#     "granite": {}
+# }
+THRESHOLD_SCORE = 0.5
```
tisnik, line 29: probably made it configurable?

scripts/validate_response.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Distance score
+        cos_score = cosine(res_vec, ans_vec)
+        euc_score = euclidean(res_vec, ans_vec)
+
+        # Naive length consideration with reduced weightage.
+        len_res, len_ans = len(response), len(answer)
+        len_score = (abs(len_res - len_ans) / (len_res + len_ans)) * 0.1
+
+        score = len_score + (cos_score + euc_score) / 2
```
tisnik, line 81: is it based on some heuristic?

is not scalar product used (instead of len_score)?

scripts/validate_response.py
```diff
@@ -0,0 +1,102 @@
...
+        # TODO: Consider both contextual/non-contextual embedding.
+
+        print(
+            f"Question: {question}\n"
+            f"cos_score: {cos_score}, euc_score: {euc_score}, len_score: {len_score}, "
+            f"final_score: {score}"
+        )
```
tisnik, line 88: later we might problably export to CSV or any other format, to be able to see + compare previous results. WDYT?

scripts/validate_response.py
```diff
@@ -0,0 +1,102 @@
...
+        "--threshold",
+        type=lambda val: float(val),
+        default=0.25,
+        help="Threshold value to be used for similarity score.",
+    )
+    return parser.parse_args(args)
+
+
+def main():
+    """Validate llm response."""
```
tisnik, line 41: LLM

scripts/validate_response.py
```diff
@@ -0,0 +1,102 @@
...
+    args = args_parser(sys.argv[1:])
+    print(f"Arguments passed: {args}")
+
+    with open("./scripts/question_answer_pair.json") as qna_f:
```
tisnik, line 45: will it work if the script will be started as `python validate_response.py`? (ie from current dir)

scripts/validate_response.py
```diff
@@ -0,0 +1,102 @@
...
+        if args.scenario == "without_rag":
+            response = response.replace(
+                "The following response was generated without access to reference content:\n\n",
+                "",
+            )
+
+        res_vec = embedding_model.get_text_embedding(response)
+        ans_vec = embedding_model.get_text_embedding(answer)
+
+        # Distance score
```
tisnik, line 73: prob. put the whole score computation into a function?

tisnik (commented):
it looks ok. just few nitpicks.

Btw would it make sense to have the ability to specify threshold per question (as optional value)?

scripts/question_answer_pair.json
bparees, line 1: please come up with a better directory structure for this content

scripts/validate_response.py
bparees, line 1: is there a reason this can't be written as an e2e test and run along w/ the rest of our e2es?

scripts/validate_response.py
tisnik: thank you

scripts/validate_response.py
tisnik: it might work from both directories. basically the script knows that the JSON is in the same directory as script itself. so it can use this info: `os.path.dirname(__file__)`

scripts/validate_response.py
tisnik: thank you

scripts/validate_response.py
```diff
@@ -0,0 +1,150 @@
...
+    result_df = ResponseValidation().get_response_quality(args, qa_pairs)
+
+    if len(result_df) > 0:
+        result_dir = f"{parent_dir}/test_results"
+        os.makedirs(result_dir, exist_ok=True)
+        result_file = (
+            f"{result_dir}/question_answer_result_{args.scenario}_{args.model}.csv"
+        )
+        result_df.to_csv(result_file, index=False)
+        print(f"Result is saved to {result_file}")
```
vishnuchalla, line 138: What's the idea behind persisting these results in a file. Why not just log them in whatever format we want? Because that file will be created in PROW environment during runtime and will be destroyed after the test ends.

scripts/validate_response.py
```diff
@@ -0,0 +1,150 @@
...
+
+        print(
+            f"cos_score: {cos_score}, "
+            f"euc_score: {euc_score}, "
+            f"len_score: {len_score}\n"
+            f"final_score: {score}"
+        )
+        return score
+
+    def get_response_quality(self, args, qa_pairs):
```
vishnuchalla, line 81: Something to start with, this simple evaluation technique seems good. I hope we have a #TODO action item to look for frameworks like RAGAS etc, to do similar evaluations in future.

bparees (commented):
The question/answer json file is a good start towards what i was hoping to see from OLS-165.  So what I like about what you've done so far:

1) we've started a collection of Q+A we can baseline the performance against
2) it's divided into w/ and w/o RAG
3) you've got a benchmark script we can run to generate the evaluation scores so we can measure as we change things

what i would like to see changed is for the e2e tests to load the Q+A from the json file so we are sharing a centralized list of known expected questions/answers.  The e2e tests can use a particular single question from the list (or really two, one for rag, one for non-rag)

ideally we'd run all the Q/A pairs in e2e but as you noted cost becomes a concern.

i'll also be slightly surprised if you don't hit issues w/ the responses being significantly different between providers, but maybe you've got the similarity score threshold set low enough that it isn't an issue.

scripts/validate_response.py
vishnuchalla: Understood.

bparees:
> I remember we were running some job weekly once.. We can run the whole script there. And only few questions regularly when we raise PR. We can do this when we have enough Q/A pair and improvised validation.

at present we're running a couple e2e jobs every day, but we can definitely setup an "llm evaluation" job to run weekly.

PR #566, opened 2024-03-08
[Bug fix] Add command to keep redis container awake

vishnuchalla:
## Description

[Bug fix] Adding a  command to keep redis container awake. Currently when a redis pod start it immediately get terminated as we do have a entrypoint command that keeps it awake. So fixing it with a sleep command and the actual redis server will be spun up with required SSL/TLS and auth args from operator's code.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

## Testing
Tested integrating with https://github.com/openshift/lightspeed-operator/pull/25

### DIFF
--- a/Containerfile.redis
+++ b/Containerfile.redis
@@ -19,3 +19,6 @@ RUN make BUILD_TLS=yes && \
 
 # Expose the default Redis port
 EXPOSE 6379
+
+# Start redis server
+ENTRYPOINT [ "redis-server", "/etc/redis/redis.conf" ]


Containerfile.redis
```diff
@@ -19,3 +19,6 @@ RUN make BUILD_TLS=yes && \
 
 # Expose the default Redis port
 EXPOSE 6379
+
+# Keep the container awake
+ENTRYPOINT [ "sleep", "infinity" ]
```
raptorsun, line 24: how about spining up Redis with default config? 
`ENTRYPOINT [ "redis-server", "/etc/redis/redis.conf" ]`

vishnuchalla: Sounds good if someone want to use our image alone.

tisnik (commented):
it looks correct, thank you

PR #575, opened 2024-03-11
Remove the constrains for providers in the config

onmete:
## Description

As one of the main benefits of the pluggability is isolated change, I decided to disable this check when I was doing the demo about llm providers' pluggability.

This is up for consideration.
We can either disable this check (postponing failure to when someone is trying to use an invalid provider) or whenever someone develops a new provider, he/she needs to update the list.

WDYT?

## Type of change

- [X] Refactor


bparees:
i don't think we should disable this because it means you don't find out you have an invalid/unsupported configuration until a request comes into OLS.  

it's much easier to understand/debug failures when you get them during startup and can immediately conclude that it's related to the env/configuration, vs failures that occur in the request flow.

admittedly it should still be pretty obvious, but given all the other changes required to add support for a new provider, adding the provider to this list doesn't seem onerous or unreasonable.

i'm open to more discussion about it, but my inclination is against removing this config validation logic.

onmete:
Would be worth the shot to try to get those dynamically?

bparees:
> Would be worth the shot to try to get those dynamically?

it doesn't feel worth the effort to me.  it's literally a one line change to add the new provider constant to the list.

onmete:
@bparees this one then: https://github.com/openshift/lightspeed-service/pull/598

PR #584, opened 2024-03-12
OLS-136: data to ingress

onmete:
## Description

Script that collects data, packages it, and sends it to ingress.

TODO:

- [x] - code needed from https://github.com/openshift/lightspeed-service/pull/673
- [x] - CI secret https://github.com/openshift/release/pull/50723

### DIFF
--- a/examples/openshift-lightspeed-tls.yaml
+++ b/examples/openshift-lightspeed-tls.yaml
@@ -93,6 +93,7 @@ spec:
           secret:
             secretName: lightspeed-certs
       containers:
+      # this example is without the data collection sidecar
       - image: quay.io/openshift/lightspeed-service-api:latest
         imagePullPolicy: Always
         name: lightspeed-w-rag
@@ -198,6 +199,9 @@ data:
         tls_key_path: /app-root/certs/tls.key
       default_provider: openai
       default_model: gpt-3.5-turbo
+      user_data_collection:
+        feedback_disabled: true
+        transcripts_disabled: true
     dev_config:
       enable_dev_ui: true
       # llm_temperature_override: 0
--- a/ols/user_data_collection/data_collector.py
+++ b/ols/user_data_collection/data_collector.py
@@ -1,0 +1,323 @@
+"""Collect insights and upload it to the Ingress service.
+
+It waits 5 min after startup before collects the data. Then it collects
+data after specified interval.
+
+When CP_OFFLINE_TOKEN is provided (either for prod or stage), it is used
+for ingress authentication instead of cluster pull-secret.
+"""
+
+import base64
+import io
+import json
+import logging
+import os
+import pathlib
+import sys
+import tarfile
+import time
+from typing import Any, Callable
+
+import kubernetes
+import requests
+
+OLS_USER_DATA_PATH = os.environ["OLS_USER_DATA_PATH"]
+OLS_USER_DATA_COLLECTION_INTERVAL = int(
+    os.environ.get("OLS_USER_DATA_COLLECTION_INTERVAL", 2 * 60 * 60)
+)  # 2 hours in seconds
+
+INITIAL_WAIT = 60 * 5  # 5 minutes in seconds
+RUN_WITHOUT_INITIAL_WAIT = bool(
+    os.environ.get("RUN_WITHOUT_INITIAL_WAIT", "false").lower() == "true"
+)
+
+INGRESS_ENV = os.environ.get("INGRESS_ENV", "stage")  # prod/stage
+if INGRESS_ENV not in {"prod", "stage"}:
+    raise ValueError(
+        f"Unknown value in INGRESS_ENV: {INGRESS_ENV}. Allowed: prod, stage"
+    )
+INGRESS_TIMEOUT = 30  # seconds
+INGRESS_BASE_DELAY = 60  # exponential backoff parameter
+INGRESS_MAX_RETRIES = 3  # exponential backoff parameter
+
+CP_OFFLINE_TOKEN = os.environ.get("CP_OFFLINE_TOKEN")
+REDHAT_SSO_TIMEOUT = 5  # seconds
+
+# TODO: OLS-473
+# OLS_USER_DATA_MAX_SIZE = 100 * 1024 * 1024  # 100 MB
+
+if INGRESS_ENV == "stage" and not CP_OFFLINE_TOKEN:
+    raise ValueError("CP_OFFLINE_TOKEN is required for stage environment")
+
+
+logging.basicConfig(
+    level=os.environ.get("LOG_LEVEL", "info").upper(),
+    stream=sys.stdout,
+    format="[%(asctime)s] %(levelname)s: %(message)s",
+)
+# silence libs
+# - urllib3 - we don't care about those debug posts
+# - kubernetes - prints resources content when debug, causing secrets leak
+logging.getLogger("kubernetes").setLevel(logging.WARNING)
+logging.getLogger("urllib3").setLevel(logging.WARNING)
+
+logger = logging.getLogger(__name__)
+
+
+class ClusterPullSecretNotFoundError(Exception):
+    """Cluster pull-secret is not found."""
+
+
+def get_ingress_upload_url() -> str:
+    """Get the Ingress upload URL based on the environment."""
+    upload_endpoint = "api/ingress/v1/upload"
+    if INGRESS_ENV == "prod":
+        return "https://console.redhat.com/" + upload_endpoint
+    elif INGRESS_ENV == "stage":
+        return "https://console.stage.redhat.com/" + upload_endpoint
+    else:
+        raise ValueError(f"Unknown value in INGRESS_ENV: {INGRESS_ENV}")
+
+
+def access_token_from_offline_token(offline_token: str) -> str:
+    """Generate "access token" from the "offline token".
+
+    Offline token can be generated at:
+        prod - https://access.redhat.com/management/api
+        stage - https://access.stage.redhat.com/management/api
+
+    Args:
+        offline_token: Offline token from the Customer Portal.
+
+    Returns:
+        Refresh token.
+    """
+    if INGRESS_ENV == "stage":
+        url = "https://sso.stage.redhat.com/"
+    else:
+        url = "https://sso.redhat.com/"
+    endpoint = "auth/realms/redhat-external/protocol/openid-connect/token"
+    data = {
+        "grant_type": "refresh_token",
+        "client_id": "rhsm-api",
+        "refresh_token": offline_token,
+    }
+
+    response = requests.post(url + endpoint, data=data, timeout=REDHAT_SSO_TIMEOUT)
+    if response.status_code == requests.codes.ok:
+        return response.json()["access_token"]
+    else:
+        raise Exception(f"Failed to generate access token. Response: {response.json()}")
+
+
+def get_cloud_openshift_pull_secret() -> str:
+    """Get the pull secret token from the cluster."""
+    kubernetes.config.load_incluster_config()
+    v1 = kubernetes.client.CoreV1Api()
+
+    # get the pull-secret from openshift-config namespace
+    secret = v1.read_namespaced_secret("pull-secret", "openshift-config")
+    try:
+        dockerconfigjson = secret.data[".dockerconfigjson"]
+        dockerconfig = json.loads(base64.b64decode(dockerconfigjson).decode("utf-8"))
+        return dockerconfig["auths"]["cloud.openshift.com"]["auth"]
+    except KeyError:
+        logger.error("failed to get token from cluster pull-secret")
+        raise ClusterPullSecretNotFoundError
+
+
+def get_cluster_id() -> str:
+    """Get the cluster_id from the cluster."""
+    kubernetes.config.load_incluster_config()
+    custom_objects_api = kubernetes.client.CustomObjectsApi()
+
+    # get the cluster version object
+    version_data = custom_objects_api.get_cluster_custom_object(
+        "config.openshift.io", "v1", "clusterversions", "version"
+    )
+    try:
+        return version_data["spec"]["clusterID"]
+    except KeyError:
+        logger.error("failed to get cluster_id from cluster")
+        return "unknown"
+
+
+def collect_ols_data_from(location: str) -> list[pathlib.Path]:
+    """Collect files from a given location.
+
+    Args:
+        location: Path to the directory to be searched for files.
+
+    Returns:
+        List of paths to the collected files.
+
+    Only JSON files from the 'feedback' and 'transcripts' directories are collected.
+    """
+    files = []
+
+    files += list(pathlib.Path(location).glob("feedback/*.json"))
+    files += list(pathlib.Path(location).glob("transcripts/*/*/*.json"))
+
+    return files
+
+
+def package_files_into_tarball(
+    file_paths: list[pathlib.Path], path_to_strip: str
+) -> io.BytesIO:
+    """Package specified directory into a tarball.
+
+    Args:
+        file_paths: List of paths to the files to be packaged.
+        path_to_strip: Path to be stripped from the file paths (not
+            included in the archive).
+
+    Returns:
+        BytesIO object representing the tarball.
+    """
+    tarball_io = io.BytesIO()
+
+    with tarfile.open(fileobj=tarball_io, mode="w:gz") as tar:
+        # arcname parameter is set to a stripped path to avoid including
+        # the full path of the root dir
+        for file_path in file_paths:
+            # skip symlinks as those are a potential security risk
+            if not file_path.is_symlink():
+                tar.add(
+                    file_path, arcname=file_path.as_posix().replace(path_to_strip, "")
+                )
+
+        # add magic file for identification of our archive on the CCX side
+        empty_file = tarfile.TarInfo("openshift_lightspeed.json")
+        empty_file.size = 0
+        tar.addfile(empty_file)
+
+    tarball_io.seek(0)
+
+    return tarball_io
+
+
+def exponential_backoff_decorator(max_retries: int, base_delay: int) -> Callable:
+    """Exponential backoff decorator."""
+
+    def decorator(func: Callable) -> Callable:
+        def wrapper(*args: Any, **kwargs: Any) -> None:
+            retries = 0
+            while retries < max_retries:
+                try:
+                    func(*args, **kwargs)
+                    return
+                except Exception as e:
+                    logger.error(f"attempt {retries + 1} failed with error: {e}")
+                    retries += 1
+                    delay = base_delay * 2**retries
+                    logger.info(f"retrying in {delay} seconds...")
+                    time.sleep(delay)
+            logger.error("max retries reached, operation failed.")
+
+        return wrapper
+
+    return decorator
+
+
+@exponential_backoff_decorator(
+    max_retries=INGRESS_MAX_RETRIES, base_delay=INGRESS_BASE_DELAY
+)
+def upload_data_to_ingress(tarball: io.BytesIO) -> requests.Response:
+    """Upload the tarball to a Ingress.
+
+    Args:
+        tarball: BytesIO object representing the tarball to be uploaded.
+
+    Returns:
+        Response object from the Ingress.
+    """
+    logger.info("sending collected data")
+    payload = {
+        "file": (
+            "ols.tgz",
+            tarball.read(),
+            "application/vnd.redhat.openshift.periodic+tar",
+        ),
+    }
+
+    if CP_OFFLINE_TOKEN:
+        logger.debug("using CP offline token to generate refresh token")
+        token = access_token_from_offline_token(CP_OFFLINE_TOKEN)
+        cluster_id = "unknown"
+    else:
+        logger.debug("using cluster pull secret to authenticate")
+        token = get_cloud_openshift_pull_secret()
+        cluster_id = get_cluster_id()
+
+    with requests.Session() as s:
+        s.headers = {
+            # TODO: OLS-486 User agent needs to be set as insights operator,
+            # otherwise JWT token is not accepted - we use just random hash
+            # "User-Agent": f"openshift-lightspeed-operator/user-data-collection cluster/{cluster_id}",  # noqa E501
+            "User-Agent": f"insights-operator/abcdefgijklmnopqrstuvwxyz01234567891011 cluster/{cluster_id}",  # noqa: E501
+            "Authorization": f"Bearer {token}",
+        }
+        if CP_OFFLINE_TOKEN:
+            # when authenticating with token, user-agent is not accepted
+            # causing "UHC services authentication failed"
+            del s.headers["User-Agent"]
+
+        response = s.post(
+            url=get_ingress_upload_url(),
+            files=payload,
+            timeout=INGRESS_TIMEOUT,
+        )
+
+    if response.status_code != 202:
+        raise requests.exceptions.HTTPError(
+            f"data upload failed with response: {response.json()}"
+        )
+
+    request_id = response.json()["request_id"]
+    logger.info(f"data uploaded with request_id: '{request_id}'")
+
+    return response
+
+
+def delete_data(file_paths: list[pathlib.Path]) -> None:
+    """Delete files from the provided paths.
+
+    Args:
+        file_paths: List of paths to the files to be deleted.
+    """
+    for file_path in file_paths:
+        logger.debug(f"removing '{file_path}'")
+        file_path.unlink()
+        if file_path.exists():
+            logger.error(f"failed to remove '{file_path}'")
+
+
+def gather_ols_user_data(data_path: str) -> None:
+    """Gather OLS user data and upload it to the Ingress service."""
+    collected_files = collect_ols_data_from(data_path)
+    if collected_files:
+        logger.info(f"collected {len(collected_files)} files from '{data_path}'")
+        tarball = package_files_into_tarball(collected_files, path_to_strip=data_path)
+        try:
+            upload_data_to_ingress(tarball)
+            delete_data(collected_files)
+            logger.info("uploaded data removed")
+        except ClusterPullSecretNotFoundError:
+            logger.error(
+                "missing cluster pull secret, upload and data removal canceled"
+            )
+        # TODO: OLS-473
+        # ensure_data_folder_is_not_bigger_than(OLS_USER_DATA_MAX_SIZE)
+    else:
+        logger.info(f"'{data_path}' contains no data, nothing to do...")
+
+
+if __name__ == "__main__":
+    if not RUN_WITHOUT_INITIAL_WAIT:
+        logger.info(
+            "collection script started, waiting 5 minutes before first collection"
+        )
+        time.sleep(INITIAL_WAIT)
+    while True:
+        gather_ols_user_data(OLS_USER_DATA_PATH)
+        time.sleep(OLS_USER_DATA_COLLECTION_INTERVAL)
--- a/tests/config/cluster_install/ols_manifests.yaml
+++ b/tests/config/cluster_install/ols_manifests.yaml
@@ -60,6 +60,24 @@ spec:
             mountPath: /app-root/certs
           - name: ols-user-data
             mountPath: /app-root/ols-user-data
+      - image: "$OLS_IMAGE"
+        name: ols-sidecar-user-data-collector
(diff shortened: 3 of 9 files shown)


Containerfile.udc
bparees, line 1: 1) we don't need to build an entirely new image just to set the command.  If you want to run a sidecar container with a different command, you can just run the sidecar w/ a different command specified in the container spec.
2) i don't really understand why we need to run a sidecar for this.  Why can't we just invoke the collection logic from the main ols process?  adding the sidecar container seems like unnecessary overhead, including having to wire the manifest configuration through the operator.

bparees: i don't think it would be too bad, but you're right that running it as a sidecar (using the same image with a different launch command) is probably simpler, so sure, proceed w/ that for now.

onmete:
@tisnik @bparees 
I've pushed some changes. This version seems to be working. It sends the tarball to ingress. Some bits are still missing.
But I need some help with testing. I've added there a docker compose file, that does the e2e testing.

You can try that with 
```bash
docker-compose -f ols/user_data_collection/docker-compose.yaml up --build
```
and then
```bash
curl -X POST -H "Content-Type: application/json" -d '{"conversation_id": "1", "user_question": "a", "llm_response": "b", "sentiment": 1}' http://0.0.0.0:8080/v1/feedback
```
that results in
```bash
user_data_collection-data_collector-1  | INFO:__main__:collected data from '/app-root/data'- '1' files
user_data_collection-data_collector-1  | INFO:__main__:sending collected data
user_data_collection-fake-ingress-1    | INFO:root:recieved something
user_data_collection-fake-ingress-1    | INFO:root:feedback/ae0455db-e3c6-48c6-a2d8-9c90d96a852a.json
user_data_collection-fake-ingress-1    | INFO:root:openshift_lightspeed.json
user_data_collection-fake-ingress-1    | INFO:     172.18.0.4:42562 - "POST /test_endpoint HTTP/1.1" 202 Accepted
user_data_collection-data_collector-1  | INFO:__main__:data uploaded with request_id: 'some-request-id'
user_data_collection-data_collector-1  | DEBUG:__main__:removing '/app-root/data/feedback/ae0455db-e3c6-48c6-a2d8-9c90d96a852a.json'
user_data_collection-data_collector-1  | INFO:__main__:uploaded data removed
user_data_collection-data_collector-1  | INFO:__main__:'/app-root/data' contains no data, nothing to do...
```
I don't have a good idea of how to translate it to our e2e suite, so any ideas are welcome. :)

examples/openshift-lightspeed-tls.yaml
```diff
@@ -100,6 +100,16 @@ spec:
             mountPath: /app-root/config/openai
           - name: tls-certs
             mountPath: /app-root/certs
+      - name: user-data-collector-sidecar
+        image: quay.io/openshift/lightspeed-service-api-udc:latest
```
bparees, line 104: should not be a separate/different image pullspec.

examples/openshift-lightspeed-tls.yaml
```diff
@@ -100,6 +100,16 @@ spec:
...
+        # TODO: launch command
+        env:
+          - name: OLS_USER_DATA_PATH
+            value: /app-root/ols-user-data
+          # TODO: token?
+        volumeMounts:
+          - name: ols-user-data
```
bparees, line 111: you need to define the volume, and also the ols main container also needs to mount it, no?

examples/openshift-lightspeed-tls.yaml
```diff
@@ -100,6 +100,16 @@ spec:
             mountPath: /app-root/config/openai
           - name: tls-certs
             mountPath: /app-root/certs
+      - name: user-data-collector-sidecar
+        image: quay.io/openshift/lightspeed-service-api-udc:latest
+        # TODO: launch command
+        env:
+          - name: OLS_USER_DATA_PATH
```
bparees, line 107: should be able to load/read the same config.yaml that the main ols container is using.

tests/config/cluster_install/ols_manifests.yaml
```diff
@@ -60,6 +60,16 @@ spec:
             mountPath: /app-root/certs
           - name: ols-user-data
             mountPath: /app-root/ols-user-data
+      - name: user-data-collector-sidecar
+        image: quay.io/openshift/lightspeed-service-api-udc:latest
```
bparees, line 64: should not be a new/separate image

bparees (commented):
can we use the stage environment for e2e testing purposes?  (upload data to stage)

examples/openshift-lightspeed-tls.yaml
onmete: Yes, there is a todo to just change entrypoint/command. The image should be the main ols.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,177 @@
...
+        List of paths to the collected files.
+
+    Only JSON files from the 'feedback' and 'transcripts' directories are collected.
+    """
+    files = []
+
+    files += list(pathlib.Path(location).glob("feedback/*.json"))
+    files += list(pathlib.Path(location).glob("transcripts/*/*/*.json"))
+
+    return files
```
vishnuchalla, line 62: What's the difference between feedback and transcripts?

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,177 @@
...
+OLS_USER_DATA_COLLECTION_INTERVAL = int(
+    os.environ.get("UDC_COLLECTION_INTERVAL", 2 * 60 * 60)
+)  # 2 hours in seconds
+
+# available ingress urls are:
+# - https://console.redhat.com/ - production
+# - https://console.stage.redhat.com/ - staging/dev
+# NOTE: Token is not working for stage...
+INGRESS_URL = os.environ.get("INGRESS_URL", "https://console.redhat.com/")
+INGRESS_ENDPOINT = os.environ.get("INGRESS_ENDPOINT", "api/ingress/v1/upload")
```
vishnuchalla, line 35: when we say data to ingress. Where does this data actually go to? And how to look at those metrics?

ols/user_data_collection/auth.py
```diff
@@ -0,0 +1,68 @@
...
+
+import requests
+
+# import urllib3
+from kubernetes import client, config
+
+REDHAT_SSO_TIMEOUT = 5
+
+
+def get_current_namespace() -> str:
```
vishnuchalla, line 14: This function seems to be unused.

ols/user_data_collection/data_collector.py
bparees: it goes to our CCX/observability ceph bucket after making its way through the ccx/observability ingress pipeline.

ols/user_data_collection/data_collector.py
bparees: feedback is the user telling this "this response was good/bad because....", along w/ the specific interaction the user had (what they asked, what the llm responded with)

transcripts are us recording all the interactions between the user and the llm

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,245 @@
...
+OLS_USER_DATA_PATH = os.environ.get("OLS_USER_DATA_PATH", "./user-data/")
+
+OLS_USER_DATA_COLLECTION_INTERVAL = int(
+    os.environ.get("UDC_COLLECTION_INTERVAL", 2 * 60 * 60)
+)  # 2 hours in seconds
+
+# available ingress urls are:
+# - https://console.redhat.com/ - production - works with refresh token
+# - https://console.stage.redhat.com/ - staging/dev - doesn't work with refresh token
+INGRESS_URL = os.environ.get("INGRESS_URL", "https://console.redhat.com/")
```
bparees, line 28: this one is ok to leave as an env var since there should be no reason to expect us to use any other value except for testing purposes.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,245 @@
+"""Collect insights and upload it to the Ingress service.
+
+It collects the data on startup and then it waits for the specified interval.
```
bparees, line 3: don't collect on startup.  have a 5 min delay before the first upload attempt at a minimum.  You do not want a case where the pod keeps crashing and then uploading data as soon as it starts, then crashes and uploads it again, rapidly.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,245 @@
...
+
+    response = requests.post(url, data=data, timeout=REDHAT_SSO_TIMEOUT)
+    if response.status_code == requests.codes.ok:
+        return response.json()["access_token"]
+    else:
+        # TODO: better error handling
+        raise Exception("Failed to generate refresh token.")
+
+
+def run_oc(args: list[str]) -> subprocess.CompletedProcess:
```
bparees, line 67: running oc commands is ok for tests.  for OLS code we need to be using a proper k8s python client lib.  we don't want to be shipping the oc binary.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,245 @@
...
+    with tarfile.open(fileobj=tarball_io, mode="w:gz") as tar:
+        # arcname parameter is set to a stripped path to avoid including
+        # the full path of the root dir
+        for file_path in file_paths:
+            tar.add(file_path, arcname=file_path.as_posix().replace(path_to_strip, ""))
+
+        # add magic file for identification of our archive on the CCX side
+        empty_file = tarfile.TarInfo("openshift_lightspeed.json")
+        empty_file.size = 0
+        tar.addfile(empty_file)
```
bparees, line 159: is there any size limit on the tar file insights will accept? 

even if there's no limit enforced by insights, I think we should put a constraint on the max size and create/upload multiple files when we hit that size limit (say 10 megs or something)

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,245 @@
...
+
+# available ingress urls are:
+# - https://console.redhat.com/ - production - works with refresh token
+# - https://console.stage.redhat.com/ - staging/dev - doesn't work with refresh token
+INGRESS_URL = os.environ.get("INGRESS_URL", "https://console.redhat.com/")
+INGRESS_ENDPOINT = os.environ.get("INGRESS_ENDPOINT", "api/ingress/v1/upload")
+INGRESS_TIMEOUT = 10  # seconds
+
+# customer portal offline token is generated from https://access.redhat.com/management/api
+CP_OFFLINE_TOKEN = os.environ.get("CP_OFFLINE_TOKEN")
```
bparees, line 33: elsewhere you indicate this is "mainly for debug purposes", if that is 100% true then this is ok as an env var.

but if there's any expectation customers will be setting this, we need to add it to the ols config(as a filepath reference, like our other cred/secret refs) and read it from there.  meaning we'd also need the OLSConfig CRD to include a secret ref and then the ols operator to mount the secret into the upload container in the ols pod.

ols/user_data_collection/data_collector.py
vishnuchalla: Understood. Thank you

examples/openshift-lightspeed-tls.yaml
```diff
@@ -92,7 +92,10 @@ spec:
         - name: tls-certs
           secret:
             secretName: lightspeed-certs
+        - name: ols-user-data
+          emptyDir: {}
       containers:
+      # this example is without the data collection sidecar
```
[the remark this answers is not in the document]
onmete: removed

ols/user_data_collection/data_collector.py
onmete: another story please: https://issues.redhat.com/browse/OLS-473

ols/user_data_collection/data_collector.py
onmete: Customers don't need to touch it. Developers can. This is for testing purposes.

ols/user_data_collection/data_collector.py
onmete: ok, I might need to have another env that disables this for tests though, like `RUN_WITHOUT_INITIAL_WAIT`.

ols/user_data_collection/data_collector.py
onmete: Oh man, I was happy that I finally figured out how to get the cluster id and pull secret via that (yes, had to install oc to the image).
Ok, so back to the beginning...

ols/user_data_collection/data_collector.py
bparees: sure

onmete:
I've reverted from config to the env var for the OLS_USER_DATA_PATH. Bringing config in just complicates it. Mounting unnecessary volumes, updating sys path, ...

tests/e2e/test_api.py
```diff
@@ -822,37 +825,120 @@ def test_transcripts_storing_standalone():
...
+    assert "contains no data, nothing to do..." in logs
 
-    transcript = cluster_utils.get_single_existing_transcript(
-        pod_name, transcripts_path
+    # get the log point for the next check
+    last_log_line = get_last_log_line(container_log)
+
+    # create a new data
+    cluster_utils.create_file(
+        pod_name, OLS_USER_DATA_PATH + "/feedback/test-file.json", '{"testy": "test"}'
```
bparees, line 930: would be better to actually submit a feedback request to OLS to get this generated, any reason not to?

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,303 @@
...
+import pathlib
+import sys
+import tarfile
+import time
+from typing import Any, Callable
+
+import kubernetes
+import requests
+
+OLS_USER_DATA_PATH = os.environ["OLS_USER_DATA_PATH"]
```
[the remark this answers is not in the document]
onmete: When we use config, we also need to mount volumes with credentials, otherwise it will fail as config expects it.

Updating a sys path is not related only to config only, but it is a more general "problem" because the ols itself is not installed (just run via uvicorn). So in every script where something from ols is needed, the sys path needs to be updated with the project root.

The collector needs to know the root of the data dir. Currently, we set the specific paths in the config for feedback/transcripts dirs. Then in collector code we would need to do a bit of checking/ensuring the paths from where data should be collected.
It is a bit inconsistent how data paths are defined in config and what script expects (all data are stored in a single root). We should either set a path for user data in config (instead of specifying for feedback/transcripts) or explicitly use both paths in collector.

Follow-up https://issues.redhat.com/browse/OLS-514.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,303 @@
...
+        raise Exception(f"Failed to generate access token. Response: {response.json()}")
+
+
+def get_cloud_openshift_pull_secret() -> str:
+    """Get the pull secret token from the cluster."""
+    kubernetes.config.load_incluster_config()
+    v1 = kubernetes.client.CoreV1Api()
+
+    # get the pull-secret from openshift-config namespace
+    secret = v1.read_namespaced_secret("pull-secret", "openshift-config")
```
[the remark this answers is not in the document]
onmete: It requires a set of roles as in comment in https://issues.redhat.com/browse/OLS-498
But those were required only when developing/testing against cluster from cluster bot.
It works in CI without any additional roles.
But I can't answer if something would need to be set extra in via our operator at the moment.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,303 @@
...
+    dockerconfigjson = secret.data[".dockerconfigjson"]
+    dockerconfig = json.loads(base64.b64decode(dockerconfigjson).decode("utf-8"))
+
+    return dockerconfig["auths"]["cloud.openshift.com"]["auth"]
```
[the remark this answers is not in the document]
onmete: There is a story for that https://issues.redhat.com/browse/OLS-500

tests/e2e/test_api.py
onmete: I found that unnecessary. With this, the test doesn't depend on any implementation from the ols service itself.
Obviously, the collector is useless without the service, but I like its independence.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,303 @@
...
+OLS_USER_DATA_COLLECTION_INTERVAL = int(
+    os.environ.get("OLS_USER_DATA_COLLECTION_INTERVAL", 2 * 60 * 60)
+)  # 2 hours in seconds
+
+INITIAL_WAIT = 60 * 5  # 5 minutes in seconds
+RUN_WITHOUT_INITIAL_WAIT = bool(
+    os.environ.get("RUN_WITHOUT_INITIAL_WAIT", "false").lower() == "true"
+)
+
+INGRESS_ENV = os.environ.get("INGRESS_ENV", "stage")  # prod/stage
```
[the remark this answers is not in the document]
onmete: I would rather have a stage as default to avoid accidentally sending data/archives to our bucket.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
bparees: that follow up story is good, but this code still needs to handle the situation in the meantime, otherwise we're going to get ugly exception/errors thrown.

tests/e2e/test_api.py
bparees: the benefit from submitting a feedback request is that it makes it a true e2e test, confirming that the collector is able to collect exactly the files (name, type, content, etc) created by the feedback endpoint.

by creating the file manually, you're introducing a potential disconnect between what is being tested(collector can collect a specific file created by the test), and what the system actually does at runtime(collector collects a file created by the feedback endpoint).  It becomes more of an integration test than a true e2e test.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
bparees: > It works in CI without any additional roles.

that may have been because CI was temporarily using traditional clusters and not hypershift clusters.  I guess we'll find out now that we've switched CI back to hypershift clusters.

if we do end needing to create a role, the one you've proposed in the jira ticket is too broad.  We should create a role that only allows reading the specific secret we need access to, not all secrets on the cluster.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
onmete: Right. The operator should ensure the secret is there before the container is deployed, so I've added a temp solution for this.

tests/e2e/test_api.py
onmete: You are correct. Updated.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
bparees: > I've added a temp solution for this.

what solution did you add? i don't see the scenario handled here.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
bparees: i can live with it either way, but it's worth thinking about set to prod when it needs to be (namely the ols operator is going to be ultimately be responsible for setting this to the correct value, and then when we run CI using the operator to install, you'll need a way to control whether the operator sets this to stage or prod)

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
bparees: sounds good, thanks.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
onmete: Latest commits?

ols/user_data_collection/data_collector.py
```diff
@@ -293,6 +293,15 @@ def gather_ols_user_data(data_path: str) -> None:
...
 
 if __name__ == "__main__":
+    # TODO: Temporary solution until OLS-500
+    try:
+        get_cloud_openshift_pull_secret()
+    except KeyError:
+        logger.error(
+            "Cluster pull secret does not contain the cloud.openshift.com auth"
+        )
+        sys.exit(1)
```
bparees: pretty sure this is going to crashloop the container which is not what we want

it needs to run and just do nothing.

ols/user_data_collection/data_collector.py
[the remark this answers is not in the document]
bparees: ah, you handled it elsewhere.  i'd have just expected you to handle the possibility that the value isn't found here in this logic.

among other things it's also possible that at startup the value doesn't exist, but later it is added.  or vice versa.  So better to check every time (and properly handle that the entry might be missing) than to do it once at startup.

not to mention the crashloop issue w/ your current startup checking logic.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+def get_cloud_openshift_pull_secret() -> str:
+    """Get the pull secret token from the cluster."""
+    kubernetes.config.load_incluster_config()
+    v1 = kubernetes.client.CoreV1Api()
+
+    # get the pull-secret from openshift-config namespace
+    secret = v1.read_namespaced_secret("pull-secret", "openshift-config")
+    dockerconfigjson = secret.data[".dockerconfigjson"]
```
bparees, line 116: also possible (if unlikely) that this returns None, so that should be handled explicilty also.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+    return dockerconfig["auths"]["cloud.openshift.com"]["auth"]
+
+
+def get_cluster_id() -> str:
+    """Get the cluster_id from the cluster."""
+    kubernetes.config.load_incluster_config()
+    custom_objects_api = kubernetes.client.CustomObjectsApi()
+
+    # get the cluster version object
+    version_data = custom_objects_api.get_cluster_custom_object(
```
bparees, line 128: error handling for the get?

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+    Returns:
+        BytesIO object representing the tarball.
+    """
+    tarball_io = io.BytesIO()
+
+    with tarfile.open(fileobj=tarball_io, mode="w:gz") as tar:
+        # arcname parameter is set to a stripped path to avoid including
+        # the full path of the root dir
+        for file_path in file_paths:
+            tar.add(file_path, arcname=file_path.as_posix().replace(path_to_strip, ""))
```
bparees, line 173: is there a way to ensure we do not add or follow symlinks?  we should do so as symlinks are one way to exploit a system into gathering files we didn't intend.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    if CP_OFFLINE_TOKEN:
+        logger.debug("using CP offline token to generate refresh token")
+        token = access_token_from_offline_token(CP_OFFLINE_TOKEN)
+        cluster_id = "unknown"
+    else:
+        logger.debug("using cluster pull secret to authenticate")
+        token = get_cloud_openshift_pull_secret()
+        cluster_id = get_cluster_id()
```
bparees, line 236: error handling for both of these in the event we can't get the pull secret or cluster id

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+
+    if response.status_code == 202:
+        request_id = response.json()["request_id"]
+        logger.info(f"data uploaded with request_id: '{request_id}'")
+    else:
+        raise requests.exceptions.HTTPError(
+            f"data upload failed with response: {response.json()}"
+        )
+
+    return response
```
bparees, line 264: simplifying refactor:

```
if statuscode!=202:
  raise exception

log success
return response
```

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+if __name__ == "__main__":
+    # TODO: Temporary solution until OLS-500
+    if INGRESS_ENV == "stage" and not CP_OFFLINE_TOKEN:
+        try:
+            get_cloud_openshift_pull_secret()
+        except KeyError:
+            logger.error(
+                "Cluster pull secret does not contain the cloud.openshift.com auth"
+            )
+            sys.exit(1)
```
bparees, line 304: noted elsewhere but to put it all in one review:  if you do this the sidecar will crashloop which is not what we want.

if we can't get the auth key from the pullsecret we should just log the error and retry on the next regular interval.

1) the key might show up later
2) the key might be removed later

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,313 @@
...
+RUN_WITHOUT_INITIAL_WAIT = bool(
+    os.environ.get("RUN_WITHOUT_INITIAL_WAIT", "false").lower() == "true"
+)
+
+INGRESS_ENV = os.environ.get("INGRESS_ENV", "stage")  # prod/stage
+if INGRESS_ENV not in {"prod", "stage"}:
+    raise ValueError(
+        f"Unknown value in INGRESS_ENV: {INGRESS_ENV}. Allowed: prod, stage"
+    )
+INGRESS_TIMEOUT = 10  # seconds
```
bparees, line 39: seems short.  does this timeout include the time taken to actually upload the data, or just to make an initial connection?

bparees (commented):
looks pretty close to ready

ols/user_data_collection/data_collector.py
onmete: Don't know, but I raised it to 30.

onmete:
@bparees Review comments addressed. I hope I didn't miss anything.

bparees:
/retest

sorry i didn't get to this one today, i promise it'll be first in my queue tomorrow

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,323 @@
...
+    kubernetes.config.load_incluster_config()
+    custom_objects_api = kubernetes.client.CustomObjectsApi()
+
+    # get the cluster version object
+    version_data = custom_objects_api.get_cluster_custom_object(
+        "config.openshift.io", "v1", "clusterversions", "version"
+    )
+    try:
+        return version_data["spec"]["clusterID"]
+    except KeyError:
```
bparees, line 140: version_data itself could be None if the get completely failed(due to permissions errors, api server error, etc), will KeyError handle that?  I don't believe so.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,323 @@
...
+        raise Exception(f"Failed to generate access token. Response: {response.json()}")
+
+
+def get_cloud_openshift_pull_secret() -> str:
+    """Get the pull secret token from the cluster."""
+    kubernetes.config.load_incluster_config()
+    v1 = kubernetes.client.CoreV1Api()
+
+    # get the pull-secret from openshift-config namespace
+    secret = v1.read_namespaced_secret("pull-secret", "openshift-config")
```
bparees, line 119: Have you checked the client code to see what errors/exceptions it can raise?  I would expect that it might return permission denied errors/exceptions, as well as api timeouts or api unavailable errors.

These are all things that should be caught and logged explicitly, not handled as a generic "keyerror" (if catching the KeyError exception would even handle those cases, which i don't think it will since you'd probably get back a None value in those cases).

Otherwise it will be difficult to debug in a customer environment what has gone wrong/what is preventing us from doing the ingress upload.

ols/user_data_collection/data_collector.py
```diff
@@ -0,0 +1,323 @@
...
+
+    # get the cluster version object
+    version_data = custom_objects_api.get_cluster_custom_object(
+        "config.openshift.io", "v1", "clusterversions", "version"
+    )
+    try:
+        return version_data["spec"]["clusterID"]
+    except KeyError:
+        logger.error("failed to get cluster_id from cluster")
+        return "unknown"
```
bparees, line 142: for the pullsecret helper you raise an exception, but for this helper you return "unknown", i think you should be consistent, and i think raising the exception is the right thing to do.  Checking for magic string values is fragile and puts an unreasonable burden on the caller to anticipate that return value.

bparees (commented):
I still have some concerns that this isn't handling all the error conditions that can occur while you're trying to retrieve the clusterid and pullsecret, but you can address them as a followup.  i'll tag this for merge as is for now.

bparees:
/lgtm
/approve

please address the error handling gaps in a follow up.

bparees:
> please address the error handling gaps in a follow up.

and please don't close OLS-136 until they're addressed

PR #585, opened 2024-03-12
OLS-302: Redis commands retry settings

tisnik:
## Description

- Redis commands retry settings
- No changes made in Redis client (yet)

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Related Issue #[OLS-302](https://issues.redhat.com//browse/OLS-302)

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -254,6 +254,9 @@ class RedisConfig(BaseModel):
     max_memory_policy: Optional[str] = None
     password: Optional[str] = None
     ca_cert_path: Optional[str] = None
+    retry_on_error: Optional[bool] = None
+    retry_on_timeout: Optional[bool] = None
+    number_of_retries: Optional[int] = None
 
     def __init__(self, data: Optional[dict] = None) -> None:
         """Initialize configuration and perform basic validation."""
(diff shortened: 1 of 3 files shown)


syedriko:
@tisnik a few questions: 
- how long is the timeout? do we need a variable for it?
- do we have to have (separate) `retry_on_*` flags?  we could set the `number_of_retries` to 0 to disable retries

tisnik:
> @tisnik a few questions:
> 
>     * how long is the timeout? do we need a variable for it?

- timeout can not be specified. But it is based on exponential retry algorithm so for high retries it will be long (tried, can be minutes even)
 
> 
>     * do we have to have (separate) `retry_on_*` flags?  we could set the `number_of_retries` to 0 to disable retries

- we'd possibly like to tune it (+ retry_on_error might be set to different value than retry_on_timeout if needed).
- that said, there are default values specified in constants, so it won't have to specified explicitly if not needed

syedriko:
I'm looking at 
https://redis-py.readthedocs.io/en/stable/connections.html
https://redis-py.readthedocs.io/en/stable/retry.html
https://redis-py.readthedocs.io/en/stable/backoff.html
and I'm seeing a bit of a different picture. There're many different retry policies, retry_on_error is a list of errors, possibly including timeout, retry_on_timeout looks like an abbreviation for retry_on_error=[TimeoutError].
I'm not sure I'm seeing how config maps to this...

tisnik:
> I'm looking at https://redis-py.readthedocs.io/en/stable/connections.html https://redis-py.readthedocs.io/en/stable/retry.html https://redis-py.readthedocs.io/en/stable/backoff.html and I'm seeing a bit of a different picture. There're many different retry policies, retry_on_error is a list of errors, possibly including timeout, retry_on_timeout looks like an abbreviation for retry_on_error=[TimeoutError]. I'm not sure I'm seeing how config maps to this...

yes, the idea is to use `retry` parameter with `ExponentialBackoff` with possibly explicitly setting `base` and `cap` as I wrote.

onmete:
Might be better to do the config change in pair with the actual implementation.

But for what is in this PR
/lgtm

PR #589, opened 2024-03-13
OLS-302: Unit tests: Redis retry settings on Redis client initialization

tisnik:
## Description

Unit tests: Redis retry settings on Redis client initialization

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit tests

- Related Issue #[OLS-302](https://issues.redhat.com//browse/OLS-302)

## Testing
- Use `make test-unit`

### DIFF
--- a/tests/unit/cache/test_redis_cache.py
+++ b/tests/unit/cache/test_redis_cache.py
@@ -144,3 +144,73 @@ def test_initialize_redis_with_tls_certs():
         assert (
             cache.redis_client.kwargs["ssl_ca_certs"] == "test/config/redis_ca_cert.crt"
         )
+
+
+def test_initialize_redis_default_retry_settings():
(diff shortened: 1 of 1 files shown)


onmete (commented):
When `retry_on_error` and `retry_on_timeout` set to `True`/`"true"`, it should be `True`, but every other value ends up as `False` (according to how it is defined in config), right?

So why are there checks against `None`?

tisnik:
> When `retry_on_error` and `retry_on_timeout` set to `True`/`"true"`, it should be `True`, but every other value ends up as `False` (according to how it is defined in config), right?
> 
> So why are there checks against `None`?

retry_on_error is not bool in the end. It's list of exception classes

PR #592, opened 2024-03-13
Explicit check for `None` values (for empty configuration)

tisnik:
## Description

Explicit check for `None` values (for empty configuration)
[context: our model allows None values, need to be fixed soon]

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

### DIFF
--- a/ols/app/main.py
+++ b/ols/app/main.py
@@ -43,8 +43,8 @@ else:
 
 # update provider and model as soon as possible so the metrics will be visible
 # even for first scraping
-metrics.selected_provider.info({"name": config.ols_config.default_provider})
-metrics.selected_model.info({"name": config.ols_config.default_model})
+metrics.selected_provider.info({"name": config.ols_config.default_provider or ""})
+metrics.selected_model.info({"name": config.ols_config.default_model or ""})
 
 
 @app.middleware("")


onmete:
So instead of setting it to `""` here, would be better to enforce provider/model is always set?

syedriko:
> So instead of setting it to `""` here, would be better to enforce provider/model is always set?

That's what I'm thinking, too. The default provider and model are validated here https://github.com/openshift/lightspeed-service/blob/2b5d485e0202faecb5419fe6c3a5d314c67d6d0c/ols/app/models/config.py#L686 and I'm TBH not sure why it's allowed for both of them to be empty.

tisnik:
> So instead of setting it to `""` here, would be better to enforce provider/model is always set?

I hope that the new config/model will take care of it.

tisnik:
> > So instead of setting it to `""` here, would be better to enforce provider/model is always set?
> 
> That's what I'm thinking, too. The default provider and model are validated here
> 
> https://github.com/openshift/lightspeed-service/blob/2b5d485e0202faecb5419fe6c3a5d314c67d6d0c/ols/app/models/config.py#L686
> and I'm TBH not sure why it's allowed for both of them to be empty.

https://github.com/openshift/lightspeed-service/blob/2b5d485e0202faecb5419fe6c3a5d314c67d6d0c/ols/app/models/config.py#L578

can be None, and IMHO *is* None during CI testing on OLS

syedriko:
> > So instead of setting it to `""` here, would be better to enforce provider/model is always set?
> 
> I hope that the new config/model will take care of it.

But the new config is copying the existing business logic. It's starting to come back to me now. That validation code has been there when there were many providers/models for different purposes, like validator, classifier, etc. These special-purpose providers and models were all optional, but it would be inconsistent for them to only specify, say, a validator_provider without a model.

tisnik:
> > > So instead of setting it to `""` here, would be better to enforce provider/model is always set?
> > 
> > 
> > I hope that the new config/model will take care of it.
> 
> But the new config is copying the existing business logic. It's starting to come back to me now. That validation code has been there when there were many providers/models for different purposes, like validator, classifier, etc. These special-purpose providers and models were all optional, but it would be inconsistent for them to only specify, say, a validator_provider without a model.

I get that but even the updated business logic will be "improper" from Pydantic + Mypy point of view, where everything is simply optional (or nullable if you wish).

ols/app/main.py
```diff
@@ -43,8 +43,8 @@
 
 # update provider and model as soon as possible so the metrics will be visible
 # even for first scraping
-metrics.selected_provider.info({"name": config.ols_config.default_provider})
-metrics.selected_model.info({"name": config.ols_config.default_model})
+metrics.selected_provider.info({"name": config.ols_config.default_provider or ""})
+metrics.selected_model.info({"name": config.ols_config.default_model or ""})
```
bparees, line 47: this doesn't seem aligned w/ the PR title.

if the change still makes sense i guess you can pull it into your new metrics PR that's touching this space

PR #595, opened 2024-03-14
OLS-387: Add user data collection config

onmete:
## Description

Add config for ols user data collection.

## Type of change

- [x] New feature

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -53,6 +53,9 @@ ols_config:
     k8s_cluster_api: "https://api.example.com:6443"
     k8s_ca_cert_path: "/Users/home/ca.crt"
     skip_tls_verification: false
+  user_data_collection:
+    feedback_disabled: false
+    feedback_storage: "/app-root/data/feedback"
 dev_config:
   enable_dev_ui: true
   # llm_params:
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -6,7 +6,8 @@ import re
 from typing import Optional
 from urllib.parse import urlparse
 
-from pydantic import BaseModel
+from pydantic import BaseModel, DirectoryPath, model_validator
+from typing_extensions import Self
 
 from ols import constants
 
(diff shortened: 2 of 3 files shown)


ols/app/models/config.py
```diff
@@ -552,6 +553,22 @@ def validate_yaml(self) -> None:
             _dir_check(self.embeddings_model_path, "Embeddings model path")
 
 
+class UserDataCollection(BaseModel):
+    """User data collection configuration."""
+
+    disabled: bool = True
+    storage_location: Optional[DirectoryPath] = None
+
+    @model_validator(mode="after")
```
tisnik, line 562: are we mixing the "old" and "new" approach of pydantic model there?

tisnik (commented):
seems ok, just not sure how to handle models (old/new approach). Now we'd use both which might be confusing

ols/app/models/config.py
onmete: Are you talking about models in general or about this model specifically? This one should be just "new".

tisnik: just about this model

onmete:
The motivation for the new (pydantic/Michal's) approach.
- Michal doesn't need to refactor this in his PR.
- If Michal is not able to finish his PR, whenever there is a need to touch some part of the config, it should be refactored as part of the work.

examples/olsconfig.yaml
```diff
@@ -53,6 +53,9 @@ ols_config:
     k8s_cluster_api: "https://api.example.com:6443"
     k8s_ca_cert_path: "/Users/home/ca.crt"
     skip_tls_verification: false
+  user_data_collection:
```
bparees, line 56: since we might find ourselves adding fields under this section for "transcript" collection also, please prefix these fields to make it clear they apply to feedback collection, not transcript collection

either that or name the top level field to "feedback_collection" or something similar.

examples/olsconfig.yaml
```diff
@@ -53,6 +53,9 @@ ols_config:
...
+    disabled: false
+    storage_location: "/app-root/ols-user-data"
```
bparees, line 58: same comment for the value of the path, make it clear it's feedback data.

ols/app/models/config.py
```diff
@@ -552,6 +553,22 @@ def validate_yaml(self) -> None:
             _dir_check(self.embeddings_model_path, "Embeddings model path")
 
 
+class UserDataCollection(BaseModel):
```
bparees, line 556: rename to align with the config yaml field names when you change them

ols/app/models/config.py
bparees: assuming the answer is "yes", i am fine w/ us starting to introduce pydantic validation in new config logic.  If we think it's better, there's no reason not to start using it where we can, and then go back and refactor the old code later when we don't have more pressing concerns.

bparees:
@onmete OLS-387 also requires an api the UI can query to check if feedback is disabled, are you intending to do that in a separate PR?  just want to make sure the acceptance criteria isn't missed.

onmete:
> [OLS-387](https://issues.redhat.com//browse/OLS-387) also requires an api the UI can query to check if feedback is disabled

I will do that in the feedback PR.

PR #599, opened 2024-03-15
OLS-353: Use keywords for query validation

asamal4:
## Description

Load keywords from a text file.
Validate query based on basic keyword matching.
Config option to use either keyword or llm based query validation.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue # [OLS-353](https://issues.redhat.com//browse/OLS-353)
- Closes #

## Testing

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -18,6 +18,7 @@ from ols.src.query_helpers.docs_summarizer import DocsSummarizer
 from ols.src.query_helpers.question_validator import QuestionValidator
 from ols.utils import config, suid
 from ols.utils.auth_dependency import auth_dependency
+from ols.utils.keywords import KEYWORDS
 
 logger = logging.getLogger(__name__)
 
@@ -206,8 +207,8 @@ def redact_query(conversation_id: str, llm_request: LLMRequest) -> LLMRequest:
         )
 
 
-def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
-    """Validate user question, raise HTTPException in case of any problem."""
+def _validate_question_llm(conversation_id: str, llm_request: LLMRequest) -> str:
+    """Validate user question using llm, raise HTTPException in case of any problem."""
     # Validate the query
     try:
         question_validator = QuestionValidator(
@@ -229,6 +230,43 @@ def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail="Error while validating question",
         )
+
+
+def _validate_question_keyword(query: str) -> str:
+    """Validate user question using keyword."""
+    # Current implementation is without any tokenizer method, lemmatization/n-grams.
+    # Add valid keywords to keywords.txt file.
+    query_temp = query.lower()
+    for kw in KEYWORDS:
+        if kw in query_temp:
+            return constants.SUBJECT_VALID
+    # query_temp = {q_word.lower().strip(".?,") for q_word in query.split()}
+    # common_words = config.keywords.intersection(query_temp)
+    # if len(common_words) > 0:
+    #     return constants.SUBJECT_VALID
+
+    logger.debug(f"No matching keyword found for query: {query}")
+    return constants.SUBJECT_INVALID
+
+
+def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
+    """Validate user question."""
+    match config.ols_config.query_validation_method:
+
+        case constants.QueryValidationMethod.DISABLED:
+            logger.debug(
+                f"{conversation_id} Question validation is disabled. "
+                f"Treating question as valid."
+            )
+            return constants.SUBJECT_VALID
+
+        case constants.QueryValidationMethod.KEYWORD:
+            logger.debug("Keyword based query validation.")
+            return _validate_question_keyword(llm_request.query)
+
+        case _:
+            logger.debug("LLM based query validation.")
+            return _validate_question_llm(conversation_id, llm_request)
 
 
 def generate_bare_response(conversation_id: str, llm_request: LLMRequest) -> str:
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -3,12 +3,25 @@
 # ruff: noqa: E501
 """Constants used in business logic."""
 
+from enum import StrEnum
+
+
+class QueryValidationMethod(StrEnum):
+    """Possible options for query validation method."""
+
+    KEYWORD = "keyword"
+    LLM = "llm"
+    DISABLED = "disabled"
+
+
 SUBJECT_VALID = "SUBJECT_VALID"
 SUBJECT_INVALID = "SUBJECT_INVALID"
 POSSIBLE_QUESTION_VALIDATOR_RESPONSES = (
     SUBJECT_VALID,
     SUBJECT_INVALID,
 )
+
+KEYWORDS_FILE_PATH = "ols/data_assets/keywords.txt"
 
 # Default responses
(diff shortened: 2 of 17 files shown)


keywords.txt
```diff
@@ -0,0 +1,40 @@
+alert
```
tisnik, line 1: do you want to keep this file in top-level?

ols/app/endpoints/ols.py
```diff
@@ -231,6 +231,41 @@ def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
         )
 
 
+def _validate_question_keyword(query: str) -> str:
+    """Validate user question using keyword."""
+    # Current implementation is without any tokenizer method, lemmatization/n-grams.
+    # Add valid keywords to keywords.txt file.
+    for kw in config.keywords:
+        if kw in query.lower():
```
tisnik, line 239: you might call `lower()` before the loop to speedup things a bit (especially for long queries)

ols/app/endpoints/ols.py
```diff
@@ -231,6 +231,41 @@ def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
...
+    # query_temp = {q_word.lower().strip(".?,") for q_word in query.split()}
+    # common_words = config.keywords.intersection(query_temp)
+    # if len(common_words) > 0:
+    #     return constants.SUBJECT_VALID
+
+    logger.info(f"No matching keyword found for query: {query}")
+    return constants.SUBJECT_INVALID
+
+
+def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
```
tisnik, line 250: prob. move into llm/query helpers part of code?

ols/utils/config.py
```diff
@@ -39,6 +41,21 @@ def load_config_from_stream(stream: TextIOBase) -> config_model.Config:
     return c
 
 
+def _init_keyword() -> None:
```
tisnik, line 44: keyword_s_

tisnik (commented):
just couple of nitpicks, but it looks ok in overall. TYVM

keywords.txt
tisnik: well `olsconfig.yaml` is loaded from path specified in `OLS_CONFIG_FILE` (which might be different from root level). I'm just worried about having ton of files in top-level, some of them devel config, some as docs, other as required work files.

tisnik (commented):
it looks ok

ols/app/endpoints/ols.py
```diff
@@ -231,6 +231,42 @@ def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
...
+    """Validate user question."""
+    if config.dev_config.disable_question_validation:
+        logger.debug(
+            f"{conversation_id} Question validation is disabled. "
+            f"Treating question as valid."
+        )
+        return constants.SUBJECT_VALID
+
+    if len(config.keywords) > 0:
```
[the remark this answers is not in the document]
bparees: also given my comment below about the current set of keywords, i think we should keep the default behavior to be using the llm for question validation, until we have more confidence we have a large enough keyword set that we're not improperly rejecting reasonable/valid questions.

bparees (commented):
aside from my other comments: 

you've introduced a number of logger.info statements that should probably be logger.debug statements.

it's a certain amount of art over science in terms of determining which to use, but especially stuff that's going to be printed on every request and isn't very interesting unless you're trying to debug a problem, should not be logged at the info level.

this is something we need to clean up throughout the codebase.  We have too much stuff logged at the info level currently, imho.

We may end choosing to default running OLS w/ logging at the debug level so we have good debug information available if someone runs into an issue, but at least then users who don't like the verbosity would have the option to switch to info level to reduce the log noise.  If we log everything at the info level, they have little recourse to reduce the noise.

ols/app/endpoints/ols.py
```diff
@@ -231,6 +231,42 @@ def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
...
+    # if len(common_words) > 0:
+    #     return constants.SUBJECT_VALID
+
+    logger.info(f"No matching keyword found for query: {query}")
+    return constants.SUBJECT_INVALID
+
+
+def validate_question(conversation_id: str, llm_request: LLMRequest) -> str:
+    """Validate user question."""
+    if config.dev_config.disable_question_validation:
```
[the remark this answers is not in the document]
bparees: 1) i'm not sure we've made a decision on whether we'd allow an admin to disable validation.  i'm not sure there's a reason we wouldn't let them, if they choose to do it, though of course it should not be our default mode of operation
2) admins don't control the config.yaml directly, they have to edit the OLS CR to set configuration, so we don't have to allow "disabled" as a valid enum on the CRD, while we can still allow it in the config yaml.  (yes technically an admin could bypass the operator and modify the config directly, but then they could also set the dev_config flags via that mechanism.  At that point they are bypassing our official apis)

moving the "disable validation" mechanism into your new field keeps our config simpler/easier to reason about, rather than having 2 fields that contribute to the overall validation behavior.

tests/unit/app/endpoints/test_ols.py
```diff
@@ -156,9 +156,33 @@ def test_store_conversation_history_improper_conversation_id(load_config):
...
+
+    assert resp == constants.SUBJECT_VALID
+    assert llm_validate_question_mock.call_count == 0
+
+
+def test_validate_question_invalid_kw(load_config):
+    """Check the behaviour of validate_question function using invalid keyword."""
+    config.keywords = {"some_text"}
+    conversation_id = suid.get_suid()
+    query = "Tell me about Kubernetes"
```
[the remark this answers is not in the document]
bparees: oh, you're overriding the keywords so it's not the default set.  got it.  but yeah, thanks in advance for updating it to something that will be more obvious.

tests/unit/app/endpoints/test_ols.py
```diff
@@ -194,6 +218,21 @@ def test_validate_question_on_validation_error(validate_question_mock, load_conf
         ols.validate_question(conversation_id, llm_request)
 
 
+@patch("ols.app.endpoints.ols.config.dev_config.disable_question_validation", True)
+@patch("ols.app.endpoints.ols._validate_question_keyword")
+@patch("ols.src.query_helpers.question_validator.QuestionValidator.validate_question")
+def test_validate_question_disabled(validate_question_mock, validate_question_kw_mock):
+    """Check the behaviour of validate_question function when it is disabled."""
+    conversation_id = suid.get_suid()
+    query = "Tell me about Kubernetes"
```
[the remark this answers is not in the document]
bparees: yeah, it doesn't really confirm that validation is disabled if the string you're sending would have passed validation even if it was on

ols/data_assets/keywords.txt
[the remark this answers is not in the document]
bparees: ok. let's make sure we leave the default validation method as llm-based until this is expanded

ols/constants.py
```diff
@@ -3,13 +3,26 @@
...
+
+
 SUBJECT_VALID = "SUBJECT_VALID"
 SUBJECT_INVALID = "SUBJECT_INVALID"
 POSSIBLE_QUESTION_VALIDATOR_RESPONSES = (
     SUBJECT_VALID,
     SUBJECT_INVALID,
 )
 
+KEYWORDS_FILE_PATH = "ols/data_assets/keywords.txt"
```
bparees, line 24: not used anymore?

bparees (commented):
one nit you can cleanup as a follow up, i'll merge this for now.

PR #634, opened 2024-03-22
OLS-440: Postgres-based cache

tisnik:
## Description

Postgres-based cache

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -313,7 +313,7 @@ class PostgresConfig(BaseModel):
     user: str = constants.POSTGRES_CACHE_USER
     password_path: Optional[str] = None
     password: Optional[str] = None
-    require_ssl: bool = False
+    ssl_mode: str = constants.POSTGRES_CACHE_SSL_MODE
     ca_cert_path: Optional[str] = None
     max_entries: PositiveInt = constants.POSTGRES_CACHE_MAX_ENTRIES
 
@@ -542,6 +542,7 @@ class ConversationCacheConfig(BaseModel):
                 self.type == other.type
                 and self.redis == other.redis
                 and self.memory == other.memory
+                and self.postgres == other.postgres
             )
         return False
 
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -102,6 +102,10 @@ POSTGRES_CACHE_DBNAME = "cache"
 POSTGRES_CACHE_USER = "postgres"
 POSTGRES_CACHE_MAX_ENTRIES = 1000
 
+# look at https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
+# for all possible options
+POSTGRES_CACHE_SSL_MODE = "prefer"
+
 
 # default indentity for local testing and deployment
 DEFAULT_USER_UID = "c1143120-551e-4a47-ad47-2748d6f3c81c"
--- a/ols/src/cache/postgres_cache.py
+++ b/ols/src/cache/postgres_cache.py
@@ -1,18 +1,109 @@
 """Cache that uses Postgres to store cached values."""
 
-from typing import Optional
+import logging
+import pickle
+from typing import Any, Optional
 
+import psycopg2
 from langchain_core.messages.base import BaseMessage
 
 from ols.app.models.config import PostgresConfig
 from ols.src.cache.cache import Cache
+from ols.src.cache.cache_error import CacheError
+
+logger = logging.getLogger(__name__)
 
 
 class PostgresCache(Cache):
-    """Cache that uses Postgres to store cached values."""
+    """Cache that uses Postgres to store cached values.
+
+    The cache itself is stored in following table:
+
+    ```
+         Column      |            Type             | Nullable | Default | Storage  |
+    -----------------+-----------------------------+----------+---------+----------+
+     user_id         | text                        | not null |         | extended |
+     conversation_id | text                        | not null |         | extended |
+     value           | bytea                       |          |         | extended |
+     updated_at      | timestamp without time zone |          |         | plain    |
+    Indexes:
+        "cache_pkey" PRIMARY KEY, btree (user_id, conversation_id)
+        "cache_key_key" UNIQUE CONSTRAINT, btree (key)
+        "timestamps" btree (updated_at)
+    Access method: heap
+    ```
+    """
+
+    CREATE_CACHE_TABLE = """
+        CREATE TABLE IF NOT EXISTS cache (
+            user_id         text NOT NULL,
+            conversation_id text NOT NULL,
+            value           bytea,
+            updated_at      timestamp,
+            PRIMARY KEY(user_id, conversation_id)
+        );
+        """
+
+    CREATE_INDEX = """
+        CREATE INDEX IF NOT EXISTS timestamps
+            ON cache (updated_at)
+        """
+
+    SELECT_CONVERSATION_HISTORY_STATEMENT = """
+        SELECT value
+          FROM cache
+         WHERE user_id=%s AND conversation_id=%s LIMIT 1
+        """
+
+    UPDATE_CONVERSATION_HISTORY_STATEMENT = """
+        UPDATE cache
+           SET value=%s, updated_at=CURRENT_TIMESTAMP
+         WHERE user_id=%s AND conversation_id=%s
+        """
+
+    INSERT_CONVERSATION_HISTORY_STATEMENT = """
+        INSERT INTO cache(user_id, conversation_id, value, updated_at)
+        VALUES (%s, %s, %s, CURRENT_TIMESTAMP)
+        """
+
+    DELETE_CONVERSATION_HISTORY_STATEMENT = """
+        DELETE FROM cache
+         WHERE (user_id, conversation_id) in
+               (SELECT user_id, conversation_id FROM cache ORDER BY updated_at LIMIT
+        """
+
+    QUERY_CACHE_SIZE = """
+        SELECT count(*) FROM cache;
+        """
 
     def __init__(self, config: PostgresConfig) -> None:
         """Create a new instance of Postgres cache."""
+        # initialize connection to DB
+        self.conn = psycopg2.connect(
+            host=config.host,
+            port=config.port,
+            user=config.user,
+            password=config.password,
+            dbname=config.dbname,
+            sslmode=config.ssl_mode,
+            sslrootcert=config.ca_cert_path,
+        )
+        self.conn.autocommit = True
+        try:
+            self.initialize_cache()
+        except Exception as e:
+            self.conn.close()
+            logger.exception(f"Error initializing Postgres cache:\n{e}")
+            raise e
+        self.capacity = config.max_entries
+
+    def initialize_cache(self) -> None:
+        """Initialize cache - clean it up etc."""
+        cur = self.conn.cursor()
+        cur.execute(PostgresCache.CREATE_CACHE_TABLE)
+        cur.execute(PostgresCache.CREATE_INDEX)
+        cur.close()
+        self.conn.commit()
 
     def get(self, user_id: str, conversation_id: str) -> Optional[list[BaseMessage]]:
         """Get the value associated with the given key.
@@ -24,7 +115,13 @@ class PostgresCache(Cache):
         Returns:
             The value associated with the key, or None if not found.
         """
-        return None
+        with self.conn.cursor() as cursor:
+            try:
+                return PostgresCache._select(cursor, user_id, conversation_id)
+            except psycopg2.DatabaseError as e:
+                logger.error(f"PostgresCache.get {e}")
+                raise CacheError("PostgresCache.get", e)
+            return None
 
     def insert_or_append(
         self, user_id: str, conversation_id: str, value: list[BaseMessage]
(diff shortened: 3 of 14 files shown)


ols/src/cache/postgres_cache.py
```diff
@@ -0,0 +1,142 @@
...
+
+class PostgresCache(Cache):
+    """Cache that uses Postgres to store cached values."""
+
+    CREATE_CACHE_TABLE = """
+        CREATE TABLE IF NOT EXISTS cache (
+            user_id         text NOT NULL,
+            conversation_id text NOT NULL,
+            key             text UNIQUE NOT NULL,
+            value           bytea,
```
onmete, line 19: `text` is not a good idea?

ols/src/cache/postgres_cache.py
```diff
@@ -0,0 +1,142 @@
...
+
+    INSERT_STATEMENT = """
+        INSERT INTO cache(user_id, conversation_id, value, updated_at)
+        VALUES (%s, %s, %s, CURRENT_TIMESTAMP)
+        """
+
+    DELETE_STATEMENT = """
+        DELETE FROM cache
+         WHERE (user_id, conversation_id) in
+               (SELECT user_id, conversation_id FROM cache ORDER BY updated_at LIMIT
```
onmete, line 50: The `LIMIT` value is missing?

ols/src/cache/postgres_cache.py
```diff
@@ -0,0 +1,142 @@
...
+
+    CREATE_INDEX = """
+        CREATE INDEX IF NOT EXISTS timestamps
+            ON cache (updated_at)
+        """
+
+    SELECT_STATEMENT = """
+        SELECT value
+          FROM cache
+         WHERE user_id=%s and conversation_id=%s limit 1
```
onmete, line 33: Upper `LIMIT`?

ols/src/cache/postgres_cache.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+
+    def __init__(self) -> None:
+        """Create a new instance of the `PostgresCache` class."""
+        import psycopg2
+
+        # initialize connection to DB
+        connection_string = "postgresql://localhost:5432"  # TODO from config
+        self.conn = psycopg2.connect(connection_string)
+        self.initialize_cache()
```
onmete, line 60: Is it a good practice to execute in init?

ols/src/cache/postgres_cache.py
onmete: Ah I see we are storing `BaseMessage`, not text.

tisnik: yup, it's serialized object basically

ols/src/cache/postgres_cache.py
tisnik: WDYM please?

onmete: ` WHERE user_id=%s and conversation_id=%s LIMIT 1` - just make the `limit` word upper

ols/src/cache/postgres_cache.py
onmete: And maybe the closing `)` too?

ols/src/cache/postgres_cache.py
tisnik: well why not? there are two outcomes:
1) the object is constructed correctly with all bits and pieces
2) it throws exception

nothing in-between

ols/src/cache/postgres_cache.py
tisnik: well not, it seems it can't be parametrized, so the string if formatted to add it

tisnik:
> I am not sure why we are choosing a RDMS for cache (i.e seems like an anit-pattern for me CMIIW), which performs row-by-row scan in order to fetch a specific key's value. Ideally I think we should use a key-value store that provides us with some inbuilt options to configure cache specific things like
> 
>     * Eviction Policy
> 
>     * TTL (time to live)
> 
>     * maxMemory and more
> 
> 
> also performs a search for key with almost O(1) time complexity, which postgres by performing a row-by-row scan can never achieve. Also those cache specific things that I have mentioned above needs to be implemented explicitly in postgres, instead I would suggest us to choose a better alternative: https://www.linuxlinks.com/cachingsystems/
> 
> Also with postgresql, things are going to get more complicated from a distributed offering perspective too.

We need to choose any tech we have images for. + the ability to use eviction mechanism if needed

ols/src/cache/postgres_cache.py
onmete: I see

ols/src/cache/postgres_cache.py
tisnik: oh got it

ols/src/cache/postgres_cache.py
onmete: Just thinking if this should happen when the object is initialized.
But I guess it is alright in this case. Otherwise, there would need to be some ensure it happens when the cache is actually used.

Btw should it print? Not close connection, log and raise?

tisnik: correct, it should close+log+re-raise (will do)

ols/src/cache/postgres_cache.py
tisnik: fixed

ols/src/cache/postgres_cache.py
tisnik: fixed

vishnuchalla:
> > I am not sure why we are choosing a RDMS for cache (i.e seems like an anit-pattern for me CMIIW), which performs row-by-row scan in order to fetch a specific key's value. Ideally I think we should use a key-value store that provides us with some inbuilt options to configure cache specific things like
> > ```
> > * Eviction Policy
> > 
> > * TTL (time to live)
> > 
> > * maxMemory and more
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > also performs a search for key with almost O(1) time complexity, which postgres by performing a row-by-row scan can never achieve. Also those cache specific things that I have mentioned above needs to be implemented explicitly in postgres, instead I would suggest us to choose a better alternative: https://www.linuxlinks.com/cachingsystems/
> > Also with postgresql, things are going to get more complicated from a distributed offering perspective too.
> 
> We need to choose any tech we have images for. + the ability to use eviction mechanism if needed
> 
> WDYM by "which performs row-by-row scan in order to fetch a specific key's value."?
> 
> There are some our observations about possible solutions: https://github.com/tisnik/persistent-lru-caches

In the select statement [here](https://github.com/openshift/lightspeed-service/pull/634/files#diff-903241cee0acd02b447d875b010f07e156a39c7b5aa0df45700759da97ede9c5R91) it simply scans the table row by row until it finds the required key's value. Initially the table might be small, but when the number of entries go on we will start seeing larger times to get responded to our queries. So I wasn't much inclined towards using RDBMS here. Instead a key-value store would help us get the value instantly.

tisnik:
> > > I am not sure why we are choosing a RDMS for cache (i.e seems like an anit-pattern for me CMIIW), which performs row-by-row scan in order to fetch a specific key's value. Ideally I think we should use a key-value store that provides us with some inbuilt options to configure cache specific things like
> > > ```
> > > * Eviction Policy
> > > 
> > > * TTL (time to live)
> > > 
> > > * maxMemory and more
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > also performs a search for key with almost O(1) time complexity, which postgres by performing a row-by-row scan can never achieve. Also those cache specific things that I have mentioned above needs to be implemented explicitly in postgres, instead I would suggest us to choose a better alternative: https://www.linuxlinks.com/cachingsystems/
> > > Also with postgresql, things are going to get more complicated from a distributed offering perspective too.
> > 
> > 
> > We need to choose any tech we have images for. + the ability to use eviction mechanism if needed
> > WDYM by "which performs row-by-row scan in order to fetch a specific key's value."?
> > There are some our observations about possible solutions: https://github.com/tisnik/persistent-lru-caches
> 
> In the select statement [here](https://github.com/openshift/lightspeed-service/pull/634/files#diff-903241cee0acd02b447d875b010f07e156a39c7b5aa0df45700759da97ede9c5R91) it simply scans the table row by row until it finds the required key's value. Initially the table might be small, but when the number of entries go on we will start seeing larger times to get responded to our queries. So I wasn't much inclined towards using RDBMS here. Instead a key-value store would help us get the value instantly.

hmm does it? user_id+conversation_id is compound key. Do you think it is not indexed?

vishnuchalla:
> > > > I am not sure why we are choosing a RDMS for cache (i.e seems like an anit-pattern for me CMIIW), which performs row-by-row scan in order to fetch a specific key's value. Ideally I think we should use a key-value store that provides us with some inbuilt options to configure cache specific things like
> > > > ```
> > > > * Eviction Policy
> > > > 
> > > > * TTL (time to live)
> > > > 
> > > > * maxMemory and more
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > also performs a search for key with almost O(1) time complexity, which postgres by performing a row-by-row scan can never achieve. Also those cache specific things that I have mentioned above needs to be implemented explicitly in postgres, instead I would suggest us to choose a better alternative: https://www.linuxlinks.com/cachingsystems/
> > > > Also with postgresql, things are going to get more complicated from a distributed offering perspective too.
> > > 
> > > 
> > > We need to choose any tech we have images for. + the ability to use eviction mechanism if needed
> > > WDYM by "which performs row-by-row scan in order to fetch a specific key's value."?
> > > There are some our observations about possible solutions: https://github.com/tisnik/persistent-lru-caches
> > 
> > 
> > In the select statement [here](https://github.com/openshift/lightspeed-service/pull/634/files#diff-903241cee0acd02b447d875b010f07e156a39c7b5aa0df45700759da97ede9c5R91) it simply scans the table row by row until it finds the required key's value. Initially the table might be small, but when the number of entries go on we will start seeing larger times to get responded to our queries. So I wasn't much inclined towards using RDBMS here. Instead a key-value store would help us get the value instantly.
> 
> hmm does it? user_id+conversation_id is compound key. Do you think it is not indexed?

As per our offline discussion: https://redhat-internal.slack.com/archives/C068JAU4Y0P/p1711388024867919?thread_ts=1711386445.100049&cid=C068JAU4Y0P looks good to me.

ols/src/cache/postgres_cache.py
```diff
@@ -40,3 +110,38 @@ def insert_or_append(
...
+        cur.close()
+
+    def _insert(self, user_id: str, conversation_id: str, value: str) -> None:
+        cur = self.conn.cursor()
+        cur.execute(PostgresCache.INSERT_STATEMENT, (user_id, conversation_id, value))
+        cur.close()
+
+    def _cleanup(self) -> None:
+        cur = self.conn.cursor()
+        cur.execute("SELECT count(*) FROM cache;")
```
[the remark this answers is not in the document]
vishnuchalla: Never mind. Looks like it turned out to be a myth.
* https://www.linkedin.com/pulse/never-use-count-your-sql-count1-instead-myth-michael-papadopoulos
* https://asktom.oracle.com/ords/asktom.search?tag=select-count1-how-it-works

tests/unit/cache/test_postgres_cache.py
bparees, line 1: test to ensure cleanup is happening/size is constrained?

tisnik: not ATM. I'm thinking about having integration tests with real Postgres, WDYT? Is it worth to do it?

bparees: an integration test would have to stand up postgres itself, right?  i think if we're going that far i'd prefer to go all the way to an e2e test, though how exactly to orchestrate that might be a bit tricky (orchestrating direct access to postgres so you can confirm the content of the tables/rows).  But it should be possible.

tisnik: should I do it in the same PR?

bparees: we certainly need some sort of e2e test that proves OLS can cache conversations in postgres and retrieve them, as part of this PR

ols/src/cache/postgres_cache.py
```diff
@@ -1,18 +1,109 @@
...
+        """
+
+    QUERY_CACHE_SIZE = """
+        SELECT count(*) FROM cache;
+        """
 
     def __init__(self, config: PostgresConfig):
         """Create a new instance of Postgres cache."""
+        # initialize connection to DB
+        self.conn = psycopg2.connect(
```
bparees, line 81: do we need to do anything at the client end to enable TLS?  I am assuming that as long as we run postgres w/ TLS enabled, and the cert is signed by the openshift cluster, then this postgres client will have access (via the pod its running in) to the appropriate CA to validate the postgres cert?

ols/src/cache/postgres_cache.py
```diff
@@ -24,7 +115,8 @@ def get(self, user_id: str, conversation_id: str) -> Optional[list[BaseMessage]]
         Returns:
             The value associated with the key, or None if not found.
         """
-        return None
+        with self.conn.cursor() as cursor:
+            return PostgresCache._select(cursor, user_id, conversation_id)
```
bparees, line 119: error handling?  the user of this postgres cache wrapper should be protected from seeing postgres specific exceptions (the message in the error can indicate what happened at a postgres level, but it shouldn't be a postgres exception)

ols/src/cache/postgres_cache.py
```diff
@@ -35,8 +127,85 @@ def insert_or_append(
...
+                    pickle.dumps(old_value, protocol=pickle.HIGHEST_PROTOCOL),
+                )
+            else:
+                PostgresCache._insert(
+                    cursor,
+                    user_id,
+                    conversation_id,
+                    pickle.dumps(value, protocol=pickle.HIGHEST_PROTOCOL),
+                )
+                PostgresCache._cleanup(cursor, self.capacity)
```
bparees, line 149: same comment here about error/exception handling.

i'd expect postgres_cache to abstract away the postgres specific errors/exceptions, other than including them in the error message, so that a client can expect to handle the same kinds of errors/exceptions regardless of which cache implementation they have chosen to use.

ols/src/cache/postgres_cache.py
tisnik: ok that makes sense, thanks

ols/src/cache/postgres_cache.py
tisnik: updated

ols/src/cache/postgres_cache.py
tisnik: yes it would be needed, updated accordingly

ols/constants.py
```diff
@@ -102,6 +102,10 @@ class QueryValidationMethod(StrEnum):
 POSTGRES_CACHE_USER = "postgres"
 POSTGRES_CACHE_MAX_ENTRIES = 1000
 
+# look at https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
+# for all possible options
+POSTGRES_CACHE_SSL_MODE = "allow"
```
bparees, line 107: should be at least "prefer"

ols/app/models/config.py
```diff
@@ -314,6 +314,7 @@ class PostgresConfig(BaseModel):
     password_path: Optional[str] = None
     password: Optional[str] = None
     require_ssl: bool = False
+    ssl_mode: str = constants.POSTGRES_CACHE_SSL_MODE
```
bparees, line 317: "require_ssl" and "ssl_mode" seem redundant, isn't one of the possible ssl modes "require"?

ols/src/cache/postgres_cache.py
```diff
@@ -1,18 +1,109 @@
...
         """Create a new instance of Postgres cache."""
+        # initialize connection to DB
+        self.conn = psycopg2.connect(
+            host=config.host,
+            port=config.port,
+            user=config.user,
+            password=config.password,
+            dbname=config.dbname,
+            sslmode=config.ssl_mode,
+            sslcert=config.ca_cert_path,
```
vishnuchalla, line 89: This param needs to be `sslrootcert=`. I have verified it with operator code.

ols/app/models/config.py
tisnik: yup, it was leftover, updated now

ols/constants.py
tisnik: updated accordingly

ols/src/cache/postgres_cache.py
tisnik: updated too

ols/constants.py
tisnik: updated

bparees:
needs a rebase, but i've got no further comments on it as it stands

/approve

PR #635, opened 2024-03-22
OLS-244: System prompt change

tisnik:
## Description

System prompt change

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


ols/src/prompts/prompt_generator.py
```diff
@@ -16,13 +16,13 @@
...
-    "PromptConfiguration", "provider model rag_exists history_exists"
+    "PromptConfiguration", "provider model"
 )
 
 # TODO: this might be expanded in the future.
 prompt_configurations = {
-    PromptConfiguration(PROVIDER_BAM, GRANITE_13B_CHAT_V1, True, True): CHAT_PROMPT,
-    PromptConfiguration(PROVIDER_OPENAI, GPT35_TURBO, True, True): CHAT_PROMPT,
+    PromptConfiguration(PROVIDER_BAM, GRANITE_13B_CHAT_V1): CHAT_PROMPT,
+    PromptConfiguration(PROVIDER_OPENAI, GPT35_TURBO): CHAT_PROMPT,
```
bparees, line 25: can you please expand this to our full set of "tier 1" provider/model pairs?

watsonx/granite13b-chatv2  
azure-openai/gpt35

also we shouldn't be using chat v1, should be chat v2.

see: 
https://github.com/openshift/lightspeed-service/blob/578a030ed1eecfbb193077dfb8db91e695265d38/tests/scripts/test-e2e-cluster.sh#L45-L54

ols/src/prompts/prompts.py
```diff
@@ -10,12 +10,16 @@
...
-Use the previous chat history to interact and help the user.
-Use the following pieces of retrieved context to answer the question.
 
+"""
+
+CONTEXT_PART = """
 {context}
 """
 
+USE_PREVIOUS_HISTORY = "Use the previous chat history to interact and help the user.\n"
```
bparees, line 20: shouldn't we drop a parameter in here so we can then parameterize the chat history into the right spot in the prompt instead of appending it at the end?

ols/src/prompts/prompts.py
```diff
@@ -10,12 +10,16 @@
...
+USE_RETRIEVED_CONTEXT = "Use the following pieces of retrieved context to answer the question.\n"
```
bparees, line 21: same comment here about using a parameter for the context, and you can drop the "CONTEXT_PART" variable.

ols/src/prompts/prompts.py
bparees: i guess chat history is handled via the MessagesPlaceholder item so you can ignore this comment

ols/src/prompts/prompt_generator.py
tisnik: added

ols/src/prompts/prompts.py
```diff
@@ -10,12 +10,18 @@
...
 
+"""
+
+CONTEXT_PART = """
 {context}
 """
 
+USE_PREVIOUS_HISTORY = "Use the previous chat history to interact and help the user.\n"
+USE_RETRIEVED_CONTEXT = (
+    "Use the following pieces of retrieved context to answer the question.\n"
```
bparees, line 22: my comment got obsoleted by your push, but:

can we not add the {context} parameter to this string and delete the `CONTEXT_PART` constant?

tisnik:
/assign @bparees 

If you agree with directions, I'll add more unit tests for all possible combinations (with RAG, without RAG, with history, without history).

ols/src/prompts/prompts.py
```diff
@@ -10,15 +10,20 @@
...
+"""
+
+USE_PREVIOUS_HISTORY = "Use the previous chat history to interact and help the user.\n"
+USE_RETRIEVED_CONTEXT = """
 Use the following pieces of retrieved context to answer the question.
 
 {context}
 """
 
 CHAT_PROMPT = ChatPromptTemplate.from_messages(
```
bparees, line 22: should this ever be used any more?  shouldn't we always have a dynamically generated prompt?

ols/src/prompts/prompt_generator.py
```diff
@@ -34,8 +45,24 @@ def prompt_for_configuration(
...
+    messages = []
+    messages.append(SystemMessagePromptTemplate.from_template(system_prompt))
+    if history_exists:
+        messages.append(MessagesPlaceholder(variable_name="chat_history"))
+    messages.append(HumanMessagePromptTemplate.from_template("{query}"))
+
+    return ChatPromptTemplate.from_messages(messages)
 
 
 def generate_prompt(
```
bparees, line 68: a number of the args to generate_prompt do not appear to be used, e.g. `query`, `conversation_id`, `referenced_documents`, and `model_options`

is there some future plan for their use or am i missing something?

tisnik: yes, we talked about this with Asutosh, these are there for usage in the future.

ols/src/prompts/prompt_generator.py
```diff
@@ -34,8 +45,24 @@ def prompt_for_configuration(
...
-    return prompt_configurations.get(selector, default_prompt)
+    # construct system prompt first
+    selector = PromptConfiguration(provider, model)
+    system_prompt = prompt_configurations.get(selector, default_prompt)
+
+    # add optional parts into system prompt
+    if history_exists:
+        system_prompt += USE_PREVIOUS_HISTORY
+    if rag_exists:
+        system_prompt += USE_RETRIEVED_CONTEXT
```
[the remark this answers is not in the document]
tisnik: ok, that actually makes sense

ols/src/prompts/prompts.py
tisnik: It's used in `DocsSummarizer._get_rag_data`. Seems like chicken-egg problem there, as we want to generate prompt, but in order to do it, we need RAG context taken using (generated?) prompt. @asamal4 WDYT?

bparees: so the point of https://github.com/openshift/lightspeed-service/blob/71725e35ca6ed7ef018000d68de30f9a1cba59aa/ols/src/query_helpers/docs_summarizer.py#L80 is to estimate what the overall prompt length is going to be, prior to including any rag content, so we can decide how much space we can dedicate to including rag content.

Note that because our priority of inclusion is:
1) rag content
2) conversation history

the interim computation assumes zero conversation history.

for purposes of the computation, we should still be calling the function to generate the dynamic prompt template, and then formatting that template.  When calling the generate function you can, as a workaround, pass a single character of rag context to ensure the generated prompt includes all the RAG context related system instructions in the prompt, and as result we'll basically get a realistic representation of what the prompt is going to be.  You don't actually need the rag context, you just need a prompt template that assumes there *will be* rag context.

does that make sense?

ols/src/prompts/prompt_generator.py
bparees: i don't love it, but if that usage is coming within the next sprint i guess it's ok.  If it's potentially further out than that, please remove them until we have an actual need for them.

ols/src/prompts/prompt_generator.py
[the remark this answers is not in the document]
bparees: > ok, that actually makes sense

good. it's rare, but it happens on occasion :)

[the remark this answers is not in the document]
tisnik: swapped

ols/src/prompts/prompts.py
tisnik: updated

ols/src/prompts/prompt_generator.py
tisnik: removed unused params we know won't be used soon

ols/src/prompts/prompt_generator.py
```diff
@@ -34,8 +45,24 @@ def prompt_for_configuration(
...
+
+    # construct chat prompt from sequence of messages
+    messages = []
+    messages.append(SystemMessagePromptTemplate.from_template(system_prompt))
+    if history_exists:
+        messages.append(MessagesPlaceholder(variable_name="chat_history"))
```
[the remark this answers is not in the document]
tisnik: ok, so it will be second return value if I understand it correctly @asamal4 ?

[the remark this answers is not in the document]
tisnik: it has been added @asamal4

tisnik:
@bparees as @asamal4 is probably N/A right now, can you please look at this one? I hope all your comments have been figured out + implemented.

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -77,7 +82,20 @@ def _get_rag_data(self, rag_index: VectorStoreIndex, query: str) -> list[dict]:
...
+
+        rag_context = ""
+        prompt, _ = generate_prompt(
+            self.provider,
+            self.model,
+            model_options,
+            query,
+            [],
+            rag_context,
+        )
```
bparees, line 94: isn't this going to generate a prompt that doesn't have the rag context stanza, since you're passing an empty string as the rag context?

tisnik: it should be non-empty string, typo, fixed

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -136,20 +154,29 @@ def summarize(
...
-
-        model_options = self._get_model_options(provider_config)
-
-        interim_prompt = CHAT_PROMPT.format(
+        prompt, _ = generate_prompt(
+            self.provider,
+            self.model,
+            model_options,
+            query,
+            [],
```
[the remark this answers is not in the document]
tisnik: added

PR #637, opened 2024-03-24
OLS-417: add support for authorization groups and fine-grained permissions

nemerna:
Refactor Authorization to include the following 2 features:
1) include Groups in the SAR Request:
this will allow roles assigned to groups that a user belongs to , to take effect in addition to the user directly assigned roles

2) fine-grained permissions support:
this will allow us to customize the dependency for each endpoint separately, where we now require diffrent permissions to access Metrics compared to the regular OLS endpoints

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/app/endpoints/feedback.py
+++ b/ols/app/endpoints/feedback.py
@@ -14,11 +14,12 @@ from ols.app.models.models import (
     StatusResponse,
 )
 from ols.utils import config
-from ols.utils.auth_dependency import auth_dependency
+from ols.utils.auth_dependency import AuthDependency
 from ols.utils.suid import get_suid
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/feedback", tags=["feedback"])
+auth_dependency = AuthDependency(virtual_path="/ols-access")
 
 
 async def ensure_feedback_enabled(request: Request) -> None:
--- a/ols/app/metrics/metrics.py
+++ b/ols/app/metrics/metrics.py
@@ -13,9 +13,10 @@ from prometheus_client import (
 )
 
 import ols.app.models.config as config_model
-from ols.utils.auth_dependency import auth_dependency
+from ols.utils.auth_dependency import AuthDependency
 
 router = APIRouter(tags=["metrics"])
+auth_dependency = AuthDependency(virtual_path="/ols-metrics-access")
 
 rest_api_calls_total = Counter(
     "rest_api_calls_total", "REST API calls counter", ["path", "status_code"]
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -202,3 +202,5 @@ DEFAULT_RESPONSE_TOKEN_LIMIT = 500
 # default indentity for local testing and deployment
 DEFAULT_USER_UID = "c1143120-551e-4a47-ad47-2748d6f3c81c"
 DEFAULT_USER_NAME = "OLS"
+# TO-DO: make this UUID dynamic per cluster (dynamic uuid for each cluster)
+DEFAULT_KUBEADMIN_UID = "b6553200-0f7b-4c82-b1c5-9303ff18e5f0"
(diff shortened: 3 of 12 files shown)


nemerna:
@bparees Added the static UUID for the kubeadmin user
https://issues.redhat.com/browse/OLS-432

examples/ols-metrics-resources.yaml
bparees, line 1: please just add these to the existing `openshift-lightspeed-tls.yaml` example yaml, i want to keep it to a single file to be applied to create all the resources for people using that install path

ols/app/endpoints/feedback.py
```diff
@@ -6,10 +6,11 @@
 from fastapi import APIRouter, Depends
 
 from ols.app.models.models import FeedbackRequest, FeedbackResponse
-from ols.utils.auth_dependency import auth_dependency
+from ols.utils.auth_dependency import AuthDependency
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/feedback", tags=["feedback"])
+auth_dependency = AuthDependency(virtual_path="/ols-metrics-access")
```
bparees, line 13: should be /ols-access since i think we've said ols access should be the same as feedback access (the case where it's not is listing feedback and deleting feedback, but since we expect to drop those apis entirely, we don't need to concern ourselves with them)

(definitely shouldn't be /ols-metrics-access regardless, guessing cut+paste error)

ols/app/metrics/metrics.py
```diff
@@ -11,9 +11,10 @@
     generate_latest,
 )
 
-from ols.utils.auth_dependency import auth_dependency
+from ols.utils.auth_dependency import AuthDependency
 
 router = APIRouter(tags=["metrics"])
+auth_dependency = AuthDependency(virtual_path="/ols-access")
```
bparees, line 17: should be /ols-metrics-access

ols/constants.py
```diff
@@ -207,3 +207,4 @@ class QueryValidationMethod(StrEnum):
 # default indentity for local testing and deployment
 DEFAULT_USER_UID = "c1143120-551e-4a47-ad47-2748d6f3c81c"
 DEFAULT_USER_NAME = "OLS"
+DEFAULT_KUBEADMIN_UID = "b6553200-0f7b-4c82-b1c5-9303ff18e5f0"
```
bparees, line 210: please add a TODO here, we want to replace this value w/ something that's dynamically generated based on the cluster and then stays constant.  That way the kubeadmin user on a given cluster has a unique uuid relative to the kubeadmin user on any other cluster.

ols/utils/auth_dependency.py
```diff
@@ -157,59 +157,69 @@ def _extract_bearer_token(header: str) -> str:
...
 
 
-async def auth_dependency(request: Request) -> tuple[str, str]:
-    """Validate FastAPI Requests for authentication and authorization.
+class AuthDependency:
+    """Create an AuthDependency Class that allows customizing the acces Scope path to check."""
 
-    Validates the bearer token from the request,
-    performs access control checks using Kubernetes TokenReview and SubjectAccessReview.
+    def __init__(self, virtual_path: str = "ols-access"):
```
bparees, line 163:
```suggestion
    def __init__(self, virtual_path: str = "/ols-access"):
```

?

bparees (commented):
please add some e2e tests for the access control, you can modify https://github.com/openshift/lightspeed-service/blob/86ea15827b02036b4d498003eb720cc4eadd7649/tests/scripts/utils.sh#L81-L86

to add users w/ only metrics access, and users who get their access via groups instead of directly.

then add/update the auth tests accordingly.

ols/app/metrics/metrics.py
nemerna: indeed looks like i mixed that durin cut/paste with the feedback one

PR #648, opened 2024-03-26
OLS-437: Model settings metrics tests

tisnik:
## Description

Model settings metrics tests

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] End to end tests

- Related Issue #[OLS-437](https://issues.redhat.com//browse/OLS-437)

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -604,6 +604,19 @@ def test_metrics() -> None:
     assert 'response_duration_seconds_sum{path="/metrics"}' in response.text
 
 
+def test_model_provider():
+    """Read configured model and provider from metrics."""
+    model, provider = get_model_and_provider(client)
+
+    # check available compbinations
+    assert model, provider in {
+        ("gpt-3.5-turbo", "openai"),
+        ("gpt-3.5-turbo", "azure_openai"),
+        ("ibm/granite-13b-chat-v2", "bam"),
+        ("ibm/granite-13b-chat-v2", "watsonx"),
+    }
+
+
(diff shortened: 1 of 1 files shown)


tests/e2e/test_api.py
```diff
@@ -604,6 +604,21 @@ def test_metrics() -> None:
     assert 'response_duration_seconds_sum{path="/metrics"}' in response.text
 
 
+@pytest.mark.cluster
+@pytest.mark.standalone
```
bparees, line 608: the default is to run tests that don't have tags, so you really only need to tag it if we need to be able to exclude it.

(the cluster tests are tagged as such because they require auth to be functional, so they can't be run against standalone where auth doesn't exist)

bparees: well, "Default" is the wrong way to describe it, but the expressions we use to pick the tests are all not-based.

so we are excluding tests w/ particular tags, not including tests w/ particular tags.

in any case, since this test can run anywhere, it does not need tags.

bparees:
/approve

remove the test tags (or convince me otherwise) and lgtm

tests/e2e/test_api.py
tisnik: actually it's not needed as we - as you wrote - can exclude tests, not include them. Initially I have in mind more specific approach based on `provider_*` marks, but it is not fully possible. Removed marks

PR #652, opened 2024-03-26
OLS-434: e2e test - load QnA from file

asamal4:
## Description

- Add question ids to QnA json file.
- Load QnA based on id from file in e2e test cases (currently model/scenario hard coded)

Script changes for passing arguments to pytest will be done separate PR.

### DIFF
--- a/scripts/validate_response.py
+++ b/scripts/validate_response.py
@@ -11,6 +11,7 @@ from llama_index.embeddings.huggingface import HuggingFaceEmbedding
 from pandas import DataFrame
 from scipy.spatial.distance import cosine, euclidean
 
+sys.path.append(os.path.dirname(os.path.dirname(__file__)))
 from ols.constants import NO_RAG_CONTENT_RESP
 
 
@@ -40,11 +41,11 @@ def _args_parser(args):
         help="Threshold value to be used for similarity score.",
     )
     parser.add_argument(
-        "-n",
-        "--num_questions",
-        type=int,
-        default=2,
-        help="Number of questions to be validated.",
+        "-q",
+        "--query_ids",
+        nargs="+",
+        default=["eval1"],
+        help="Ids of questions to be validated. Check json file for valid ids.",
     )
     return parser.parse_args(args)
 
@@ -85,9 +86,9 @@ class ResponseValidation:
         """Get response quality."""
         result_dict = defaultdict(list)
 
-        for idx in range(min(len(qa_pairs), args.num_questions)):
-            question = qa_pairs[idx]["question"]
-            answer = qa_pairs[idx]["answer"]
+        for query_id in args.query_ids:
+            question = qa_pairs[query_id]["question"]
+            answer = qa_pairs[query_id]["answer"]
 
             response = requests.post(
                 # API question validator can be disabled.
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -1,5 +1,6 @@
 """Integration tests for basic OLS REST API endpoints."""
 
+import json
 import os
 
 import pytest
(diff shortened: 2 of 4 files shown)


scripts/validate_response.py
```diff
@@ -11,6 +11,7 @@
 from pandas import DataFrame
 from scipy.spatial.distance import cosine, euclidean
 
+sys.path.append(os.path.dirname(os.path.dirname(__file__)))
```
tisnik, line 14: it's interesting that `make verify` is happy with statement before import, but ok ;)

tests/e2e/test_api.py
```diff
@@ -384,13 +380,15 @@ def test_query_call_with_improper_payload():
         assert "missing" in response.text
 
 
-def test_valid_question() -> None:
+def test_valid_question(qna_data) -> None:
     """Check the REST API /v1/query with POST HTTP method for valid question and no yaml."""
     endpoint = "/v1/query"
+    eval_query, eval_answer = get_eval_question_answer(qna_data, "eval1")
```
tisnik, line 386: hmm won't it be possible to use some fixture magic and specify `eval_query` and `eval_answer` as test parameters (via fixture)? https://pytest-with-eric.com/fixtures/pytest-fixture-with-arguments/

tisnik (commented):
It looks ok, I just have two comments, but these are just thoughts, not something that have to be there right now

tests/e2e/test_api.py
```diff
@@ -407,14 +405,15 @@ def test_valid_question() -> None:
...
         score = ResponseValidation().get_similarity_score(
-            json_response["response"], ANSWER1
+            json_response["response"], eval_answer
         )
         assert score <= SCORE_THRESHOLD
 
 
 @pytest.mark.standalone
-def test_valid_question_tokens_counter() -> None:
+def test_valid_question_tokens_counter(qna_data) -> None:
```
bparees, line 414: we need to keep our tests more focused.  This test is for checking the token counter metric.  It should not also be evaluating the quality of the response, let's keep that in a separate test.

Otherwise understanding the implications of a test failure (what does it mean is broken when this test fails) becomes harder, as does understanding what tests exist to test various functionality.

tests/e2e/test_api.py
```diff
@@ -26,29 +27,24 @@
...
+def qna_data():
+    """Load Question/answer pair for evaluation."""
+    with open("tests/test_data/question_answer_pair.json") as qna_f:
+        qa_pairs = json.load(qna_f)
+
+    # TODO: Use arguments
+    return qa_pairs["gpt"]["with_rag"]
+
+
+def get_eval_question_answer(qna_pair, qna_id):
```
bparees, line 41: add an argument here so we can choose between getting a RAG Q/A pair, and a no-rag one, and be clear about which we're getting/using.

tests/e2e/test_api.py
```diff
@@ -384,13 +380,15 @@ def test_query_call_with_improper_payload():
         assert "missing" in response.text
 
 
-def test_valid_question() -> None:
+def test_valid_question(qna_data) -> None:
```
bparees, line 383: with your current changes (using a Q/A pair loaded from the RAG dataset), this test needs `@pytest.mark.rag`, either that or you should use a non-rag Q/A pair.

tests/e2e/test_api.py
bparees: > As application is already started when we call e2e test. Model, with & without RAG option are already decided.

yes but this function can be called by multiple tests, and those tests can be marked as rag or no-rag

tests/e2e/test_api.py
bparees: > I am not evaluating response quality here, only picking a question..

ah, true.

> But I will revert to earlier hardcoded question to avoid loading file and possible error because of this.

yeah probably still a good idea not to use the eval question dataset for tests that aren't focused on the question/response.  keeps them slightly simpler.

PR #663, opened 2024-03-27
OLS-214: Add transcripts capture configuration

onmete:
## Description

Add transcripts capture configuration.

https://issues.redhat.com/browse/OLS-285 says:

> This is a post june task because we do not want to give OLS to anyone who's going to turn off data collection for now (the whole reason we are giving it to them is to get the data about how it performs)

My plan is to have it available in config, but hardcode it can't be turned off (just log a msg) later in the relevant place.

## Type of change

- [x] New feature

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -57,6 +57,8 @@ ols_config:
   user_data_collection:
     feedback_disabled: false
     feedback_storage: "/app-root/data/feedback"
+    transcripts_disabled: false
+    transcripts_storage: "/app-root/data/transcripts"
   tls_config:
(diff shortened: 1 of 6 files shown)


bparees:
> My plan is to have it available in config, but hardcode it can't be turned off (just log a msg) later in the relevant place.

this is fine, you don't have to hardcode it to stay on, we just wont expose a config setting for it in the operator CRD, so the operator will always generate an olsconfig.yaml that has it turned on.

cc @xrajesh @raptorsun 

/lgtm
/approve

xrajesh:
@bparees Copy . No config in CR for `transcripts_disabled`. By default, Userdata is sent , but disconnected clusters will not.

PR #664, opened 2024-03-27
OLS-434: Pass arguments to e2e

asamal4:
## Description

Pass arguments to e2e from script

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@
 ARTIFACT_DIR := $(if $(ARTIFACT_DIR),$(ARTIFACT_DIR),tests/test_results)
 TEST_TAGS := $(if $(TEST_TAGS),$(TEST_TAGS),"")
 SUITE_ID := $(if $(SUITE_ID),$(SUITE_ID),"nosuite")
+MODEL := $(if $(MODEL),$(MODEL),"gpt-3.5-turbo")
 
 images: ## Build container images
 	scripts/build-container.sh
@@ -53,7 +54,7 @@ check-coverage: test-unit test-integration  ## Unit tests and integration tests 
 test-e2e: ## Run e2e tests - requires running OLS server
 	@echo "Running e2e tests..."
 	@echo "Reports will be written to ${ARTIFACT_DIR}"
-	python -m pytest tests/e2e -o junit_suite_name="${SUITE_ID}" -m "${TEST_TAGS}" --junit-prefix="${SUITE_ID}" --junit-xml="${ARTIFACT_DIR}/junit_e2e_${SUITE_ID}.xml"
+	python -m pytest tests/e2e -o junit_suite_name="${SUITE_ID}" -m "${TEST_TAGS}" --junit-prefix="${SUITE_ID}" --junit-xml="${ARTIFACT_DIR}/junit_e2e_${SUITE_ID}.xml" --eval_model "${MODEL}"
 
 coverage-report:	test-unit ## Export unit test coverage report into interactive HTML
 	coverage html --data-file="${ARTIFACT_DIR}/.coverage.unit"
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -3,7 +3,9 @@
 
 def pytest_addoption(parser):
     """Argument parser for pytest."""
-    parser.addoption("--eval_model", default="gpt", help="Model to be evaluated.")
     parser.addoption(
-        "--eval_threshold", default="0.5", help="Evaluation threshold value."
+        "--eval_model",
+        default="gpt",
+        type=lambda v: "gpt" if "gpt" in v else "granite",
+        help="Model to be evaluated.",
     )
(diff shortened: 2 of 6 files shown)


tests/scripts/test-e2e-cluster.sh
```diff
@@ -42,16 +42,16 @@ rc=0
 set +e
 # runsuite arguments:
 # suiteid test_tags provider provider_keypath provider_url provider_project_id provider provider_deployment_name llm_model ols_image
-run_suite "azure_openai" "not standalone" "azure_openai" "$AZUREOPENAI_PROVIDER_KEY_PATH" "https://ols-test.openai.azure.com/" "" "0301-dep" "gpt-3.5-turbo" "$OLS_IMAGE"
+run_suite "azure_openai" "not standalone" "azure_openai" "$AZUREOPENAI_PROVIDER_KEY_PATH" "https://ols-test.openai.azure.com/" "" "0301-dep" "gpt-3.5-turbo" "$OLS_IMAGE" "0.5"
```
bparees, line 45: we should just default this in the test and if we feel the need to change it, change the default.  this is our baseline, so it doesn't make much sense to me that we'd need to pass it as an arg

Makefile
```diff
@@ -7,6 +7,9 @@ ARTIFACT_DIR := $(if $(ARTIFACT_DIR),$(ARTIFACT_DIR),tests/test_results)
 TEST_TAGS := $(if $(TEST_TAGS),$(TEST_TAGS),"")
 SUITE_ID := $(if $(SUITE_ID),$(SUITE_ID),"nosuite")
 
+EVAL_MODEL := $(if $(EVAL_MODEL),$(EVAL_MODEL),"gpt")
```
bparees, line 10: we already set provider+model env vars:
https://github.com/openshift/release/blob/7a74353dcd55f263e7755ee28f59b11c0b23812e/ci-operator/config/openshift/lightspeed-service/openshift-lightspeed-service-main.yaml#L68-L74

we should key off those, rather than introducing another env var that could end up out of sync w/ the ols config.

Makefile
```diff
@@ -7,6 +7,9 @@ ARTIFACT_DIR := $(if $(ARTIFACT_DIR),$(ARTIFACT_DIR),tests/test_results)
...
+EVAL_THRESHOLD := $(if $(EVAL_THRESHOLD),$(EVAL_THRESHOLD),"0.5")
```
bparees, line 11: default this in the test code and be done with it, i don't think we need a way to control it from outside the test code.

Makefile
```diff
@@ -53,7 +56,9 @@ check-coverage: test-unit test-integration  ## Unit tests and integration tests
 test-e2e: ## Run e2e tests - requires running OLS server
 	@echo "Running e2e tests..."
 	@echo "Reports will be written to ${ARTIFACT_DIR}"
-	python -m pytest tests/e2e -o junit_suite_name="${SUITE_ID}" -m "${TEST_TAGS}" --junit-prefix="${SUITE_ID}" --junit-xml="${ARTIFACT_DIR}/junit_e2e_${SUITE_ID}.xml"
+	# Currently eval model is gpt always, till we add QnA for different model.
+	python -m pytest tests/e2e -o junit_suite_name="${SUITE_ID}" -m "${TEST_TAGS}" --junit-prefix="${SUITE_ID}" --junit-xml="${ARTIFACT_DIR}/junit_e2e_${SUITE_ID}.xml" \
+	--eval_model "gpt" --eval_threshold ${EVAL_THRESHOLD}
```
bparees, line 61: see comments above about defaulting threshold and getting the model from existing env vars

tests/scripts/test-e2e-standalone.sh
```diff
@@ -84,4 +84,10 @@ fi
 set -e
 
 echo Done waiting for OLS server start, running e2e
-SUITE_ID=standalone TEST_TAGS="not cluster and not rag" make test-e2e
+
+if [[ $MODEL = *"gpt"* ]]; then
+  EVAL_MODEL="gpt"
```
bparees, line 89: rather than doing this here, i'd prefer the test code make this determination.

we have too much logic in our Makefile + test execution scripts already (for which i share the blame), it makes it hard for people to understand the flow as well as makes it harder to reuse the tests in new environments

bparees:
looks like at least part of the failures here are the same thing i'm tackling here:
https://github.com/openshift/lightspeed-service/pull/675/commits/d4169bfcdce0b0aabeab7806525771da8effc23a

bparees:
> May be chunk being retrieved is slightly different. Have we re-created index with new embedding model ? rag image sha is still same as before

new embedding model(and embeddings) should not be in use yet, not until we switch the SHA.  My understanding is OLS picks the embedding model up from the rag content image, along w/ the rag content itself.  So neither should change w/o a SHA change.

And any time we switch the embedding content(switch the SHA), that change has to go through CI in this repo, so we shouldn't have a case where we change the RAG content/embedding under the covers and break CI.  If we managed to do that, we need to figure out how that happened and prevent it in the future.

in my PR at least, the failures were also specific to bam + watsonx, not openai, which made me assume perhaps they'd deployed a new version of their models that tweaked the output slightly.

PR #671, opened 2024-03-28
Setup Uvicorn log level

tisnik:
## Description

Setup Uvicorn log level

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -73,4 +73,5 @@ if __name__ == "__main__":
             ssl_keyfile=config.ols_config.tls_config.tls_key_path,
             ssl_certfile=config.ols_config.tls_config.tls_certificate_path,
             ssl_keyfile_password=config.ols_config.tls_config.tls_key_password,
+            log_level="debug",
         )


runner.py
```diff
@@ -73,4 +73,5 @@ def configure_gradio_ui_envs() -> None:
             ssl_keyfile=config.ols_config.tls_config.tls_key_path,
             ssl_certfile=config.ols_config.tls_config.tls_certificate_path,
             ssl_keyfile_password=config.ols_config.tls_config.tls_key_password,
+            log_level="debug",
```
bparees, line 76: what's the default level?  is this increasing the verbosity or decreasing it?

also seems like this should be configured via another field in https://github.com/openshift/lightspeed-service/blob/main/ols/app/models/config.py#L559-L560

tisnik: it is basically decreasing # messages in logs

bparees:
/hold

discussed in slack, this sets the loglevel for uvicorn, so we need a way to control it or we won't be able to enable these messages on a customer cluster for debug purposes.

PR #672, opened 2024-03-28
Add latest vec db folder to gitignore/example

asamal4:
## Description

Update latest default vector db folder name.


bparees:
/approve
/lgtm

thanks

PR #676, opened 2024-03-29
OLS-375: Log OLS HTTP request+response payloads

syedriko:
## Description

Implemented logging of HTTP requests and responses going to and from OLS FastAPI. Said logging only occurs if `ols_config.logging_config.app_log_level` is set to `debug`. For correlation of requests and responses, client_IP:port are logged.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes # https://issues.redhat.com/browse/OLS-375

## Testing

An example of log for a conversation with Gradio UI:
```
2024-03-28 23:09:09,140 [ols.app.main:main.py:76] DEBUG: Request from 127.0.0.1:33730 Headers({'host': '127.0.0.1:8080', 
'user-agent': 'python-requests/2.31.0', 'accept-encoding': 'gzip, deflate', 'accept': 'application/json', 'connection': 'keep-alive', 
'content-type': 'application/json', 'content-length': '53'}), Body: {"query": "What is the latest version of OpenShift?"}
---
2024-03-28 23:09:16,072 [ols.app.main:main.py:83] DEBUG: Response to 127.0.0.1:33730 MutableHeaders({'content-length': 
'603', 'content-type': 'application/json'})
2024-03-28 23:09:16,073 [ols.app.main:main.py:92] DEBUG: Response to 127.0.0.1:33730 Body chunk: 
{"conversation_id":"8771b244-3fdc-4c56-90b4-3d23c36ac381","response":"As of my last update in April 2023, the latest 
version of Red Hat OpenShift Container Platform is 4.10. However, it's important to note that Red Hat frequently updates 
OpenShift with new versions and features, so there may be a newer version available now. I recommend checking the 
official Red Hat OpenShift website or contacting their support for the most current information on OpenShift 
versions.","referenced_documents":["https://docs.openshift.com/container-platform/4.15/whats_new/deprecated-
features.html"],"truncated":false}
```

### DIFF
--- a/ols/app/main.py
+++ b/ols/app/main.py
@@ -1,9 +1,10 @@
 """Entry point to FastAPI-based web service."""
 
 import logging
-from collections.abc import Awaitable, Callable
+from collections.abc import AsyncGenerator, Awaitable, Callable
 
 from fastapi import FastAPI, Request, Response
+from starlette.responses import StreamingResponse
 
 from ols.app import metrics, routers
 from ols.src.ui.gradio_ui import GradioUI
@@ -50,6 +51,55 @@ async def rest_api_counter(
     if not path.endswith("/metrics"):
         # just update metrics
         metrics.rest_api_calls_total.labels(path, response.status_code).inc()
+    return response
+
+
+@app.middleware("")
+async def log_requests_responses(
+    request: Request, call_next: Callable[[Request], Awaitable[Response]]
+) -> Response:
+    """Middleware for logging of HTTP requests and responses, at debug level."""
+    # Bail out early if not logging
+    if not logger.isEnabledFor(logging.DEBUG):
+        return await call_next(request)
+
+    request_log_message = (
+        f"Request from {request.client.host}:{request.client.port}"
+        f" {request.headers}, Body: "
+    )
+
+    request_body = await request.body()
+    if request_body:
+        request_log_message += f"{request_body.decode('utf-8')}"
+    else:
+        request_log_message += "None"
+
+    logger.debug(request_log_message)
+
+    response = await call_next(request)
+
+    logger.debug(
+        f"Response to {request.client.host}:{request.client.port} {response.headers}"
+    )
+
+    async def stream_response_body(
+        response_body: AsyncGenerator[bytes, None]
+    ) -> AsyncGenerator[bytes, None]:
+        async for chunk in response_body:
+            logger.debug(
+                f"Response to {request.client.host}:{request.client.port} "
+                f"Body chunk: {chunk.decode('utf-8', errors='ignore')}"
+            )
+            yield chunk
+
+    if isinstance(response, StreamingResponse):
+        # The response is already a streaming response
+        response.body_iterator = stream_response_body(response.body_iterator)
+    else:
+        # Convert non-streaming response to a streaming response to log its body
+        response_body = response.body
+        response = StreamingResponse(stream_response_body(iter([response_body])))
+
     return response
 
 


ols/app/main.py
```diff
@@ -53,4 +54,60 @@ async def rest_api_counter(
...
+        ) -> AsyncGenerator[bytes, None]:
+            async for chunk in response_body:
+                if logger.isEnabledFor(logging.DEBUG):
+                    logger.debug(
+                        f"Response {request.client.host}:{request.client.port} "
+                        f"chunk: {chunk.decode('utf-8', errors='ignore')}"
+                    )
+                yield chunk
+
+        if logger.isEnabledFor(logging.DEBUG):
```
tisnik, line 98: it will be `True` here everytime, right? so no condition needed

ols/app/main.py
```diff
@@ -53,4 +54,60 @@ async def rest_api_counter(
...
+        # Log response headers
+        logger.debug(
+            f"Response {request.client.host}:{request.client.port} {response.headers}"
+        )
+
+        async def stream_response_body(
+            response_body: AsyncGenerator[bytes, None]
+        ) -> AsyncGenerator[bytes, None]:
+            async for chunk in response_body:
+                if logger.isEnabledFor(logging.DEBUG):
```
tisnik, line 91: dtto

ols/app/main.py
```diff
@@ -53,4 +54,60 @@ async def rest_api_counter(
...
+            request_log_message += f"{request_body.decode('utf-8')}"
+        else:
+            request_log_message += "None"
+
+        logger.debug(request_log_message)
+
+    # Process request
+    response = await call_next(request)
+
+    if logger.isEnabledFor(logging.DEBUG):
```
raptorsun, line 81: we can return response here to use only one checking on logger.isEnabledFor(logging.DEBUG)
```
if not logger.isEnabledFor(logging.DEBUG): 
  return response
```

ols/app/main.py
```diff
@@ -53,4 +54,60 @@ async def rest_api_counter(
     return response
 
 
+@app.middleware("")
+async def log_requests_responses(
+    request: Request, call_next: Callable[[Request], Awaitable[Response]]
+) -> Response:
+    """Middleware for logging of HTTP requests and responses, at debug level."""
+    if logger.isEnabledFor(logging.DEBUG):
```
raptorsun, line 62: this section can be moved after the check on line 81 to make it more clear :)

raptorsun: not sure about how long a log line can be, we can send both request and response in one log entry , or we send 2 log entries at very close times after we have the response. So that request and response can be close in log.

bparees: request and response should be separate log entries, because OLS will do(and log) other things in between getting the request and sending the response, which we want to see in order.

ols/app/main.py
syedriko: true, my oversight

ols/app/main.py
syedriko: I'm not sure what you mean. I don't think we can get away with just a single check, we need to produce and return a response even if we're not logging
```
if logger.isEnabledFor(logging.DEBUG):
    # log request
response = await call_next(request)
if logger.isEnabledFor(logging.DEBUG):
    # log response
return response
```

ols/app/main.py
syedriko: Can't do one entry because the response body is streamed. There are at least 3:
- request
- response headers
- response body
They are all correlated by client_ip:port so we don't have to manually hunt for which is which.

ols/app/main.py
raptorsun: the current version is good. I mean that it seems more clear to me if the code is arranged like:
```python
if not logger.isEnabledFor(logging.DEBUG): 
    return call_next(request) 
# log request 
response = await call_next(request) 
# log response 
return response
```
I like the less indented layout and put the most probable code path at the beginning.

syedriko: Got it, thanks!

PR #682, opened 2024-04-02
chore: rename e2e util files

asamal4:
## Description

renamed files:

- `test_` to `helper_` (file not having a test case).
- `consts` to `constanst`


tisnik (commented):
looks ok

PR #685, opened 2024-04-02
Update readme for local vector db

asamal4:
## Description

Update readme to use new vector db to local


tisnik (commented):
it looks ok, thank you

PR #688, opened 2024-04-02
OLS-434: make target for response evaluation

asamal4:
## Description

- Make target for response evaluation.
- Run only if flag is set for evaluation.

- Closes #

### DIFF
--- a/scripts/validate_response.py
+++ b/scripts/validate_response.py
@@ -31,20 +31,21 @@ def _args_parser(args):
         "--model",
         choices=["gpt", "granite"],
         default="gpt",
+        type=lambda v: "gpt" if "gpt" in v else "granite",
         help="Model for which responses will be evaluated.",
     )
     parser.add_argument(
         "-t",
         "--threshold",
         type=float,
-        default=0.25,
+        default=0.7,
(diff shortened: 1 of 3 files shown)


scripts/validate_response.py
```diff
@@ -31,20 +31,21 @@ def _args_parser(args):
         "--model",
         choices=["gpt", "granite"],
         default="gpt",
+        type=lambda v: "gpt" if "gpt" in v else "granite",
```
tisnik, line 34: is not it `str`?

tests/scripts/utils.sh
```diff
@@ -107,6 +107,12 @@ function run_suite() {
   fi
   
   SUITE_ID=$1 TEST_TAGS=$2 MODEL=$8 make test-e2e
+
+  if ! [ -z ${RESPONSE_EVALUATION:-} ]; then
```
bparees, line 111: the intent is that we're going to have a wholly separate test job that sets this env variable and thereby runs the response evaluation check, right?

given that, do you actually want "make test-e2e" to also run, when you're doing that?  or should this be an either/or situation?  (either run the e2e tests, or run the response evaluation report/check)

PR #694, opened 2024-04-02
postgres container file for OLS

vishnuchalla:
## Description
Postgres Containerfile for lightspeed service.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Closes # https://issues.redhat.com/browse/OLS-438

## Testing
Tested in local by building the image. Made sure its running inside a container and also checked the filesystem
Build and Push
```
podman build -f Containerfile.postgres -t=quay.io/vchalla/postgres:latest .
podman push quay.io/vchalla/postgres:latest
```
Verification
```
podman run -d -it quay.io/vchalla/postgres:latest
podman exec -it 852b4f0a3a77
cat /var/lib/pgsql/data/postgresql.conf
```

### DIFF
--- a/Containerfile.postgres
+++ b/Containerfile.postgres
@@ -1,0 +1,5 @@
+# Use offical quay.io postgres image
+FROM quay.io/sclorg/postgresql-13-c9s:latest
+
+# Expose the PostgreSQL port
+EXPOSE 5432
--- a/Containerfile.postgres
+++ b/Containerfile.postgres
@@ -1,5 +1,0 @@
-# Use offical quay.io postgres image
-FROM quay.io/sclorg/postgresql-13-c9s:latest
-
-# Expose the PostgreSQL port
-EXPOSE 5432
--- a/postgres/Containerfile
+++ b/postgres/Containerfile
@@ -1,0 +1,56 @@
+FROM quay.io/sclorg/s2i-core-c9s:c9s
+
+ENV POSTGRESQL_VERSION=13 \
+    HOME=/var/lib/pgsql \
+    PGUSER=postgres \
+    APP_DATA=/opt/app-root
+
+EXPOSE 5432
+
+COPY root/usr/libexec/fix-permissions /usr/libexec/fix-permissions
+
+# This image must forever use UID 26 for postgres user so our volumes are
+# safe in the future. This should *never* change, the last test is there
+# to make sure of that.
+RUN  { yum -y module enable postgresql:13 || :; } && \
+    INSTALL_PKGS="rsync tar gettext bind-utils nss_wrapper postgresql-server postgresql-contrib" && \
+    INSTALL_PKGS="$INSTALL_PKGS pgaudit" && \
+    yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    postgres -V | grep -qe "$POSTGRESQL_VERSION\." && echo "Found VERSION $POSTGRESQL_VERSION" && \
+    yum -y clean all --enablerepo='*' && \
+    localedef -f UTF-8 -i en_US en_US.UTF-8 && \
+    test "$(id postgres)" = "uid=26(postgres) gid=26(postgres) groups=26(postgres)" && \
+    mkdir -p /var/lib/pgsql/data && \
+    /usr/libexec/fix-permissions /var/lib/pgsql /var/run/postgresql
+
+# Get prefix path and path to scripts rather than hard-code them in scripts
+ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/postgresql \
+    ENABLED_COLLECTIONS=
+
+COPY root /
+COPY ./s2i/bin/ $STI_SCRIPTS_PATH
+
+# Hard links are not supported in Testing Farm approach during sync to guest
+# operation system. Therefore tests are failing on error
+# /usr/libexec/s2i/run no such file or directory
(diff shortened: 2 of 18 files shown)


tisnik (commented):
it looks ok so far. Do you want/plan to change the 
`/var/lib/pgsql/data/pg_hba.conf` in any way (how to connect, who can connect etc.)?

vishnuchalla:
> it looks ok so far. Do you want/plan to change the `/var/lib/pgsql/data/pg_hba.conf` in any way (how to connect, who can connect etc.)?

No. Not at this moment with the requirements we have I believe.

Containerfile.postgres
```diff
@@ -0,0 +1,22 @@
...
+    rm -rf /var/cache/dnf
+
+# Create a directory for PostgreSQL data
+RUN mkdir -p /var/lib/pgsql/data && \
+    chown -R postgres:postgres /var/lib/pgsql && \
+    chmod -R 700 /var/lib/pgsql/data
+
+# Initialize the PostgreSQL database
+USER postgres
+RUN initdb -D /var/lib/pgsql/data
```
bparees, line 16: have you tested running this image on an OCP cluster using a random uid?  i think you're going to have permission errors between this (using user "postgres" to init the filesystem at build time, when postgres won't be the user at runtime), and the chmod 700 (not 755) above which means the random uid won't have permission to write to /var/lib/pgsql/data

also /var/lib/pgsql/data is presumably going to be a volume (even if only an emptydir volume initially).

so you need to think differently about how the filesystem gets initialized at runtime.

vishnuchalla: Yes we need this `Dockerfile` with this format. I have tested it in openshift with the below yaml and it works fine. For some reason it doesn't work with any other UIDs and only with postgres UID (i.e 26). Looks like inbuilt postgres package is configured that way.

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: postgresql
spec:
  replicas: 1
  selector:
    matchLabels:
      app: postgresql
  template:
    metadata:
      labels:
        app: postgresql
    spec:
      containers:
      - name: postgresql
        image: quay.io/vchalla/postgres:latest
        imagePullPolicy: Always
        ports:
        - containerPort: 5432
        securityContext:
          runAsUser: 26
          fsGroup: 26
          runAsNonRoot: false
```
Adding `emptyDir` as volume will override the ` /var/lib/pgsql/data`  (i.e leading to runtime issues) but adding a persistent volume to the data won't. So I was thinking initially we can simply go with the provided `/var/lib/pgsql/data` folder instead of `emptyDir` because both have the same purpose to store data in-memory.

bparees: We cannot be constrained to running the container as uid 26, it will require us to request elevated OCP permissions which will raise concerns for customers.  Our own SCL postgres example image does not have that constraint from what I see (I was involved in creating it, but it was a long time ago at this point):
https://github.com/sclorg/postgresql-container/blob/master/examples/postgresql-ephemeral-template.json#L69-L183 

> Adding emptyDir as volume will override the  /var/lib/pgsql/data (i.e leading to runtime issues) but adding a persistent volume to the data won't. 

why does adding an emptydir cause problems that a persistent volume does not cause?

as i mentioned on our call, we need to include logic in a startup script that properly initializes the DB on the mounted filesystem *at runtime* and not at image build time.  So it should not matter if the storage dir is empty initially.

> So I was thinking initially we can simply go with the provided /var/lib/pgsql/data folder instead of emptyDir because both have the same purpose to store data in-memory.

it's not the same, the storage is allocated and controlled differently (it's also not necessarily in-memory.  emptydir volumes are allocated from the node disk by default, though they can be in memory.  i don't think the container filesystem would ever be in memory).  Using an emptydir puts us in an easy position to pivot to persistent storage in the future.  Using container storage does not, that's why i want the image using a mounted volume from the start.

vishnuchalla: Understood the above concerns. Looks like we already have an official postgres image available in [Quay](https://github.com/quay/quay-operator/blob/master/kustomize/components/postgres/postgres.deployment.yaml#L35). Do we want to use the same to start building resources using the same pattern followed in the shared repo directly in our operator code?

bparees: if you are asking "should we start adding logic to the operator to construct a postgres deployment that follows the pattern seen in your link" then my answer is "the operator needs to construct a deployment for our postgres image", i don't care what it is based on/copied from, as long as the deployment works correctly with our postgres image.

vishnuchalla: Sorry for a bit late response on this. I just wanted to thoroughly test and confirm it. I have updated the `Containerfile` with a stable quay.io image that we can use. Tested and verified it by integrating with https://github.com/openshift/lightspeed-operator/pull/108 and https://github.com/openshift/lightspeed-service/pull/634. Works and fulfills all our requirements. Let me know if we are missing on something. Thanks

Containerfile.postgres
```diff
@@ -0,0 +1,5 @@
+# Use offical quay.io postgres image
+FROM quay.io/sclorg/postgresql-13-c9s:latest
```
bparees, line 2: I was anticipating that you'd base our containerfile on the one that builds this scl image, not just use the image directly, because this leaves us at the mercy of changes that can happen to this image w/o our control.

vbelouso:
@vishnuchalla can you please close this PR in favor of #1040

PR #704, opened 2024-04-03
OLS-420: New WatsonxLLM used instead of deprecated one

tisnik:
## Description

New WatsonxLLM used instead of deprecated one
It requires changes in code and tests too

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/src/llms/providers/watsonx.py
+++ b/ols/src/llms/providers/watsonx.py
@@ -3,14 +3,11 @@
 import logging
 from typing import Any
 
-from ibm_watson_machine_learning.foundation_models import Model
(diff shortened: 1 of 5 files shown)


tisnik:
@ondrejmular do we need to keep the unit test? The thing is that WatsonxLLM performs several net transfers in initialization phase, and it's quite hard to mock everything right. So the solution would be to mock the WatsonxLLM class itself, but the test would be a bit pointless. WDYT Ondro?

PR #705, opened 2024-04-03
Refactor question validation logic

bparees:
Make the question validation logic more tolerant (only reject questions when the llm definitively responds with a rejection)

Make the keywords more distinct (REJECT/ALLOW instead of VALID/INVALID) to hopefully help the LLM do a better job of picking the right keyword.

Rephrase the question validation prompt template

Make the question validation helper return true/false, not the keyword from the llm, so we can isolate the decision making process of interpretting the llm evaluation, from other OLS logic.

Make generate_response only responsible for asking for an inference from the llm, not doing question validation or interpretting a question validation result argument.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -57,14 +57,21 @@ def conversation_request(
 
     # Validate the query
     if not previous_input:
-        validation_result = validate_question(conversation_id, llm_request)
+        valid = validate_question(conversation_id, llm_request)
     else:
         logger.debug("follow-up conversation - skipping question validation")
-        validation_result = constants.SUBJECT_VALID
+        valid = True
 
-    response, referenced_documents, truncated = generate_response(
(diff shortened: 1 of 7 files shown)


bparees:
@asamal4 updated, ptal

bparees:
> Now that we have added Explain your answer instruction to the prompt

oof, actually i didn't mean to leave that in, i was debugging something (it was fairly hard to get granite to respond properly to even our test questions.  i'm still concerned it will be fragile)

but interesting that it seems to still work despite me leaving that in place

/hold

bparees:
/hold cancel

@asamal4 ok, this is ready again

PR #710, opened 2024-04-05
OLS-413: Make LLMResponse include page titles for referenced documents

syedriko:
## Description

LLMResponse now contains referenced documents that are pairs of (docs_url, title). Title comes from the "title" element of metadata in embedding nodes, added in https://github.com/openshift/lightspeed-rag-content/pull/9, which needs to merge first.
RAG content is now based on the sentence-transformers/all-mpnet-base-v2 embedding model, which necessitated RAG_SIMILARITY_CUTOFF_L2 increase and changes to a few asserts about what documents are retrieved for what queries.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes # https://issues.redhat.com/browse/OLS-413

## Testing

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,5 +1,5 @@
 # vim: set filetype=dockerfile
-ARG LIGHTSPEED_RAG_CONTENT_DIGEST=sha256:d15bf56776c40a8709b0e648e3b0f043de63b24ad8f59eeea6f8d965dfcbe4e3
+ARG LIGHTSPEED_RAG_CONTENT_DIGEST=sha256:c40bdfe55a827f46fd7775b4dcfec39bb915c8338bc6c0c085e0284ca8a08ebd
 
 FROM quay.io/openshift/lightspeed-rag-content@${LIGHTSPEED_RAG_CONTENT_DIGEST} as lightspeed-rag-content
 
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -10,6 +10,7 @@ from llama_index.indices.vector_store.base import VectorStoreIndex
 
 from ols.app.metrics import TokenMetricUpdater
 from ols.app.models.config import ProviderConfig
+from ols.app.models.models import ReferencedDocument
 from ols.constants import NO_RAG_CONTENT_RESP, RAG_CONTENT_LIMIT
 from ols.src.prompts.prompt_generator import generate_prompt
 from ols.src.query_helpers.query_helper import QueryHelper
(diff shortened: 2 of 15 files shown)


ols/src/query_helpers/docs_summarizer.py
```diff
@@ -96,7 +97,12 @@ def summarize(
             logger.warning("Proceeding without RAG content. Check start up messages.")
 
         rag_context = "\n\n".join(rag_context_data.get("text", []))
-        referenced_documents = rag_context_data.get("docs_url", [])
+        referenced_documents = [
+            ReferencedDocument(docs_url=docs_url, title=title)
+            for docs_url, title in zip(
+                rag_context_data.get("docs_url", []), rag_context_data.get("title", [])
```
bparees, line 103: do we not need an updated RAG image for this metadata to be present?

bparees (commented):
please add an e2e test for the api behavior.

also we should not merge this until @kyoto can confirm it won't break the console

kyoto:
Response format change LGTM. Merging shouldn't break anything except that the reference docs won't show up until the frontend is updated to match. I'll made that change after this PR merges.

ols/src/query_helpers/docs_summarizer.py
syedriko: which is now done

syedriko:
> please add an e2e test for the api behavior.
> 
> also we should not merge this until @kyoto can confirm it won't break the console

who's confirmation we now have https://github.com/openshift/lightspeed-service/pull/710#issuecomment-2041904910

Containerfile
```diff
@@ -1,5 +1,5 @@
 # vim: set filetype=dockerfile
-ARG LIGHTSPEED_RAG_CONTENT_DIGEST=sha256:d15bf56776c40a8709b0e648e3b0f043de63b24ad8f59eeea6f8d965dfcbe4e3
+ARG LIGHTSPEED_RAG_CONTENT_DIGEST=sha256:56f11869a44c9079c8f8279e801a325d21d9a235a9cca3691567eb0e459012b4
```
bparees, line 2: i've tagged this sha in quay so it won't be GCed

bparees (commented):
is there no e2e test we can update that confirms the responses have the full RAG reference (url and title)?

if there isn't one already, then we need to introduce one.

syedriko:
> is there no e2e test we can update that confirms the responses have the full RAG reference (url and title)?
> 
> if there isn't one already, then we need to introduce one.

I'll modify https://github.com/openshift/lightspeed-service/blob/25f8fd71fed5549a6440e9007736b0fc4456d292/tests/e2e/test_api.py#L402:L404 and the other places in the e2e test.

syedriko:
@bparees PTAL

syedriko:
@asamal4 This is the PR that introduces sentence-transformers/all-mpnet-base-v2 embedding model, so I bumped RAG_SIMILARITY_CUTOFF_L2 changed a few asserts about what documents are retrieved for what queries.

PR #712, opened 2024-04-05
remove unused constants

asamal4:
## Description

Removed unused question validation / prompt related constants.


ols/src/prompts/prompts.py
```diff
@@ -50,87 +50,87 @@
...
-Statement: No, I don't think that is wise.
-Response: 0
-
-Statement: Apples are red.
-Response: 9
-
-Statement: {statement}
-Response:
-"""
+# SUMMARIZATION_TEMPLATE = """
```
bparees, line 53: you can just delete all of these.  If we find ourselves implementing functionality where we want them again we're probably going to want to start fresh w/ the prompt text anyway, there was no special research effort invested in coming up with these particular prompt templates.

PR #713, opened 2024-04-05
e2e: Check for max default model+provider

tisnik:
## Description

e2e: Check for max default model+provider, to make sure the logic to setup metrics or config loading/validation is ok

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] e2e tests

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -426,6 +426,15 @@ def test_model_provider():
     }
 
 
+def test_one_default_model_provider():
+    """Check if one model and provider is selected as default."""
+    states = metrics_utils.get_enable_status_for_all_models(metrics_client)
+    enabled_states = [state for state in states if state is True]
+    assert (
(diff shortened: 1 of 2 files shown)


tests/e2e/test_api.py
```diff
@@ -426,6 +426,15 @@ def test_model_provider():
     }
 
 
+def test_max_one_default_model_provider():
+    """Check if max. one model and provider is selected as default."""
+    states = metrics_utils.get_enable_status_for_all_models(metrics_client)
+    enabled_states = [state for state in states if state is True]
+    assert (
+        len(enabled_states) <= 1
```
bparees, line 434: we should expect exactly one.  We control the configuration when we are testing.

tisnik: updated

PR #724, opened 2024-04-09
OLS-375: added redaction of sensitive headers

syedriko:
## Description

This is an improvement for https://github.com/openshift/lightspeed-service/pull/676 that adds redaction of sensitive HTTP headers, such as "authorization", "proxy-authorization", "cookie", "www-authenticate", "proxy-authenticate" and "set-cookie".

## Type of change

- [] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes # https://issues.redhat.com/browse/OLS-375

## Testing
Now sensitive header values are replaces with XXXXX:
```
2024-04-08 20:12:23,437 [ols.app.main:main.py:88] DEBUG: Request from 127.0.0.1:36978 Headers({"host":"127.0.0.1:8080", "user-agent":"curl/8.2.1", "authorization":"XXXXX", "accept":"application/json", "content-type":"application/json", "content-length":"29"}), Body: {"query": "What is the OLM?"}

```

### DIFF
--- a/ols/app/main.py
+++ b/ols/app/main.py
@@ -4,8 +4,10 @@ import logging
 from collections.abc import AsyncGenerator, Awaitable, Callable
 
 from fastapi import FastAPI, Request, Response
+from starlette.datastructures import Headers
 from starlette.responses import StreamingResponse
 
+from ols import constants
 from ols.app import metrics, routers
 from ols.src.ui.gradio_ui import GradioUI
 from ols.utils import config
@@ -54,6 +56,15 @@ async def rest_api_counter(
     return response
 
 
+def _log_headers(headers: Headers, to_redact: set[str]) -> str:
+    """Serialize headers into a string while redacting sensitive values."""
+    pairs = []
+    for h, v in headers.items():
+        value = "XXXXX" if h.lower() in to_redact else v
+        pairs.append(f'"{h}":"{value}"')
+    return "Headers({" + ", ".join(pairs) + "})"
+
+
 @app.middleware("")
 async def log_requests_responses(
     request: Request, call_next: Callable[[Request], Awaitable[Response]]
@@ -63,10 +74,11 @@ async def log_requests_responses(
     if not logger.isEnabledFor(logging.DEBUG):
         return await call_next(request)
 
-    request_log_message = (
-        f"Request from {request.client.host}:{request.client.port}"
-        f" {request.headers}, Body: "
+    request_log_message = f"Request from {request.client.host}:{request.client.port} "
+    request_log_message += _log_headers(
+        request.headers, constants.HTTP_REQUEST_HEADERS_TO_REDACT
     )
+    request_log_message += ", Body: "
 
     request_body = await request.body()
     if request_body:
@@ -78,9 +90,13 @@ async def log_requests_responses(
 
     response = await call_next(request)
 
-    logger.debug(
-        f"Response to {request.client.host}:{request.client.port} {response.headers}"
+    response_headers_log_message = (
+        f"Response to {request.client.host}:{request.client.port} "
     )
+    response_headers_log_message += _log_headers(
+        response.headers, constants.HTTP_RESPONSE_HEADERS_TO_REDACT
+    )
+    logger.debug(response_headers_log_message)
 
     async def stream_response_body(
         response_body: AsyncGenerator[bytes, None]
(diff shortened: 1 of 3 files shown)


syedriko:
> are you going to update tests too?
yep, just need to figure out how to test this

ols/app/main.py
```diff
@@ -54,6 +56,15 @@ async def rest_api_counter(
     return response
 
 
+def _log_headers(headers: Headers, to_redact: Set[str]) -> str:
```
tisnik, line 59: can be just `set`

syedriko: fixed

syedriko:
@tisnik I added a test for header redaction, PTAL

tisnik (commented):
it looks  ok. Frankly, the test written in Bash is a bit flaky, but all Bash scripts are ;)

tests/scripts/test-e2e-standalone.sh
```diff
@@ -87,3 +98,15 @@ set -e
 
 echo Done waiting for OLS server start, running e2e
 OLS_URL="http://localhost:8080" SUITE_ID=standalone TEST_TAGS="not cluster and not rag" make test-e2e
+rc=$?
```
tisnik, line 101: is not it always 0?

syedriko: not if e2e tests fail

tisnik: and should we continue testing then?

tests/scripts/test-e2e-standalone.sh
```diff
@@ -51,8 +55,15 @@ if [ ! -e "$PROVIDER_KEY_PATH" ]; then
 fi
 
 export MODEL="${MODEL:-gpt-3.5-turbo-1106}"
+if [ -z ${TEST_HTTP_HEADER_REDACTION:-} ]; then
+  LOG_LEVEL=info
+else
+  LOG_LEVEL=debug
+fi
+export LOG_LEVEL
```
bparees, line 63: we can just run OLS w/ debug level logging in all our e2e jobs (cluster and standalone). I thought we already were.

tests/scripts/test-e2e-standalone.sh
```diff
@@ -83,7 +94,18 @@ if [ $? -ne 0 ]; then
...
+
+if [ ! -z ${TEST_HTTP_HEADER_REDACTION:-} ]; then
+  for h in authorization proxy-authorization cookie; do
+    curl -X 'POST' 'http://localhost:8080/v1/query' -H 'accept: application/json' \
+    -H 'Content-Type: application/json' -H "$h: random_value" \
+    -d '{"query": "write a deployment yaml for the mongodb image"}'
+
+    grep -e "\"$h\":\"XXXXX\"" "$OLS_LOGS" || { echo "header $h was not redacted"; rc=1; }
+  done
+fi
```
bparees, line 110: writing tests directly into our bash script isn't sustainable or maintainable.

we've got other python-based tests that inspect the log output, @onmete has written a few, you can copy his examples.  You can mark the test for cluster-based testing only, if that makes it easier.

over time i'd like to see test-e2e-standalone.sh go away entirely, but for now:
1) it's a useful way for developers to do some local testing of the current code
2) some tests would be more difficult to write for running against OLS running in a cluster

but in general, testing against ols running in a cluster should be our preference as it reflects the way OLS actually works, and long term the goal for our e2e CI should be:

1) use the latest operator image to deploy the latest ols images to a cluster (where "latest" means last published, or built from the PR being tested, as applicable)
2) run our e2e against that env

bparees (commented):
the implementation looks ok, the testing needs work.

though i also find myself wondering how much we care about logging the headers.  The real goal I had in mind for logging the requests/responses was so we could see the body.  I think it's unlikely the headers are going to be particularly interesting, especially once redacted.  About all it will tell us is "yes they sent an auth header" or "no they didn't", and i have reasonable confidence that our auth logic properly handles+logs that information already.

So wdyt about just removing the header logging logic entirely, rather than spending the effort on doing the testing "right"?

syedriko:
> the implementation looks ok, the testing needs work.
> 
> though i also find myself wondering how much we care about logging the headers. The real goal I had in mind for logging the requests/responses was so we could see the body. I think it's unlikely the headers are going to be particularly interesting, especially once redacted. About all it will tell us is "yes they sent an auth header" or "no they didn't", and i have reasonable confidence that our auth logic properly handles+logs that information already.
> 
> So wdyt about just removing the header logging logic entirely, rather than spending the effort on doing the testing "right"?

I gave it another shot, in Python, before ripping logging of headers out.

PR #725, opened 2024-04-09
Proper names for fixtures not returning values

tisnik:
## Description

By convention, fixtures that don't return a value should be named with a
leading underscore, while fixtures that do return a value should not.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/integration/test_authorized.py
+++ b/tests/integration/test_authorized.py
@@ -15,7 +15,7 @@ from tests.mock_classes.mock_k8s_api import (
 
 
 @pytest.fixture(scope="module")
-def setup():
+def _setup():
     """Setups the test client."""
     global client
     config.init_config("tests/config/valid_config.yaml")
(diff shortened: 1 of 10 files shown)


onmete:
¯\\\_(ツ)\_/¯

It makes sense to some extent, but I'm a bit afraid the number of various rules and conventions is getting out of hand.
I certainly wouldn't be able to remember them all. So it would be up to you to keep it in mind in reviews.

/lgtm

bparees:
> It makes sense to some extent, but I'm a bit afraid the number of various rules and conventions is getting out of hand.
I certainly wouldn't be able to remember them all. So it would be up to you to keep it in mind in reviews.

+1, also seems to provide marginal value.  What benefit does one gain from this knowledge being encoded in the naming convention?  I generally don't call a function w/o looking at its definition to know what args it takes and what it returns, at a minimum.

bparees:
doesn't touch e2e tests so we can merge this w/o passing e2e-cluster

PR #731, opened 2024-04-09
Correct L2 score range info

asamal4:
As L2 is a distance metric, the value can be > 1. Faiss L2 index gives squared L2 as score. This is not normalized between 0 to 1.

Currently we are still using bge embedding model (for RAG), so it is not very evident but once we move to sentence-tranformers embedding model, we can see the value > 1

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -79,7 +79,7 @@ RAG_CONTENT_LIMIT = 1
 # And if we set a very low value, then there is risk of discarding all the chunks,
 # as there won't be perfect matching chunk. This also depends on embedding model
 # used during index creation/retrieval.
-# Range: 0 to 1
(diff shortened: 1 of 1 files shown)


ols/constants.py
```diff
@@ -79,7 +79,7 @@ class QueryValidationMethod(StrEnum):
 # And if we set a very low value, then there is risk of discarding all the chunks,
 # as there won't be perfect matching chunk. This also depends on embedding model
 # used during index creation/retrieval.
-# Range: 0 to 1
+# Range: +ve float value (can be > 1)
```
tisnik, line 82: what `+ve` mean please?

bparees: please just rewrite it as `positive`

PR #748, opened 2024-04-12
OLS-434: Use cluster url for evaluation

asamal4:
## Description

Option to use api url retrieved from cluster for response evaluation.

This is required as weekly job for evaluation run against api deployed in a cluster (https://github.com/openshift/release/pull/50793)

### DIFF
--- a/scripts/validate_response.py
+++ b/scripts/validate_response.py
@@ -6,13 +6,15 @@ import os
 import sys
 from collections import defaultdict
 
-import requests
 from llama_index.embeddings.huggingface import HuggingFaceEmbedding
 from pandas import DataFrame
 from scipy.spatial.distance import cosine, euclidean
 
 sys.path.append(os.path.dirname(os.path.dirname(__file__)))
 from ols.constants import NO_RAG_CONTENT_RESP
+from tests.e2e.cluster_utils import create_user, get_user_token, grant_sa_user_access
+from tests.e2e.constants import LLM_REST_API_TIMEOUT
+from tests.e2e.helper_utils import get_http_client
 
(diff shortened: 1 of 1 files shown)


scripts/validate_response.py
```diff
@@ -6,13 +6,15 @@
...
-import requests
 from llama_index.embeddings.huggingface import HuggingFaceEmbedding
 from pandas import DataFrame
 from scipy.spatial.distance import cosine, euclidean
 
 sys.path.append(os.path.dirname(os.path.dirname(__file__)))
 from ols.constants import NO_RAG_CONTENT_RESP
+from tests.e2e.cluster_utils import create_user, get_user_token, grant_sa_user_access
+from tests.e2e.constants import LLM_REST_API_TIMEOUT
+from tests.e2e.helper_utils import get_http_client
```
tisnik, line 17: it works today, but in future it might lead to circular imports

scripts/validate_response.py
```diff
@@ -97,12 +99,15 @@ def get_response_quality(self, args, qa_pairs):
...
-                # API question validator can be disabled.
-                "http://localhost:8080/v1/query",
+            response = api_client.post(
+                "/v1/query",
                 json={"query": question},
-                timeout=90,
-            ).json()["response"]
+                timeout=LLM_REST_API_TIMEOUT,
+            )
+            if response.status_code != 200:
```
tisnik, line 107: `requests` contains all HTTP code if you would like to avoid using magic constants

tisnik (commented):
it seems ok, just two nitpicks

PR #751, opened 2024-04-15
OLS-508: New suite with collector

onmete:
## Description

The e2e scripts update to run new marker with collector sidecar.

## Type of change

- [x] Refactor
- [x] New feature

## Prerequisite
- [ ] https://github.com/openshift/lightspeed-service/pull/584

### DIFF
--- a/tests/config/cluster_install/ols_manifests_with_collector.yaml
+++ b/tests/config/cluster_install/ols_manifests_with_collector.yaml
@@ -1,5 +1,6 @@
 # Env vars for use w/ envsubst:
 # $OLS_IMAGE - the pullspec for the OLS container image to deploy
+# $CP_OFFLINE_TOKEN - the "offline" token to console.stage.redhat.com
 
 ---
 apiVersion: apps/v1
@@ -58,6 +59,24 @@ spec:
             mountPath: /app-root/config/llmcreds
           - name: tls-certs
             mountPath: /app-root/certs
+          - name: ols-user-data
+            mountPath: /app-root/ols-user-data
+      - image: "$OLS_IMAGE"
+        name: ols-sidecar-user-data-collector
+        imagePullPolicy: Always
+        env:
+          - name: OLS_USER_DATA_PATH
+            value: /app-root/ols-user-data
+          - name: OLS_USER_DATA_COLLECTION_INTERVAL
+            value: "10"
+          - name: RUN_WITHOUT_INITIAL_WAIT
+            value: "true"
+          - name: INGRESS_ENV
+            value: "stage"
+          - name: CP_OFFLINE_TOKEN
+            value: "$CP_OFFLINE_TOKEN"
+        command: ["python3.11", "/app-root/ols/user_data_collection/data_collector.py"]
+        volumeMounts:
           - name: ols-user-data
             mountPath: /app-root/ols-user-data
       dnsPolicy: ClusterFirst
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+omit = 
+    ols/user_data_collection/data_collector.py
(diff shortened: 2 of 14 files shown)


tests/config/cluster_install/ols_manifests_with_collector.yaml
bparees, line 1: what's the motivation to introduce a separate test configuration for this rather than always running this way during testing?

bparees (commented):
is there a reason you don't think we should just always configure the sidecar/collector during the e2e runs, and run these tests against it?

i'd rather not introduce a new unique configuration if we don't need it.

tests/config/cluster_install/ols_manifests_with_collector.yaml
onmete: Sidecar with the collector is influencing the results of other tests - mainly the feedback/transcript ones. As we check the files being stored in those tests, the collector can package/send, and remove them during the test run. Resulting in random failures of these tests (depending on how they hit the collection interval).

See https://issues.redhat.com/browse/OLS-508

bparees: if the main issue is that we can't control when the upload/deletion occurs, i wonder if we can introduce an alternate mechanism to trigger the upload/deletion, such as touching a file on disk (that the upload script watches for).

then when under test we could just set the upload interval very high and directly control/trigger the uploading.

or perhaps even better, introduce a "don't collect/delete" marker file.  If that file is present, it disables the uploading/deletion until the file is removed.  That way you can "pause" the upload/deletion, run your test logic, and then unpause and confirm the upload happens.

onmete: Both options present an alternate way of how the collector is configured outside the deployment (envs/config).

Is there an issue with introducing a new e2e configuration I don't see?

bparees: It fragments our test coverage.  Now these new tests are only being run in the BAM provider environment(as you currently have it written) instead of all of them.  If we introduce a new environment we need to think about whether these tests need to run against them or not.  (I acknowledge in this case there's no reason to think these tests need to run against all our providers, but in general it's best not to rely on such assumptions.  We want to know that all our functionality works in all the environments we expect to run with)

there's also setup time involved for setting up ols w/ each configuration(waiting for OLS to redeploy), and there's complexity in reasoning about "when is a given e2e test going to be run", as seen in the new marker/tag logic you added.  In general it's not a pattern i want to encourage, it's more of a last resort.  In general i want less logic in our test-runner scripts, not more.  I approved some complexity Asutosh made only because it wasn't related to the actual e2e execution path.

I agree that introducing a "side" method of configuring/controlling collector behavior has its own drawbacks, but i think in this case it's an overall win that will make it easier for to test the functionality and be able to run those tests against OLS w/o having to worry about timing/race conditions.

onmete: I don't like to introduce a way how to configure/disable the data collection just to make the e2e easier. But ok. Once the base PR is merged, I'll make the changes here.

bparees: thank you.  I understand and appreciate your reluctance, like I said i don't love it either, but in this case the tradeoff feels worth it to me, unless we can come up with a third option.

I think the other compromise i'd accept would be putting logic into our pytest modules themselves that did the reconfigure/redeploy of OLS.  Ultimately that would be my goal even for the existing "suite" logic in bash:  move it into the test code itself rather than bash logic.

The idea would be to group tests that need a certain configuration together and have setup logic that deploys OLS w/ the needed configuration, runs those tests, and then undeploys OLS.  Then setup for the next group of tests, etc.

But that feels like an even larger undertaking that needs to be done carefully to give us a good foundation we can build on, so i'm not sure now is the time to take on that exercise.

onmete:
Closed in favor of https://github.com/openshift/lightspeed-service/pull/789

PR #769, opened 2024-04-17
Use freeform prompt template for conversation

asamal4:
## Description

Use freeform text as prompt. This is done to avoid ChatPromptTemplate, as granite was adding role tag to the response.. This also gives us more control how we want to modify prompt and add model specific prompt.
Currently the prompt has few granite specific tags.. GPT still works well with this prompt..

This is as per the experimentation done in [OLS-452](https://issues.redhat.com/browse/OLS-452)

Possible future works:

- As we are not using chat template anymore, we can think of alternate/more generic way of saving history.
- Model specific prompts.

### DIFF
--- a/ols/src/prompts/prompt_generator.py
+++ b/ols/src/prompts/prompt_generator.py
@@ -1,22 +1,15 @@
 """Prompt generator based on model / context."""
 
 from collections import namedtuple
-from typing import TYPE_CHECKING, Any, Optional
+from typing import Any, Optional
 
-if TYPE_CHECKING:
-    from langchain_core.prompts.chat import MessageLike
-
-from langchain.prompts import (
-    ChatPromptTemplate,
-    HumanMessagePromptTemplate,
-    MessagesPlaceholder,
-    SystemMessagePromptTemplate,
-)
+from langchain.prompts import PromptTemplate
 from langchain_core.messages.base import BaseMessage
 
 from ols.constants import (
     GPT35_TURBO,
     GRANITE_13B_CHAT_V2,
+    NO_RAG_CONTENT_RESP,
     PROVIDER_AZURE_OPENAI,
     PROVIDER_BAM,
     PROVIDER_OPENAI,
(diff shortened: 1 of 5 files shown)


ols/src/prompts/prompt_generator.py
```diff
@@ -88,32 +58,35 @@ def generate_prompt(
...
+            + conversation.content.strip().replace(NO_RAG_CONTENT_RESP, "")
+            for conversation in history
+            if conversation
+        ]
+        llm_input_values["chat_history"] = "\n".join(formatted_history)
+
+    # if len(rag_context) > 0:
+    #     prompt += CONTEXT_PLACEHOLDER
+    # if len(history) > 0:
+    #     prompt += HISTORY_PLACEHOLDER
```
bparees, line 88: delete?

ols/src/prompts/prompts.py
```diff
@@ -6,18 +6,49 @@
...
+# Note::
+# Right now templates are somewhat alligned to make granite work better.
+# GPT still works well with this. Ideally we should have model specific tags.
+# For history we can laverage ChatPromptTemplate from langchain,
+# but that is not done as granite was adding role tags like `Human:` in the response.
+# With PromptTemplate, we have more control how we want to structure the prompt.
+
+QUERY_SYSTEM_INSTRUCTION = """
+You are an assistant for question-answering tasks \
+related to the openshift and kubernetes container orchestration platforms.
```
bparees, line 19: these system instructions seem a bit brief....is there more we should consider adding here?  telling the model to be polite, not to make up answers, only give answers it's confident in, explicitly tell it not to try to answer non-openshift questions, etc?

(i know you just copied the existing instructions, but we've never really played w/ this at all)

ols/src/prompts/prompts.py
```diff
@@ -6,18 +6,49 @@
...
+
+USE_HISTORY_INSTRUCTION = """
+Use the previous chat history to interact and help the user.
+"""
+
+HISTORY_PLACEHOLDER = """
+
+[HISTORY]
+{chat_history}
+[END]
```
bparees, line 43: there are considered "generic" marker tags? (they won't confuse/break some models?)

ols/src/prompts/prompts.py
```diff
@@ -6,18 +6,49 @@
...
+
+"""
+
+QUERY_PLACEHOLDER = """
+<|user|>
+{query}
+<|assistant|>
```
bparees, line 50: same for these....they are considered to be generic marker tags?  aren't likely to confuse some models?

bparees:
some nits and questions, but general concept seems sound.

/approve

PR #789, opened 2024-04-22
OLS-508: Ability to control collector via file existence

onmete:
## Description

Ability to control collector via file existence

## Type of change

- [x] New feature

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -1169,6 +1169,10 @@ def test_user_data_collection():
         cluster_utils.remove_dir(pod_name, OLS_USER_DATA_PATH + "/transcripts")
         assert cluster_utils.list_path(pod_name, OLS_USER_DATA_PATH) == []
 
+    # safety wait for the script to start after beying disabled by other
+    # tests
+    time.sleep(OLS_USER_DATA_COLLECTION_INTERVAL + 1)
+
     # data shoud be pruned now and this is the point from which we want
     # to check the logs
     container_log = cluster_utils.get_container_log(
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -55,6 +55,7 @@ metrics_client = None
 # constant from tests/config/cluster_install/ols_manifests.yaml
 OLS_USER_DATA_PATH = "/app-root/ols-user-data"
 OLS_USER_DATA_COLLECTION_INTERVAL = 10
+OLS_COLLECTOR_DISABLING_FILE = OLS_USER_DATA_PATH + "/disable_collector"
 
 
 def setup_module(module):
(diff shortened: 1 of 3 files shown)


tests/e2e/test_api.py
```diff
@@ -804,42 +805,50 @@ def test_feedback_storing_standalone():
...
+    feedback_data = cluster_utils.get_single_existing_feedback(pod_name, feedbacks_path)
+
+    assert feedback_data["user_id"]  # we don't care about actual value
+    assert feedback_data["conversation_id"] == CONVERSATION_ID
+    assert feedback_data["user_question"] == "what is OCP4?"
+    assert feedback_data["llm_response"] == "Openshift 4 is ..."
+    assert feedback_data["sentiment"] == 1
+
+    # ensure script is enabled again after successful test
+    cluster_utils.remove_file(pod_name, OLS_COLLECTOR_DISABLING_FILE)
```
bparees, line 848: you need to do this in a finally block because otherwise if any of the asserts fail, this won't get executed.

https://stackoverflow.com/questions/56938028/execute-code-after-an-assert-in-case-of-assertionerror

tests/e2e/test_api.py
```diff
@@ -973,43 +982,51 @@ def test_transcripts_storing_standalone():
...
+    # we don't want llm response influence this test
+    assert "query_is_valid" in transcript
+    assert "llm_response" in transcript
+    assert "referenced_documents" in transcript
+    assert transcript["referenced_documents"][0]["docs_url"]
+    assert transcript["referenced_documents"][0]["title"]
+    assert "truncated" in transcript
+
+    # ensure script is enabled again after successful test
+    cluster_utils.remove_file(pod_name, OLS_COLLECTOR_DISABLING_FILE)
```
bparees, line 1026: finally block

tests/e2e/test_api.py
```diff
@@ -1152,6 +1169,10 @@ def get_last_log_line(logs: str) -> str:
         cluster_utils.remove_dir(pod_name, OLS_USER_DATA_PATH + "/transcripts")
         assert cluster_utils.list_path(pod_name, OLS_USER_DATA_PATH) == []
 
+    # safety wait for the script to start after beying disabled by other
```
bparees, line 1172:
```suggestion
    # safety wait for the script to start after being disabled by other
```

tests/e2e/test_api.py
```diff
@@ -804,42 +805,53 @@ def test_feedback_storing_standalone():
...
+        assert feedback_data["user_id"]  # we don't care about actual value
+        assert feedback_data["conversation_id"] == CONVERSATION_ID
+        assert feedback_data["user_question"] == "what is OCP4?"
+        assert feedback_data["llm_response"] == "Openshift 4 is ..."
+        assert feedback_data["sentiment"] == 1
+
+    finally:
+        # ensure script is enabled again after test (succesfull or not)
+        cluster_utils.remove_file(pod_name, OLS_COLLECTOR_DISABLING_FILE)
+        assert "disable_collector" not in cluster_utils.list_path(
```
bparees, line 852: @tisnik do you know if having a passing assert execute after a failing assert, will result in the test being reported as a pass?

i'm wondering if one of the asserts above fails, so we jump into the finally block, and then this assert passes, that we'll report the overall test as a pass.  I would hope not, but i feel like the question has to be asked :)

bparees: I assume the original exception that resulted from the failed assert still ends up getting thrown and thus pytest records the failure

PR #790, opened 2024-04-22
Add sentiment field validation for feedback model

onmete:
## Description

Something I noticed during Pavel's demo. We don't have an explicit check for the value. We allow either good(1) or bad(-1).

## Type of change

- [x] New feature

### DIFF
--- a/ols/app/models/models.py
+++ b/ols/app/models/models.py
@@ -269,7 +269,7 @@ class FeedbackRequest(BaseModel):
                     "user_question": "foo",
                     "llm_response": "bar",
                     "user_feedback": "Great service!",
-                    "sentiment": 5,
+                    "sentiment": 1,
                 }
             ]
         }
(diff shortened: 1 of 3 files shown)


ols/app/models/models.py
```diff
@@ -283,6 +283,14 @@ def check_uuid(cls, value: str) -> str:
             raise ValueError(f"Improper conversation ID {value}")
         return value
 
+    @field_validator("sentiment")
+    @classmethod
+    def check_sentiment(cls, value: int) -> int:
```
tisnik, line 288: is `None` allowed?

tisnik: so it's `Optional[int]` -> `Optional[int]`?

tisnik: I mean, just specify that the member in model is optional, the parameter is optional, and indeed, return value is optional too ;)

onmete: Ok, you are talking about typing. Updated.

PR #795, opened 2024-04-23
Collect logs from all containers in artifacts

onmete:
## Description

Collect logs from all containers in artifacts

## Type of change

- [x] New feature


tisnik (commented):
looks good

PR #802, opened 2024-04-23
Context window size can be specified for provider + model

tisnik:
## Description

Context window size can be specified for provider + model
When provider+model is not known, the default size will be used.
(Please note that Azure OpenAI != OpenAI now!)

## Type of change

- [ ] Refactor
- [x] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -55,14 +55,26 @@ DEFAULT_CONTEXT_WINDOW_SIZE = 8000
 DEFAULT_RESPONSE_TOKEN_LIMIT = 500
 MINIMUM_CONTEXT_TOKEN_LIMIT = 1
 
-# Model-specific context window size
+# Provider and Model-specific context window size
 # see https://platform.openai.com/docs/models/gpt-4-turbo-and-gpt-4
 # and https://www.ibm.com/docs/en/cloud-paks/cp-data/4.8.x?topic=models-supported-foundation
 CONTEXT_WINDOW_SIZES = {
-    GRANITE_13B_CHAT_V1: 8192,
-    GRANITE_13B_CHAT_V2: 8192,
-    GPT4_TURBO: 128000,
-    GPT35_TURBO: 16385,
+    PROVIDER_BAM: {
+        GRANITE_13B_CHAT_V1: 8192,
+        GRANITE_13B_CHAT_V2: 8192,
+    },
+    PROVIDER_WATSONX: {
+        GRANITE_13B_CHAT_V1: 8192,
+        GRANITE_13B_CHAT_V2: 8192,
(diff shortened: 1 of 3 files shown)


ols/constants.py
```diff
@@ -55,14 +55,26 @@ class QueryValidationMethod(StrEnum):
...
+    PROVIDER_BAM: {
+        GRANITE_13B_CHAT_V1: 8192,
+        GRANITE_13B_CHAT_V2: 8192,
+    },
+    PROVIDER_WATSONX: {
+        GRANITE_13B_CHAT_V1: 8192,
+        GRANITE_13B_CHAT_V2: 8192,
+    },
+    PROVIDER_AZURE_OPENAI: {
+        GPT4_TURBO: 128000,
```
bparees, line 71: this value also isn't correct for azure, here's the full set of limits:

https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#model-summary-table-and-region-availability

image

bparees: unfortunately the "model version" (e.g. "vision-preview" vs "0125-preview", etc) appears to be something chosen by the user in their azure openai instance setup, not an actual "model id"

see: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/working-with-models?tabs=powershell

i assume @thoraxe picked a model version when he setup our instance

tisnik: hmm what do you expect there? The table seems to say "128000"

bparees: the table says anywhere from 8k to 128k depending on which gpt4 "version" is being used.

i guess for now we can assume people are using one of the versions that allows 128k (and our docs should steer them to pick one of those when doing an install) but longer term this will need some more thought.

bparees: 3.5 also appears to vary, it sounds like if @thoraxe updates our model version we can get the 16k context limit instead.

ols/constants.py
```diff
@@ -55,14 +55,26 @@ class QueryValidationMethod(StrEnum):
...
+        GPT35_TURBO: 8192,
```
bparees, line 72: since the newer azure openai gpt35 model versions support 16k, let's go ahead and set this to 16k for now, and work with @thoraxe to get our particular azure openai model instance updated to a new version ("0125" is the latest)

ols/constants.py
tisnik: ah ok, I was under impression that "our" `GPT4_TURBO` equals to the last three rows, which all have the same 128000  limit

PR #803, opened 2024-04-23
Use consistent default values for truncation

asamal4:
## Description

Use consistent value for default max response token/context window used by truncation logic.

- Related Issue # [OLS-537](https://issues.redhat.com//browse/OLS-537)


tisnik (commented):
it's ok, but in practice it won't be used: https://github.com/openshift/lightspeed-service/pull/802

PR #808, opened 2024-04-24
Add buffer to token count

asamal4:
## Description
We are using precise context window now and token count is currently an approximation. So adding buffer to token count.
With this change we will increase token count by some weight.

Currently a single constant is used, but this can be extended further to have model/provider specific weights/percentage. Experimentation is required to understand in general how far off is our estimation.

Alternatively we could decrease context window by some buffer. But decided against that as error message may be confusing. Context window is well know and token calculation is an estimation anyway. So modifying count is more reasonable.

- Related Issue # [OLS-537](https://issues.redhat.com//browse/OLS-537)

cc: @bparees

### DIFF
--- a/ols/utils/token_handler.py
+++ b/ols/utils/token_handler.py
@@ -2,6 +2,7 @@
 
 import logging
 from collections import defaultdict
+from math import ceil
 
 from langchain_core.messages.base import BaseMessage
 from llama_index.schema import NodeWithScore
@@ -12,6 +13,7 @@ from ols.constants import (
     DEFAULT_TOKENIZER_MODEL,
     MINIMUM_CONTEXT_TOKEN_LIMIT,
     RAG_SIMILARITY_CUTOFF_L2,
+    TOKEN_BUFFER_WEIGHT,
 )
 
 logger = logging.getLogger(__name__)
@@ -31,6 +33,8 @@ class TokenHandler:
 
     def __init__(self, encoding_name: str = DEFAULT_TOKENIZER_MODEL) -> None:
         """Initialize the class instance."""
+        # Note: We need an approximate tokens count.
+        # For different models, exact tokens may vary due to different tokenizer.
         self._encoder = get_encoding(encoding_name)
 
     def text_to_tokens(self, text: str) -> list[int]:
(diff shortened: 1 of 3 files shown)


ols/utils/token_handler.py
```diff
@@ -92,7 +102,7 @@ def get_available_tokens(self, prompt: str, model_config: ModelConfig) -> int:
             f"Response token limit: {response_token_limit}"
         )
 
-        prompt_token_count = len(self.text_to_tokens(prompt))
+        prompt_token_count = self._get_token_count(self.text_to_tokens(prompt))
```
tisnik, line 105: it is static method, don't have to be called via self

ols/utils/token_handler.py
```diff
@@ -140,7 +150,7 @@ def truncate_rag_context(
                 break
 
             tokens = self.text_to_tokens(node.get_text())
-            tokens_count = len(tokens)
+            tokens_count = self._get_token_count(tokens)
```
tisnik, line 153: dtto, staticmethod

tisnik (commented):
commented

ols/utils/token_handler.py
```diff
@@ -159,22 +169,22 @@ def truncate_rag_context(
...
         for message in reversed(history):
-            message_length = len(token_handler_obj.message_to_tokens(message))
+            message_length = TokenHandler._get_token_count(
+                self.message_to_tokens(message)
+            )
             total_length += message_length
             # if total length of already checked messages is higher than limit
             # then skip all remaining messages (we need to skip from top)
             if total_length > limit:
+                logger.debug("History truncated, it exceeds the limit.")
```
bparees, line 187: should print the length and the limit here to be truly useful for debugging

tests/unit/utils/test_token_handler.py
```diff
@@ -95,7 +97,7 @@ def test_available_tokens_for_regular_prompt(self):
             - model_config.response_token_limit
             - prompt_length
         )
-        assert available_tokens == expected_value
+        assert available_tokens == expected_value - ceil(prompt_length * 0.05)
```
bparees, line 100: not sure why you did this here instead of when assigning `expected_value` ?

tests/unit/utils/test_token_handler.py
```diff
@@ -133,8 +136,9 @@ def test_available_tokens_specific_model_config(self):
...
```
bparees, line 139: same here, why not set the expected_value correctly in the first place?

bparees (commented):
some nits but generally lgtm

ols/utils/token_handler.py
bparees: yeah i was wondering about the value of a static method for this, over just making it a standalone utility function.  But i didn't want to bikeshed :)

PR #810, opened 2024-04-24
OLS-538: Check final token count before reaching LLM

tisnik:
## Description

Check final token count before reaching LLM
That should avoid "Internal server error"

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-538](https://issues.redhat.com//browse/OLS-538)
- Closes #

## Testing

### DIFF
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -121,6 +121,13 @@ class DocsSummarizer(QueryHelper):
             history,
             rag_context,
         )
+
+        # Final check if we don't use more tokens than permitted by provider+model
+        # Handles: OLS-538
+        token_handler.get_available_tokens(
+            final_prompt.format(**llm_input_values), model_config
+        )
+
         chat_engine = LLMChain(
             llm=bare_llm,
             prompt=final_prompt,


ols/src/query_helpers/docs_summarizer.py
```diff
@@ -123,6 +124,13 @@ def summarize(
             history,
             rag_context,
         )
+
+        # Final check if we don't use more tokens than permitted by provider+model
+        # Handles: OLS-538
+        self.check_token_limit(
+            token_handler, final_prompt, llm_input_values, model_config
+        )
```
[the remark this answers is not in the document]
tisnik: ah nice, so it will be super easy in the end :-)

bparees:
> But later we can include role or any additional character during truncation itself.. and we will achieve below..

@asamal4 do we have a story for implementing that change?

bparees:
/approve

but you broke a unit test (and we should have an e2e test for this behavior)

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -123,6 +123,13 @@ def summarize(
             history,
             rag_context,
         )
+
+        # Final check if we don't use more tokens than permitted by provider+model
+        # Handles: OLS-538
+        token_handler.get_available_tokens(
+            final_prompt.format(**llm_input_values), model_config
+        )
```
bparees, line 131: does question_validator have a similar check?  Since the prompt might go there before it ever gets to docs_summarizer

tisnik: on line 64, but it does not catch all issues with total length (no history nor rag)

bparees: > on line 64, but it does not catch all issues with total length (no history nor rag)

sure, i just wanted to be sure that if the prompt itself (w/o rag+history) is too long, the question validator isn't going to blow up, but rather will also return a clean error.

i don't see that check on line 64 of question_validator currently, but perhaps you're referring to an unmerged pr?

tisnik: it's this line
```
        TokenHandler().get_available_tokens(query, model_config)
```

the function has horrible name as it does two things:
- returns available tokens if > 0
- throws an exception instead

so probably better name would be check_and... or ensure_... or we can split into two functions, but probably it is nitpicking right now

bparees: yes it's a bit surprising/confusing that "get_available_tokens" is intended to raise an exception.  Even more so confusing because validate_question (the function that's calling get_available_tokens) isn't attempting to catch that exception.  So you're relying on the caller of validate_question to know that validate_question is calling a "get" function that may throw an exception, and the caller of validate_question is expected to handle that exception.

too many layers of indirection.

bparees: (but i'm glad that the check is being done in validate_question in addition to docs_summarizer)

tisnik:
> /approve
> 
> but you broke a unit test (and we should have an e2e test for this behavior)

yup broke, it was on purpose. With @asamal4 's PR, it is fixed now

bparees:
/lgtm

concerns about https://github.com/openshift/lightspeed-service/pull/810/files#r1581220802 can be addressed in a refactor followup

ols/src/query_helpers/docs_summarizer.py
tisnik: yup, let's create function `ensure_available_tokens(available_tokens, needed_tokens=0)` ATM the `needed_tokens` wont be needed (sic!), but in future might be useful to check against some minimal required value, and not just with 0.

PR #820, opened 2024-04-25
consider role for history tokens

asamal4:
## Description

Include role and new line for history token calculation.
This will make sure that final token count is not beyond context window.

[OLS-548](https://issues.redhat.com/browse/OLS-548)

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -184,8 +184,14 @@ def generate_response(
         docs_summarizer = DocsSummarizer(
             provider=llm_request.provider, model=llm_request.model
         )
+        # TODO: OLS-551 Store history in more readable format
+        history = [
+            conversation.type + ": " + conversation.content.strip()
+            for conversation in previous_input
+            if conversation
+        ]
         llm_response = docs_summarizer.summarize(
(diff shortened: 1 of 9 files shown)


ols/app/endpoints/ols.py
```diff
@@ -184,8 +184,14 @@ def generate_response(
         docs_summarizer = DocsSummarizer(
             provider=llm_request.provider, model=llm_request.model
         )
+        # Note: This may change, if we change how we store history currently.
```
tisnik, line 187: I'd probably refactor it, or defer this work to `DocsSummarizer`. Then the change in API logic will be minimal

tisnik (commented):
it overall it's in sync with our aggreement. I'd probably keep `app/endpoints/ols.py` intact and do everything in `DocsSummarizer` (as it is internal work for this module/class). WDYT?

ols/app/endpoints/ols.py
tisnik: well yes, in the end it will make the cache content "readable". I'm just thinking there we might focus on short-term solution, to fix the token limit problem first. Then (after internal release+EA) we'll have time to refactor everything else.

tisnik:
it looks ok to me

PR #823, opened 2024-04-26
OLS-506: must-gather in tests

onmete:
## Description

Move after tests must-gather logic to python tests

## Type of change

- [x] Refactor
- [x] New feature

### DIFF
--- a/tests/e2e/cluster_utils.py
+++ b/tests/e2e/cluster_utils.py
@@ -14,6 +14,19 @@ def run_oc(args: list[str]) -> subprocess.CompletedProcess:
     )
 
 
+def run_oc_and_store_stdout(
+    args: list[str], stdout_file: str
+) -> subprocess.CompletedProcess:
+    """Run a command in the OpenShift cluster and store the stdout in a file."""
+    try:
+        result = run_oc(args)
+        assert result.returncode == 0
+        with open(stdout_file, "w") as fout:
+            fout.write(result.stdout)
+    except subprocess.CalledProcessError as e:
+        raise Exception("Error running oc command") from e
+
+
 def create_user(name: str) -> None:
     """Create a service account user for testing."""
     try:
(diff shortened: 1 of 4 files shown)


onmete:
The original must_gather from utils.sh is still in place as it is used in `run_suite` in case the ols fails to become available. I assume the function will be completely removed when we refactor the rest of the e2e script.

tests/e2e/cluster_utils.py
```diff
@@ -74,20 +87,34 @@ def get_ols_url(route_name: str) -> str:
         raise Exception("Error getting route hostname") from e
 
 
+def get_pods() -> list[str]:
+    """Get the names of all pods in the cluster."""
```
bparees, line 91: this should take a namespace arg and pass it to `oc`, otherwise you don't actually know which namespace you're getting the pods for (it will just be whatever the current NS happens to be)

you can default the arg value to `openshift-lightspeed`

tests/e2e/cluster_utils.py
```diff
@@ -74,20 +87,34 @@ def get_ols_url(route_name: str) -> str:
...
+    try:
+        result = run_oc(["get", "pods", "-o", "jsonpath='{.items[*].metadata.name}'"])
+        return result.stdout.strip("'").split()
+    except subprocess.CalledProcessError as e:
+        raise Exception("Error getting pods") from e
+
+
 def get_single_existing_pod_name() -> str:
     """Return name of the single pod that is in the cluster."""
```
bparees, line 100: this should take a NS arg also.  you can default it to `openshift-lightspeed` but it should be passed to oc.

tests/e2e/test_api.py
```diff
@@ -85,8 +86,45 @@ def setup_module(module):
...
-    # TODO: OLS-506 Move the program logic to gather cluster artifacts from utils.sh into e2e module
-    pass
+    if on_cluster:
+        must_gather()
+
+
+def must_gather():
+    """Collect logs and state from openshift-lightspeed namespace."""
+    artifact_dir = os.environ.get(
+        "ARTIFACT_DIR", "/home/ometelka/projects/lightspeed-service/ci-testic"
```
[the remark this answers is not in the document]
onmete: dev leftover

tests/e2e/test_api.py
```diff
@@ -85,8 +86,45 @@ def setup_module(module):
...
+    )
+    suite_id = os.environ.get("SUITE_ID", "blabol")
```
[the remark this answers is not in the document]
onmete: dev leftover

tests/scripts/utils.sh
```diff
@@ -126,13 +126,12 @@ function run_suite() {
...
   if [ -z ${RESPONSE_EVALUATION:-} ]; then  
-    SUITE_ID=$1 TEST_TAGS=$2 MODEL=$8 make test-e2e
+    SUITE_ID=$1 TEST_TAGS=$2 MODEL=$8 ARTIFACT_DIR=$ARTIFACT_DIR make test-e2e
   else
     export SCENARIO="${SCENARIO:-with_rag}"
     MODEL=$8 SCENARIO=$SCENARIO make response-sanity-check
   fi
 
   local rc=$?
-  must_gather $1
```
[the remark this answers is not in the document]
onmete: removed -> I had to move the logic from test to separate script to be able to trigger it separately

tests/e2e/cluster_utils.py
```diff
@@ -74,22 +87,45 @@ def get_ols_url(route_name: str) -> str:
...
+    """Return name of the single pod that is in the cluster."""
+    try:
+        result = get_pods(namespace)
+        assert len(result) == 1
+        return result[0]
     except subprocess.CalledProcessError as e:
         raise Exception("Error getting pod name") from e
 
 
+def get_pod_containers(pod) -> list[str]:
```
bparees, line 118: this one should take a NS too

tests/scripts/utils.sh
```diff
@@ -113,26 +100,30 @@ function run_suite() {
...
+  return 1
+
   # wait for the ols api server to come up
   wait_for_ols "$OLS_URL"
   if [ $? -ne 0 ]; then
     echo "Timed out waiting for OLS to become available"
     echo "${JUNIT_TEMPLATE}" | sed "s/SUITE_ID/${1}/g" > $ARTIFACT_DIR/junit_setup_${1}.xml
 
-    must_gather $1
+    ARTIFACT_DIR=$ARTIFACT_DIR SUITE_ID=$1 python tests/scripts/must_gather.py
```
bparees, line 114: what about moving wait_for_ols into a pytest setup step and then we don't need this separate invocation of must_gather because the teardown will cover it?

would just need to ensure the setup step records a junit failure if it times out (i'm not sure if that's possible from pytest)

onmete: Should be possible via `unittest`. As it's beyond the scope of OLS-506, I've created a [follow-up story](https://issues.redhat.com/browse/OLS-581) for that.

PR #824, opened 2024-04-26
Add init to scripts package

onmete:
## Description

Add init to scripts package. As we are importing from scripts in our app, this would be a proper approach.

## Type of change

- [x] Refactor


onmete:
closing in favor of https://github.com/openshift/lightspeed-service/pull/825

PR #825, opened 2024-04-26
OLS-506: Move testing script ot tests

onmete:
## Description

Move validation script to tests. It is only used there. It is causing some problems with imports.

## Type of change

- [x] Refactor

### DIFF
--- a/tests/scripts/validate_response.py
+++ b/tests/scripts/validate_response.py
@@ -147,7 +147,7 @@ def main():
     result_df = ResponseValidation().get_response_quality(args, qa_pairs, api_client)
 
     if len(result_df) > 0:
-        result_dir = "test_results"
+        result_dir = "tests/test_results"
         os.makedirs(result_dir, exist_ok=True)
         result_file = (
             f"{result_dir}/question_answer_result_{args.scenario}_{args.model}.csv"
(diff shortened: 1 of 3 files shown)


tests/scripts/validate_response.py
```diff
@@ -5,15 +5,21 @@
...
 from llama_index.embeddings.huggingface import HuggingFaceEmbedding
 from pandas import DataFrame
 from scipy.spatial.distance import cosine, euclidean
 
-sys.path.append(os.path.dirname(os.path.dirname(__file__)))
-from tests.e2e.cluster_utils import create_user, get_user_token, grant_sa_user_access
-from tests.e2e.constants import LLM_REST_API_TIMEOUT
-from tests.e2e.helper_utils import get_http_client
+project_root = Path(__file__).resolve().parent.parent.parent.as_posix()
+sys.path.append(project_root)
```
bparees, line 15: is there actually a reason this code has to be run before the imports?  @asamal4 ?

bparees (commented):
i'm fine w/ the file move, questions about some of the code itself

tests/scripts/validate_response.py
tisnik: IMHO now it is not needed. It was there because of "cross-package" imports (and it's bit dirty hack ;)

onmete:
@asamal4 Path import removed and makefile updated.

PR #826, opened 2024-04-26
OLS-506: Remove unnecessary "import as" in test_api.py

onmete:
## Description

Remove unnecessary "import as" in test_api.py

## Type of change

- [x] Refactor


tisnik (commented):
lgtm

just: what `helper_utils` means ;) that's the classic catch-all name

PR #828, opened 2024-04-26
OLS-527: bump up libs

asamal4:
bump openai from 1.23.3 to 1.23.6
bump ibm-watson-machine-learning from 1.0.355 to 1.0.356

skipped below
redis from 5.0.4 to 5.0.3 (downgraded) -> as we are not using
llama-index from 0.9.39 to 0.10.31 -> code change is required, we have a separate story.


tisnik (commented):
looks good

PR #829, opened 2024-04-26
OLS-538: Unit test for too long prompt exception

tisnik:
## Description

[OLS-538](https://issues.redhat.com//browse/OLS-538): Unit test for too long prompt exception

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Unit test

- Related Issue #[OLS-538](https://issues.redhat.com//browse/OLS-538)
- Closes #

### DIFF
--- a/tests/unit/app/endpoints/test_ols.py
+++ b/tests/unit/app/endpoints/test_ols.py
@@ -17,6 +17,7 @@ from ols.app.models.models import LLMRequest, ReferencedDocument
 from ols.src.llms.llm_loader import LLMConfigurationError
 from ols.utils import config, suid
 from ols.utils.query_filter import QueryFilter, RegexFilter
+from ols.utils.token_handler import PromptTooLongError
 
 
 @pytest.fixture(scope="module")
@@ -171,6 +172,20 @@ def test_validate_question_valid_kw(llm_validate_question_mock, _load_config):
 
     assert resp
     assert llm_validate_question_mock.call_count == 0
+
+
+@patch(
+    "ols.src.query_helpers.question_validator.QuestionValidator.validate_question",
+    side_effect=PromptTooLongError("Prompt length 10000 exceeds LLM"),
+)
+def test_validate_question_too_long_query(llm_validate_question_mock, _load_config):
+    """Check the behaviour of validate_question function with too long query."""
+    conversation_id = suid.get_suid()
+    query = "Tell me about Kubernetes?"
+    llm_request = LLMRequest(query=query, conversation_id=conversation_id)
+    # PromptTooLongError should be caught and HTTPException needs to be raised
+    with pytest.raises(HTTPException, match="413: {'response': 'Prompt is too long'"):
+        ols.validate_question(conversation_id, llm_request)
 
 
 @patch(


tests/unit/app/endpoints/test_ols.py
```diff
@@ -173,6 +174,20 @@ def test_validate_question_valid_kw(llm_validate_question_mock, _load_config):
...
 
 
+@patch(
+    "ols.src.query_helpers.question_validator.QuestionValidator.validate_question",
+    side_effect=PromptTooLongError("Prompt length 10000 exceeds LLM"),
+)
+def test_validate_question_too_long_query(llm_validate_question_mock, _load_config):
+    """Check the behaviour of validate_question function with too long query."""
+    conversation_id = suid.get_suid()
+    query = "Tell me about Kubernetes?" * 10000
```
bparees, line 184: is this necessary?  you've already setup a side effect that validate_question is going to raise the exception regardless.

tests/unit/app/endpoints/test_ols.py
```diff
@@ -173,6 +174,20 @@ def test_validate_question_valid_kw(llm_validate_question_mock, _load_config):
...
+    llm_request = LLMRequest(query=query, conversation_id=conversation_id)
+    # PromptTooLongError should be caught and HTTPException needs to be raised
+    with pytest.raises(HTTPException, match="413: {'response': 'Prompt is too long'"):
+        ols.validate_question(conversation_id, llm_request)
```
bparees, line 188: what is this really testing?  you setup a mock such that the exception is guaranteed to be raised.  So at best this is just testing that ols.validate_question is catching and reraising that exception?

bparees (commented):
this test is useful, but not as critical to me as a test that validates that we detect and handle a too long prompt (by actually submitting a prompt that is too long, not by mocking raising an exception).

Also there are two cases to cover:

1) initial question that goes through question validation
2) follow up question because it won't go through question validation

In order to ensure that prompts that are too long get handled in both cases (one case where question validation is processing the prompt, the other in the case where docs summarizer is)

those would be e2e tests of course.

tests/unit/app/endpoints/test_ols.py
tisnik: just this, it's UT, nothing on integration nor e2e level now

tests/unit/app/endpoints/test_ols.py
tisnik: not right now

bparees: please remove it then, because as it is now it's misleading as to what the test is actually doing/testing.

tisnik: just did it :)

tests/unit/app/endpoints/test_ols.py
tisnik: for record, e2e waits for https://github.com/openshift/lightspeed-service/pull/820 and then https://github.com/openshift/lightspeed-service/pull/810 . when these two will be in, e2e should be created with real end-to-end behavior checking

bparees:
/approve
/lgtm

but i remain skeptical of the value of tests like this, or rather the value relative to the effort to write+maintain them.  Effectively this test just confirms that ols.validate_question catches an exception that is raised by question_validator.validate_question.  (and it's worth noting that question_validator doesn't even normally raise the exception itself, it's raised by a helper that question_validator calls.  So if the behavior of question_validator were changed to not call that helper, this test would continue to pass despite not serving any purpose.  That's somewhat a general limitation of unit tests, but somewhat a particular limitation of this specific test and its use of mocks)

PR #831, opened 2024-04-29
Fix issue OLS-553: process files with wrong paths gracefully

tisnik:
## Description

Fix issue [OLS-553](https://issues.redhat.com//browse/OLS-553): process files with wrong paths gracefully

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-553](https://issues.redhat.com//browse/OLS-553)

## Testing
- Run `pytest -v scripts/feedback_aggregator.py`

### DIFF
--- a/scripts/feedback_aggregator.py
+++ b/scripts/feedback_aggregator.py
@@ -253,6 +253,9 @@ def check_key(key: str) -> bool:
 def construct_filename(key: str) -> str:
     """Construct filename from key."""
     groups = key_match(key)
+    if groups is None:
+        raise Exception(f"Can not construct filename from key {key}")
+
     cluster_id = groups[1].lower()
     year = groups[2][:4]
     month = groups[2][4:]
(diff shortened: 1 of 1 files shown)


scripts/feedback_aggregator.py
```diff
@@ -468,6 +471,9 @@ def test_check_key(key):
     "archives/compressed/ff/01234567-89ab-cdef-0123-456789abcdef/2024Z4/08/073607.tar.gz",
     "archives/compressed/fZ/01234567-89ab-cdef-0123-456789abcdef/202404/08/073607.tar.gz",
     "archives/compressed/ff/01234567-89ab-cdZf-0123-456789abcdef/202404/08/073607.tar.gz",
+    "archives/compressed/00/00000000-0000-0000-0000-000000000001/20240/08/073607.tar.gz",
+    "archives/compressed/00/00000000-0000-0000-0000-000000000001/202404/8/073607.tar.gz",
+    "archives/compressed/00/00000000-0000-0000-0000-000000000001/202404/08/73607.tar.gz",
```
bparees, line 476: i'm missing what's invalid about this last key?

also, not related to this PR, but why is the filename format "YYYYMM/DD" instead of "YYYY/MM/DD"?

bparees: oh, it's invalid because the final time segment is 5, not 6, chars.

tisnik: last key: the time has not 6 digits
unrelated question: well it's how CCX service was configured at beginning because the original plan 4 years ago was that there will be daily batch processing and that number of objects in bucket is too big to store/compute diff between new objects and already processed objects. So IIRC it was preliminary optimization (and not a good one IMHO).

PR #839, opened 2024-04-30
OLS-503: Fine-tune system prompt

syedriko:
## Description

A better system prompt, based on EA feedback.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [x] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issues # https://issues.redhat.com/browse/OLS-572, https://issues.redhat.com/browse/OLS-545, https://issues.redhat.com/browse/OLS-528
- Closes # https://issues.redhat.com/browse/OLS-503

## Testing

### DIFF
--- a/ols/src/prompts/prompts.py
+++ b/ols/src/prompts/prompts.py
@@ -15,9 +15,20 @@ from ols.constants import SUBJECT_ALLOWED, SUBJECT_REJECTED
 # With PromptTemplate, we have more control how we want to structure the prompt.
 
 QUERY_SYSTEM_INSTRUCTION = """
-You are an assistant for question-answering tasks \
-related to the openshift and kubernetes container orchestration platforms.
+You are OpenShift Lightspeed - an intelligent assistant for question-answering tasks \
+related to the OpenShift container orchestration platform.
 
+Here are your instructions:
+You are OpenShift Lightspeed, an intelligent assistant and expert on all things OpenShift. \
+Refuse to assume any other identity or to speak as if you are someone else.
+If the context of the question is not clear, consider it to be OpenShift.
+Never include URLs in your replies.
+Refuse to answer questions or execute commands not about OpenShift.
+Do not mention your last update. You have the most recent information on OpenShift.
+
+Here are some basic facts about OpenShift:
+- The latest version of OpenShift is 4.15.
+- OpenShift is a distribution of Kubernetes. Everything Kubernetes can do, OpenShift can do and more.
 """
 
 USE_CONTEXT_INSTRUCTION = """
--- a/tests/config/valid_config.yaml
+++ b/tests/config/valid_config.yaml
@@ -8,8 +8,8 @@ llm_providers:
       - name: m1
         url: "https://murl1"
         credentials_path: tests/config/secret.txt
-        context_window_size: 200
-        response_token_limit: 50
+        context_window_size: 400
+        response_token_limit: 100
       - name: m2
         url: "https://murl2"
   - name: p2
--- a/tests/unit/utils/test_config.py
+++ b/tests/unit/utils/test_config.py
@@ -707,8 +707,8 @@ def test_valid_config_file():
                                 "name": "m1",
                                 "url": "https://murl1",
                                 "credentials_path": "tests/config/secret.txt",
-                                "context_window_size": 200,
-                                "response_token_limit": 50,
+                                "context_window_size": 400,
+                                "response_token_limit": 100,
                             },
                             {
                                 "name": "m2",


ols/src/prompts/prompts.py
```diff
@@ -15,9 +15,20 @@
...
+If the context of the question is not clear, consider it to be OpenShift.
+Never include URLs in your replies.
+Refuse to answer questions or execute commands not about OpenShift.
+Do not mention your last update. You have the most recent information on OpenShift.
+
+Here are some basic facts about OpenShift:
+- You are OpenShift Lightspeed, an intelligent assistant and expert on all things OpenShift.
+- The latest version of OpenShift is 4.15.
+- OpenShift is a distribution of Kubernetes. Everything Kubernetes can do, OpenShift can do, only better.
+- OpenShift supports Vertical Pod Autoscaler (VPA).
```
bparees, line 31: this one seems slightly odd.  The others are pretty generic, but this seems like an instruction added to address a specific question that OLS got wrong.  That's not gonna scale so i'm not inclined to start including such fine grained bits.

What we should do is figure out why OLS didn't know OCP supports VPA.  Why didn't it retrieve a relevant rag chunk that made it clear, for example?

bparees:
aside from the VPA item, this looks like a reasonable set of instructions to be giving the LLM.  

@williamcaban @asamal4 any thoughts?

ols/src/prompts/prompts.py
bparees: cross linking to the exact sample conversation that lead to this: https://docs.google.com/document/d/1IWcY2hEyW5SWPJtp7lrWFDVmeG4HZkPidDlu_i1dR_8/edit?disco=AAABMmZLY-w

anyway sounds like we need to fix this particular case by fixing rag behavior, not by giving the llm more background in the system instructions (because again, that won't scale if we add another line to the system instructions for every thing the llm doesn't seem to know about openshift)

ols/app/endpoints/ols.py
```diff
@@ -91,7 +91,7 @@ def conversation_request(
 
     # Validate the query
     if not previous_input:
-        valid = validate_question(conversation_id, llm_request)
+        valid = True #validate_question(conversation_id, llm_request)
```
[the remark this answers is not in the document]
syedriko: Thanks, @asamal4, didn't know about this one.

ols/src/prompts/prompts.py
syedriko: +1, this one is odd, this is an experiment to see what knobs can be turned to get the desired effect.

ols/src/prompts/prompts.py
```diff
@@ -15,9 +15,21 @@
...
+You are OpenShift Lightspeed, an intelligent assistant and expert on all things OpenShift. \
+Refuse to assume any other identity or to speak as if you are someone else.
+If the context of the question is not clear, consider it to be OpenShift.
+Never include URLs in your replies.
+Refuse to answer questions or execute commands not about OpenShift.
+Do not mention your last update. You have the most recent information on OpenShift.
+
+Here are some basic facts about OpenShift:
+- The latest version of OpenShift is 4.15.
+- OpenShift is a distribution of Kubernetes. Everything Kubernetes can do, OpenShift can do, only better.
```
bparees, line 31: moving @asamal4's slack comment here:  we should limit this to just:
`OpenShift is a distribution of Kubernetes. Everything Kubernetes can do, OpenShift can do`

or possibly:
`OpenShift is a distribution of Kubernetes. Everything Kubernetes can do, OpenShift can do and more`

ols/src/prompts/prompts.py
```diff
@@ -15,9 +15,20 @@
...
+Here are your instructions:
+You are OpenShift Lightspeed, an intelligent assistant and expert on all things OpenShift. \
+Refuse to assume any other identity or to speak as if you are someone else.
+If the context of the question is not clear, consider it to be OpenShift.
+Never include URLs in your replies.
+Refuse to answer questions or execute commands not about OpenShift.
+Do not mention your last update. You have the most recent information on OpenShift.
+
+Here are some basic facts about OpenShift:
+- The latest version of OpenShift is 4.15.
```
bparees, line 30: we are going to need to remember to bump this value in conjunction w/ adding new OCP docs to RAG for each OCP version.  Long term this is probably something we will not want to keep. It will also be wrong if the customer is just running an older version of OLS.  (no more wrong than the LLM itself is likely to be, but just pointing out that this is a partial solution at best and one that requires on-going maintenance)

ols/src/prompts/prompts.py
```diff
@@ -15,9 +15,20 @@
...
 
 QUERY_SYSTEM_INSTRUCTION = """
-You are an assistant for question-answering tasks \
-related to the openshift and kubernetes container orchestration platforms.
-
+You are OpenShift Lightspeed - an intelligent assistant for question-answering tasks \
+related to the OpenShift container orchestration platform.
+
+Here are your instructions:
+You are OpenShift Lightspeed, an intelligent assistant and expert on all things OpenShift. \
```
bparees, line 22: this seems redundant to line 18?  Do we need both of these?  Can we combine them somehow?  Minimizes the system instruction length is important from a context-window perspective and response latency perspective.

syedriko: I'll drop this and do a spot check.

bparees:
i'm good w/ merging this as is and doing https://github.com/openshift/lightspeed-service/pull/839#discussion_r1604122376 as a followup, but you've got some failing test jobs that need to be addressed too, since your system instruction changes have disrupted a few of our tests.

ols/src/prompts/prompts.py
syedriko: If I dedupe lines 22 and 18, the LLM doesn't know what "OpenShift Lightspeed" is anymore. Instead of 

```
Q: What is OpenShift Lightspeed?

A: OpenShift Lightspeed is the identity I embody as an intelligent assistant...
```

it starts saying

```
Q: What is OpenShift Lightspeed?

A: OpenShift Lightspeed is not an official feature or component of OpenShift. It appears there might be a 
misunderstanding or confusion in the question. If you have any questions about OpenShift, such as its 
features, functionalities, or how to use it for container orchestration and management, feel free to ask!
```

ols/src/prompts/prompts.py
syedriko: This doesn't actually depend on what OCP version the customer is running :)
Hopefully we will find a better way to feed "facts about the world" to the LLM, this was just an attempt to see if this crude method would work. Until we do, yes, we will need to bump this, hopefully only once or twice.

ols/src/prompts/prompts.py
bparees: ok, just leave it i guess.  but you might consider adding a test case for this specific behavior to ensure we don't regress on being able to answer this question in the future.

for that matter you should consider tests for the other behaviors you're trying to enforce w/ these instructions, so that in the future if we revisit the instructions we have tests that ensure we maintain the desired behavior (they may have to be updated to accomodate changes in response, but at least having the tests makes it clear to future contributors what the expected behavior is that needs to be maintained)

ols/src/prompts/prompts.py
bparees: > This doesn't actually depend on what OCP version the customer is running :)

no but it depends on the version of OLS they are running.  If they don't update to a new version of OLS where we changed this instruction to "4.16" in the future, their OLS will continue to tell them 4.15 is the latest.

syedriko: Right, you said OLS, not OCP

syedriko:
@bparees PTAL

PR #844, opened 2024-04-30
OLS-506: Remove unused tests markers

onmete:
## Description

Remove unused tests markers.
The `standalone` marker will be removed in a separate PR with the rest of the relevant code.

## Type of change

- [x] Refactor


bparees:
/lgtm
/approve

i'm ok w/ removing them since we're not using them, but i can envision a future where we do have provider specific e2e tests we want to mark, so these might come back in the future and that's ok.

PR #850, opened 2024-04-30
Pin dependencies to exact version

tisnik:
## Description

Pin dependencies to exact version

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


pyproject.toml
```diff
@@ -63,11 +63,11 @@ dependencies = [
...
-    "setuptools>=69.5.1",
-    "prometheus-client>=0.19.0",
-    "kubernetes>=29.0.0",
-    "pytest-asyncio>=0.23.6",
-    "psycopg2-binary>=2.9.9",
+    "setuptools==69.5.1",
+    "prometheus-client==0.20.0",
+    "kubernetes==29.0.0",
+    "pytest-asyncio==0.23.6",
+    "psycopg2-binary==2.9.9",
```
bparees, line 70: there are also some dev deps that use version ranges. do you care about those?

tisnik: that's good question. Personally I'd use `==` there too, dunno what others think about it

bparees: just do it, though as long as we have the .lock file it shouldn't matter a whole lot.

based on https://redhat-internal.slack.com/archives/C068JAU4Y0P/p1714491696491739?thread_ts=1714484369.505429&cid=C068JAU4Y0P it seems like the right thing to do if we're going to rely on dependabot to tell us when new versions are available.

tisnik: aaand...done

PR #851, opened 2024-05-01
Return value fix

tisnik:
## Description

Return value fix (we can simply return `None`, the only distinction between success and failure is signaled by exception raised)

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/e2e/cluster_utils.py
+++ b/tests/e2e/cluster_utils.py
@@ -112,7 +112,7 @@ def list_path(pod_name: str, path: str) -> list[str]:
 def remove_dir(pod_name: str, directory: str) -> None:
     """Remove a directory in a pod."""
     try:
-        return run_oc(["rsh", pod_name, "rm", "-rf", directory])
+        run_oc(["rsh", pod_name, "rm", "-rf", directory])
     except subprocess.CalledProcessError as e:
         raise Exception("Error removing directory") from e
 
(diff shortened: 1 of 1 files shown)


tests/e2e/cluster_utils.py
```diff
@@ -176,6 +176,6 @@ def create_file(pod_name: str, path: str, content: str) -> None:
 def remove_file(pod_name: str, path: str) -> None:
     """Remove a file in a pod."""
     try:
-        return run_oc(["exec", pod_name, "--", "rm", path])
+        run_oc(["exec", pod_name, "--", "rm", path])
```
bparees, line 179: do you happen to know if the "CalledProcessError" is raised if the process returns a non-zero exit code?

if not, we should fix these helpers up to ensure we throw an error/exception if the command returns non-zero.  (that can be done in run_oc itself)

tisnik: yes, because `check=True` is used in `subprocess.run`:

```
>>> import subprocess
>>> subprocess.run(["false"])
CompletedProcess(args=['false'], returncode=1)
>>> subprocess.run(["false"], check=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['false']' returned non-zero exit status 1.
```

bparees: i'm going to give myself credit that i did that intentionally when i wrote run_oc and just forgot about it.

PR #858, opened 2024-05-02
OLS-527: bump up libs

asamal4:
## Description

bump llama-index from 0.10.31 to 0.10.33
bump langchain-openai from 0.1.3 to 0.1.4

Other dependent package (due to above change):
langchain-core from 0.1.45 to 0.1.48


tisnik (commented):
looks good

PR #867, opened 2024-05-03
Fix logging in the number of attempts

onmete:
## Description

Fix logging in the number of attempts

## Type of change

- [x] Refactor


tisnik (commented):
heh I recall I fixed it too recently. Too magic constants on two lines ;)

PR #870, opened 2024-05-03
Print e2e tests durations

tisnik:
## Description

Print e2e tests durations


bparees:
where will these values be visible?  prow is already showing the test durations:

image

tisnik:
it should sort tests according to durations + print slowest ones at beginning. It's just test if it's useful or not (looking at BAM-slownes)

onmete:
It doesn't really change what is displayed in the CI UI. You can find the test duration in the `build-log.txt` if needed.

/lgtm

PR #872, opened 2024-05-06
Bug fix: check if score is not None by not accessing field directly

tisnik:
## Description

Bug fix: check if score is not None by not accessing field directly

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


ols/utils/token_handler.py
```diff
@@ -125,7 +125,7 @@ def truncate_rag_context(
 
         for node in retrieved_nodes:
 
-            score = float(node.score)
+            score = float(node.get_score())
```
onmete, line 128: Maybe `get_score(raise_error=False)` to ensure `0` is returned instead of `None`.

tisnik: hmm do we need to put the error under the carpet? but can do it

tisnik: updated @onmete

onmete: Just thinking it may be safer due to changes in llm libraries in general. :)

PR #874, opened 2024-05-06
Use proper in memory cache config class name

tisnik:
## Description

`MemoryConfig` is basically meaningles, let's use better self-documented name instead

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -118,9 +118,9 @@ RAG_SIMILARITY_CUTOFF_L2 = 1.4
 
 
 # cache constants
-IN_MEMORY_CACHE = "memory"
+CACHE_TYPE_MEMORY = "memory"
 IN_MEMORY_CACHE_MAX_ENTRIES = 1000
-REDIS_CACHE = "redis"
+CACHE_TYPE_REDIS = "redis"
 REDIS_CACHE_HOST = "lightspeed-redis-server.openshift-lightspeed.svc"
 REDIS_CACHE_PORT = 6379
 REDIS_CACHE_MAX_MEMORY = "1024mb"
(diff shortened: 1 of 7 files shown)


ols/constants.py
```diff
@@ -118,9 +118,9 @@ class GenericLLMParameters:
 
 
 # cache constants
-IN_MEMORY_CACHE = "memory"
+CACHE_TYPE_MEMORY = "memory"
```
onmete, line 121: `"in-memory"`?

tisnik: probably not at this moment (summing time), as it would be user-visible change (`olsconfig.yaml` etc.)

bparees: please no :)

it was in-memory originally, then we changed it to memory because i didn't like how "in-memory" manifested in the user facing config.

tbh i debated nacking this whole PR for that reason, but i appreciated the change from "MemoryConfig" to MemoryCacheConfig" regardless of the change to "InMemory" vs "Memory"

xrajesh:
> CACHE_TYPE_POSTGRES

Do we want to use the name Postgres ? I was planning to have NO reference of DB flavor. Keep it generic as DB . Currently we have `RedisSpec` in CRD,  was planning to  make it as `DBSpec` - did not want the CRD to be tied to any specific DB name. WDYT @tisnik @bparees

tisnik:
> > CACHE_TYPE_POSTGRES
> 
> Do we want to use the name Postgres ? I was planning to have NO reference of DB flavor. Keep it generic as DB . Currently we have `RedisSpec` in CRD, was planning to make it as `DBSpec` - did not want the CRD to be tied to any specific DB name. WDYT @tisnik @bparees

How it would work? This is all about DB-specific configuration options, nothing universal that can be used by any (relational?, no-SQL??) database.

xrajesh:
> what is your concern/motivation to keep postgres out of the name?
-- Thinking how much DB specific configs we want to expose to the user . Since our use case is just a cache and not planning to store a lot of data. We should be able to have a pre-tuned version as a postgres image. Even if we have to pass a new DB parameter , we can pass it as ENV Arguments to the container. My only intention was to keep the store as generic/abstract from OLS stand point.

bparees:
I agree we need to be very careful/minimal about what config options we expose to the end user via the CRD.

But that doesn't preclude naming the config section "postgres" or defining a "postgres" cache storage type.

PR #877, opened 2024-05-06
Config object example - WIP

onmete:
## Description

Example of config object and usage.


onmete:
There are some alternative approaches too:
- we can use singleton too, if we want to ensure that only one instance of config is used, eg.
```python
class ConfigSingleton:
    pass
config = ConfigSingleton("path/to/config.yaml")
```
- we can either initialize the singleton here and store it under the `config` variable, or we can just use the singleton directly in the other parts of the code (like logging at the top of each module that needs logging), eg. 
```python
from ols.utils.config import load_config
config = load_config()  # <- this will return the singleton instance
```

As we don't want to execute the query filters and index loader right away, lazily evaluated singleton might be a better fit than just module constant.

onmete:
closing, implemented in another PR

PR #882, opened 2024-05-07
OLS-546: Use cluster id as user id

onmete:
## Description

Use cluster id as user id

## Type of change

- [x] New feature

### DIFF
--- a/ols/utils/auth_dependency.py
+++ b/ols/utils/auth_dependency.py
@@ -10,9 +10,16 @@ from kubernetes.client.rest import ApiException
 from kubernetes.config import ConfigException
 
 from ols import config
-from ols.constants import DEFAULT_KUBEADMIN_UID, DEFAULT_USER_NAME, DEFAULT_USER_UID
+from ols.constants import DEFAULT_USER_NAME, DEFAULT_USER_UID, RUNNING_IN_CLUSTER
 
 logger = logging.getLogger(__name__)
+
+
+CLUSTER_ID_LOCAL = "local"
+
+
+class ClusterIDUnavailableError(Exception):
+    """Cluster ID is not available."""
 
 
 class K8sClientSingleton:
@@ -26,6 +33,7 @@ class K8sClientSingleton:
     _api_client = None
     _authn_api: kubernetes.client.AuthenticationV1Api
     _authz_api: kubernetes.client.AuthorizationV1Api
+    _cluster_id = None
 
     def __new__(cls: type[Self]) -> Self:
         """Create a new instance of the singleton, or returns the existing instance.
@@ -89,6 +97,7 @@ class K8sClientSingleton:
                 )
                 api_client = kubernetes.client.ApiClient(configuration)
                 cls._api_client = api_client
+                cls._custom_objects_api = kubernetes.client.CustomObjectsApi(api_client)
                 cls._authn_api = kubernetes.client.AuthenticationV1Api(api_client)
                 cls._authz_api = kubernetes.client.AuthorizationV1Api(api_client)
             except Exception as e:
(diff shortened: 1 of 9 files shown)


ols/utils/auth_dependency.py
```diff
@@ -144,6 +155,27 @@ def get_user_info(token: str) -> Optional[kubernetes.client.V1TokenReview]:
         )
 
 
+def get_cluster_id() -> str:
```
tisnik, line 158: we already have exactly the same code in your data collector. so perhaps refactor it?

onmete: Yup, see the `# TODO: use the one from auth_dependency.py` in there.
As I need to most likely update syspath to make it work with import, more testing is involved and I didn't wanted to do it in this PR.

tests/unit/utils/test_auth_dependency.py
```diff
@@ -96,3 +129,35 @@ def test_auth_dependency_config(_setup):
...
+
+
+@patch("ols.utils.auth_dependency.K8sClientSingleton.get_custom_objects_api")
+def test_get_cluster_id(mock_get_custom_objects_api):
+    """Test get_cluster_id function."""
+    cluster_id = {"spec": {"clusterID": "some-cluster-id"}}
+    mocked_call = MagicMock()
+    mocked_call.get_cluster_custom_object.return_value = cluster_id
+    mock_get_custom_objects_api.return_value = mocked_call
+    assert get_cluster_id() == "some-cluster-id"
```
tisnik, line 141: prob. use real UUID? will it be tested in the code btw?

onmete: Do you mean ensuring what we get from the cluster as id is really uuid? That seems like an overkill to me, but I can add it if you think it is reasonable.

ols/utils/auth_dependency.py
```diff
@@ -114,6 +115,16 @@ def get_authz_api(cls) -> kubernetes.client.AuthorizationV1Api:
             cls()
         return cls._authz_api
 
+    @classmethod
+    def get_custom_objects_api(cls) -> kubernetes.client.CustomObjectsApi:
+        """Return the custom objects API instance.
+
+        Ensures the singleton is initialized before returning the Authorization API client.
+        """
+        if cls._instance is None or cls._authz_api is None:
```
xrajesh, line 124: Should this be `cls._custom_objects_api is None`

onmete: definitely, thx

onmete:
@bparees @xrajesh The user id in the e2e CI is "system:serviceaccount:openshift-lightspeed:test-user". In order to have the feature in the PR covered by e2e test, I need to hit the /authorized endpoint as the "kube:admin". Would you know an easy way how to do it besides updating the the request model of that endpoint to allow post user (which I'm not sure if we want to do)? Another option would be to remove the e2e test.

xrajesh:
@onmete The tests are using the test-user's token. [here ](https://github.com/openshift/lightspeed-service/blob/fea536180a16f2e1b516aba6f209d4a76a2002ad/tests/e2e/test_api.py#L74) . You can make a API call as kubeadmin by getting the kubeadmin's token. I am thinking we are logging in as the kubeadmin. Hence  `kubeadmin_token=$(oc whoami -t)` should give us the kubeadmin token. We will need a function just like `def get_user_token(name: str) -> str:` to run the `oc whoami -t` . If you make use of that token and make a OLS api call, it should get you `"kube:admin"`

ols/utils/auth_dependency.py
```diff
@@ -206,7 +238,7 @@ async def __call__(self, request: Request) -> tuple[str, str]:
                 status_code=403, detail="Forbidden: Invalid or expired token"
             )
         if user_info.user.username == "kube:admin":
-            user_info.user.uid = DEFAULT_KUBEADMIN_UID
+            user_info.user.uid = get_cluster_id()
```
bparees, line 241: the cluster id isn't going to change between user queries, so we should just fetch it once during OLS startup and then use that value every time we get a request from kubeadmin.

this will also help avoid anther issue you're not handling here, which is what happens if getting the cluster id fails?  you log the errors in the helper, but ultimately return `None` to the caller.

if you fetch it during startup you can retry + refuse to start until we successfully get the cluster id.

it also removes the extra/repeated api calls.

tests/e2e/test_api.py
```diff
@@ -728,6 +728,29 @@ def test_forbidden_user():
     assert response.status_code == requests.codes.forbidden
 
 
+# TODO: disabled until we figure out how to get the token that returns
+# user kube:admin in the CI
```
bparees, line 732: we can at least have a legitimate test of the get_cluster_id helper, right?  (not really an e2e test since you won't be going through the OLS server, but since we have a cluster stood up, you can test that the get_cluster_id helper gets the right value.

tests/config/cluster_install/ols_manifests.yaml
```diff
@@ -187,3 +187,30 @@ rules:
     - "/ols-metrics-access"
   verbs:
     - "get"
+---
+# permissions to access the clusterversions object
```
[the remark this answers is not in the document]
xrajesh: @bparees This is only on the test . But for the operator we already have `get` for `clusterversions`
```
			{
				APIGroups: []string{"config.openshift.io"},
				Resources: []string{"clusterversions"},
				Verbs:     []string{"get"},
			},
```
Do we need watch and list here @onmete ?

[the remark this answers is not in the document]
onmete: Yes, I knew that someone already did these changes in operator at the time when we were solving data collector, as there was/is function accessing `clusterversions` resource to get the `cluster_id`. I think `get` should be enough, but I can't guarantee as I don't know much about that area.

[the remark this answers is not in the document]
bparees: ok sounds good.  No, we do not need list+watch for clusterversion at present since the only thing we look at is the cluster_id and that should not change.

tisnik (commented):
if you rebase, all your tests will pass

tests/config/cluster_install/ols_manifests.yaml
```diff
@@ -187,3 +187,30 @@ rules:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: clusterversion-viewer
+  labels:
+    app: ols
+rules:
+- apiGroups: ["config.openshift.io"]
+  resources: ["clusterversions"]
+  verbs: ["get", "watch", "list"]
```
xrajesh, line 201: @onmete this can be just ["get"] - We dont need rest of the 2.

xrajesh:
@onmete minor change proposed - otherwise - this LGTM.

PR #885, opened 2024-05-07
Fix: Remove default test tags

onmete:
## Description

Remove default test tag.

My intention was that cluster tests are omitted when running locally, but the result was that tests were omitted in CI too as the TEST_TAGS was set to "" -> default with "not cluster" was used.

## Type of change

- [x] Bug fix


tisnik (commented):
nice catch

PR #886, opened 2024-05-07
Script to list objects stored in Ceph bucket

tisnik:
## Description

Script to list objects stored in Ceph bucket
(will be also used as an example in our internal documentation)

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] Utility

### DIFF
--- a/scripts/list_objects_in_ceph.py
+++ b/scripts/list_objects_in_ceph.py
@@ -1,0 +1,44 @@
+"""List all objects stored in Ceph bucket.
+
+Four environment variables should be set:
+    ENDPOINT_URL - Ceph endpoint URL
+    AWS_ACCESS_KEY - access key for user or service
+    AWS_SECRET_ACCESS_KEY - secret access key for user or service
+    BUCKET - bucket name, for example QA-OLS-ARCHIVES or PROD-OLS-ARCHIVES
+
+"""
+
+import os
+from pprint import pprint
+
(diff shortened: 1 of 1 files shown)


onmete (commented):
Is this something that should be part of this repo? I mean there is a lot of clutter and stuff that doesn't necessary belong here anyway. So when we do a cleanup, this should be moved elsewhere (docs) too right?

tisnik:
> Is this something that should be part of this repo? I mean there is a lot of clutter and stuff that doesn't necessary belong here anyway. So when we do a cleanup, this should be moved elsewhere (docs) too right?

I had exactly the same question about sidecar and other scripts.

onmete:
So the answer is: _it is fiiine for now_ 😅 

/lgtm

bparees:
> I had exactly the same question about sidecar and other scripts.

the sidecar is logic is literally part of the core OLS application + image, so this isn't the same as that.

That said, i don't see a reason to go off and create another repo for this.  But i do think the directory structure could use some work.  The "scripts" dir contains quite a variety of unrelated bits at this point, i'd like to see some better directory/subdirectory name+structure so it's not just a dumping ground of random dev/tool/reporting/testing scripts.

PR #895, opened 2024-05-09
OLS-473: Chunk collected user data to archives

onmete:
## Description

Chunk collected user data to archives and ensure the data dir can't grow over the limit.

## Type of change

- [x] New feature


onmete:
Implementation-wise, we check the limit against files, not an actual archive -> when we hit the limit, the archive is actually smaller than the limit (100MB).

Given we probably never hit the limit, this was easier to implement because we constructed the archive in memory. I didn't find an easy way how to check the archive limit without either "closing" tar after each file added and checking its size (doesn't seem efficient) or having another virtual one that was checked each file added (might cause double memory consumption for archive object).
So I liked the current implementation enough to go with that instead.

PR #906, opened 2024-05-13
Remove unreachable return from postgres cache

onmete:
## Description

Remove unreachable return from postgres cache

## Type of change

- [x] Refactor


tisnik:
it looks ok!

PR #907, opened 2024-05-13
OLS-551: Remove langchain messages objects in favor of simple dicts

onmete:
## Description

Remove langchain messages objects in favor of simple dicts

## Type of change

- [x] Refactor


onmete:
@tisnik @asamal4 Review, please.

This feels like a step back though. It would be better to have a defined interface for messages instead of "random" dict.

ols/app/endpoints/ols.py
```diff
@@ -222,13 +219,23 @@ def generate_response(
...
+    ```
+    """
     try:
         if config.conversation_cache is not None:
             logger.info(f"{conversation_id} Storing conversation history.")
-            chat_message_history = ChatHistory.get_chat_message_history(
-                llm_request.query, response or ""
-            )
+            chat_message_history = [
+                {"type": "human", "content": llm_request.query},
```
tisnik, line 236: I'd probably add "type", "human" etc. into constants (to avoid typos) or rather have two constructors prepared (that simply return dicts)

onmete:
If you don't have more comments, could
@asamal4 do lgtm
and
@tisnik approve?

ols/app/endpoints/ols.py
```diff
@@ -219,16 +216,36 @@ def generate_response(
...
 def store_conversation_history(
     user_id: str, conversation_id: str, llm_request: LLMRequest, response: Optional[str]
 ) -> None:
-    """Store conversation history into selected cache."""
+    """Store conversation history into selected cache.
+
+    History is stored as simple dictionaries in the following format:
+    ```python
+        [
+            {"type": "human/ai", "content"},
```
tisnik, line 237: "content": "...."

PR #910, opened 2024-05-13
`AnyHttpUrl` type parameter instead of raw `str`

tisnik:
## Description

`AnyHttpUrl` type parameter instead of raw `str`

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


onmete (commented):
Isn't `_is_valid_http_url` redundant? Pydantic checks the value agains `AnyHttpUrl`, right?

bparees:
> Isn't _is_valid_http_url redundant? Pydantic checks the value agains AnyHttpUrl, right?

I was also wondering this.

what happens to calls like this if the value is not a url?

https://github.com/openshift/lightspeed-service/blob/65e44ad1c5f9ce0762b4fa00842c818ddf990dd3/ols/app/models/config.py#L242

i assume it'll fail right there in pydantic validation, before it ever gets to:
https://github.com/openshift/lightspeed-service/blob/65e44ad1c5f9ce0762b4fa00842c818ddf990dd3/ols/app/models/config.py#L284

?

onmete:
We were talking about this with Pavel offline. TLDR, as we are overriding the init (and not passing the data through `super`) in most of our config objects, we are doing the validation on our side.
That is one of the issues Michal was solving in his PR.

PR #912, opened 2024-05-13
`yaml_port` is used in `raise` block so it need to be defined outside `try` block

tisnik:
## Description

`yaml_port` is used in `raise` block so it need to be defined outside `try` block

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -366,8 +366,8 @@ class RedisConfig(BaseModel):
             return
         self.host = data.get("host", constants.REDIS_CACHE_HOST)
 
+        yaml_port = data.get("port", constants.REDIS_CACHE_PORT)
         try:
-            yaml_port = data.get("port", constants.REDIS_CACHE_PORT)
             self.port = int(yaml_port)
             if not (0 < self.port < 65536):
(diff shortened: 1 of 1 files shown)


onmete:
There is a commit from another PR here.

ols/app/models/config.py
```diff
@@ -366,8 +366,8 @@ def __init__(self, data: Optional[dict] = None) -> None:
             return
         self.host = data.get("host", constants.REDIS_CACHE_HOST)
 
+        yaml_port = data.get("port", constants.REDIS_CACHE_PORT)
```
bparees, line 369: i don't understand why this is necessary?

it seems to work fine here as is:

https://www.online-python.com/Y1PO2MxmAC

i could see an issue if the variable was used outside the try block, but it is not.

tisnik: it works only because the exception is thrown after variable initialization. But if new code will be added in the try block, it won't work.

PR #913, opened 2024-05-14
Better retry logic for flaky end-to-end tests

tisnik:
## Description

Better retry logic for flaky end-to-end tests
- it is possible to specify error type
- it is possible to specify which text must be included in error message

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/e2e/test_decorators.py
+++ b/tests/e2e/test_decorators.py
@@ -4,8 +4,24 @@ import time
 from functools import wraps
 
 
-def retry(max_attempts=3, wait_between_runs=5):
-    """Construct decorator that allows to retry running selected test."""
+def retry(
+    max_attempts=3,
+    wait_between_runs=5,
+    on_error=AssertionError,
+    expected_error_message="",
+):
+    """Construct decorator that allows to retry running selected test.
+
+    Args:
+        max_attempts: how many times the test will be rerun in case of failure
+        wait_between_runs: time duration between consecutive test runs
+        on_error: expected error type (other errors will mark test failure immediatelly)
+        expected_error_message: text that must be contained in error message
+                                (if not, the test failure will be mark immediatelly)
+
+    Returns:
+        decorated test function
+    """
 
     def retry_test_decorator(test_function):
         @wraps(test_function)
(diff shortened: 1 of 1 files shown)


tests/e2e/test_decorators.py
```diff
@@ -4,22 +4,45 @@
 from functools import wraps
 
 
-def retry(max_attempts=3, wait_between_runs=5):
-    """Construct decorator that allows to retry running selected test."""
+def retry(
+    max_attempts=3,
+    wait_between_runs=5,
+    error_type=AssertionError,
```
onmete, line 10: on_error/on_errors?

tests/e2e/test_decorators.py
```diff
@@ -4,22 +4,45 @@
...
+    expected_error_message="",
+):
+    """Construct decorator that allows to retry running selected test.
+
+    Args:
+        max_attempts: how many times the test will be rerun in case of failure
+        wait_between_runs: time duration between consecutive test runs
+        error_type: expected error type (other errors will mark test failure immediatelly)
+        expected_error_message: text that must be contained in error messge
```
onmete, line 19: mess**a**ge

tisnik: fixed

tests/e2e/test_decorators.py
tisnik: fixed @onmete

tests/e2e/test_decorators.py
tisnik: fixed @onmete

tests/e2e/test_decorators.py
```diff
@@ -4,22 +4,45 @@
...
+        decorated test function
+    """
 
     def retry_test_decorator(test_function):
         @wraps(test_function)
         def wrapper(*args, **kwargs):
             retry_count = 1
 
-            while retry_count < max_attempts:
+            while retry_count <= max_attempts:
```
bparees, line 31: since you make one final attempt at invoking the test function outside the loop, i think `<` was correct?

tisnik: hmm yes. I was thinking the opposite - how much times to retry the test after failure, not the overall number of test runs. But can change it to `<` if you prefer (have no strong opinion).

bparees: well you called the argument "max attempts" and not "max retries", so i'd enforce it as "maximum number of times we will invoke the test"

(I suspect that even max retries is often interpreted by people to mean max attempts/tries, not "max attempts after the first one")

it hardly matters overall, but we might as well get it right while we're looking at it.

PR #915, opened 2024-05-14
OLS-595: Prolong wait interval for data collector in e2e test

onmete:
## Description

Prolong wait interval for data collector in e2e test.
The cause of the "bug story" is because we collected logs too fast. As seen from the logs, the last log was about posting the payload. We pulled the logs before the post was completed and logged.

The issue might occur again, eg. when console.stage.redhat.com is responding slowly, but the wait in this PR should be enough for standard situations.

## Type of change

- [x] Bug fix

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -1120,7 +1120,9 @@ def test_user_data_collection():
         timeout=BASIC_ENDPOINTS_TIMEOUT,
     )
     assert response.status_code == requests.codes.ok
-    time.sleep(OLS_USER_DATA_COLLECTION_INTERVAL + 1)
+    # ensure the script have enought time to send the payload before
+    # we pull its logs
+    time.sleep(OLS_USER_DATA_COLLECTION_INTERVAL + 5)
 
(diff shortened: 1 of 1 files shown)


tisnik (commented):
why not to increase the constant itself?

tests/e2e/test_api.py
```diff
@@ -1120,7 +1120,9 @@ def get_last_log_line(logs: str) -> str:
         timeout=BASIC_ENDPOINTS_TIMEOUT,
     )
     assert response.status_code == requests.codes.ok
-    time.sleep(OLS_USER_DATA_COLLECTION_INTERVAL + 1)
+    # ensure the script have enought time to send the payload before
+    # we pull its logs
+    time.sleep(OLS_USER_DATA_COLLECTION_INTERVAL + 5)
```
bparees, line 1125: @onmete if you find us hitting this issue again, i suggest replacing this w/ a retry loop that checks the logs every few seconds, giving up to, say, 30s for the expected log entry to appear.

that way we don't unnecessarily prolong the test in most cases, but also don't flake when things are slow.

PR #917, opened 2024-05-14
Timestamp column in feedback sheet

tisnik:
## Description

Timestamp column in feedback sheet based on time when the tarball was send into RH network

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


bparees:
/lgtm
/approve

to be replaced by getting the timestamp from the feedback data itself, once that is added

PR #932, opened 2024-05-16
OLS-379: Remove dev_config global in favor of using it from the main Config class

onmete:
## Description

Remove dev_config global in favor of using it from the main Config class.

## Type of change

- [x] Refactor

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -823,7 +823,7 @@ class Config(BaseModel):
 
     llm_providers: Optional[LLMProviders] = None
     ols_config: Optional[OLSConfig] = None
-    dev_config: Optional[DevConfig] = None
+    dev_config: Optional[DevConfig] = DevConfig()
 
     def __init__(self, data: Optional[dict] = None) -> None:
         """Initialize configuration and perform basic validation."""
--- a/ols/src/llms/providers/provider.py
+++ b/ols/src/llms/providers/provider.py
@@ -246,10 +246,10 @@ class LLMProvider(AbstractLLMProvider):
         updated_params = {**self.default_params, **params}
 
         # config params overrides everything
-        if config.dev_config.llm_params:
+        if config.config.dev_config.llm_params:
             logger.debug(
-                f"overriding LLM params with debug options {config.dev_config.llm_params}"
+                f"overriding LLM params with debug options {config.config.dev_config.llm_params}"
             )
-            updated_params = {**updated_params, **config.dev_config.llm_params}
+            updated_params = {**updated_params, **config.config.dev_config.llm_params}
 
         return updated_params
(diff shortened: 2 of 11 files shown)


ols/app/models/config.py
```diff
@@ -823,7 +823,7 @@ class Config(BaseModel):
 
     llm_providers: Optional[LLMProviders] = None
     ols_config: Optional[OLSConfig] = None
-    dev_config: Optional[DevConfig] = None
+    dev_config: Optional[DevConfig] = DevConfig()
```
onmete, line 826: This is the main change, the rest is just replacing how the config is accessed - not from the `config` module, but from the `config` variable holding the `Config` class.

tisnik: is it optional then?

ols/src/llms/providers/provider.py
```diff
@@ -246,10 +246,10 @@ def _override_params(self, params: dict[Any, Any]) -> dict[Any, Any]:
         updated_params = {**self.default_params, **params}
 
         # config params overrides everything
-        if config.dev_config.llm_params:
+        if config.config.dev_config.llm_params:
```
tisnik, line 249: I'm still thinking that by using "better" import strategy, just the `import` statement might change,nothing else in such modules

ols/app/models/config.py
onmete: From the `Config` perspective, yes. It is optional, we just init `DevConfig` with default values instead of `None`. Which is the same as "before", where the `DevConfig` was initialized under global.

tisnik: I mean, if it's still Optional from type system perspective.

tisnik: anyway, it's nitpick

ols/src/llms/providers/provider.py
onmete: Sure, we can do `from ols.utils.config import config` in main `__init__.py`, then we would just import `from ols import config` and the rest of the code should be untouched. But it would mess it up at this point as some parts of the code accessing the config module and others config holding `Config`. But yes, that is the goal.

ols/utils/config.py
```diff
@@ -24,11 +23,9 @@ def init_empty_config() -> None:
     global config
     global ols_config
     global llm_config
-    global dev_config
```
bparees, line 27: intent is to do subsequent separate PRs to remove the other globals once this PR/approach is agreed on?

onmete: Though so. I started with other globals in the meantime and it would take many PRs - with the right order - to make the change (prolong the effort significantly). So I did [943](https://github.com/openshift/lightspeed-service/pull/943) instead.

bparees:
/approve

@tisnik any more concerns on this one?  it lgtm

onmete:
Closed in favor of https://github.com/openshift/lightspeed-service/pull/943
See comment: https://github.com/openshift/lightspeed-service/pull/932#discussion_r1606912049

PR #934, opened 2024-05-16
An example: howto mock timestamp

tisnik:
An example: howto mock timestamp

### DIFF
--- a/tests/unit/app/endpoints/test_feedback.py
+++ b/tests/unit/app/endpoints/test_feedback.py
@@ -2,6 +2,7 @@
 
 import json
 from unittest.mock import patch
+from datetime import datetime
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/unit/app/endpoints/test_feedback.py
```diff
@@ -45,16 +46,19 @@ def test_get_feedback_status(feedback_location):
     assert not feedback.is_feedback_enabled()
 
 
-def test_store_feedback(feedback_location):
+@patch('ols.app.endpoints.feedback.datetime')
```
bparees, line 49: can't we just do:

```suggestion
@patch('ols.app.endpoints.feedback.datetime.utcnow', return_val=datetime(2000,1,1,1,23,45))
```
and avoid the lambda below?

tisnik: `utcnow` returns `callable`, not directly a value

bparees: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

```
This is like [now()](https://docs.python.org/3/library/datetime.html#datetime.datetime.now), but returns the current UTC date and time, as a naive [datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime) object.
```

??

tisnik:
no longer needed -> closing it

PR #943, opened 2024-05-20
OLS-379: Refactor configuration

onmete:
## Description

Config refactoring. More on https://github.com/openshift/lightspeed-service/pull/943#pullrequestreview-2066410752

## Type of change

- [x] Refactor


ols/utils/config.py
```diff
@@ -2,76 +2,100 @@
...
-
-def init_vector_index() -> None:
-    """Initialize vector index."""
-    # TODO: OLS-380 Config object mirrors configuration
-    global rag_index
-    rag_index = IndexLoader(ols_config.reference_content).vector_index
+from ols.utils.query_filter import QueryFilters
+
+
+class AppConfig:
```
onmete, line 17: This is the main change. The globals and initializing various parts is defined as a class now. Config subparts are accessible through properties.

ols/utils/query_filter.py
```diff
@@ -14,17 +14,17 @@
 # TODO: OLS-380 Config object mirrors configuration
 
 
-class QueryFilter:
+class QueryFilters:
```
onmete, line 17: There was a conflict with `QueryFilter` config model. The new name seems to be more proper anyway.

I've also put `query_filters` as mandatory input instead of using that from the config (it is initialized in the `AppConfig`).

onmete (commented):
I've tried to sum-up the main changes in the comments below.

Some tests were removed - eg. `test_config_no_llm_providers`, which checks that we raise when subconfig is not defined in yaml. We still raise when the required field is missing, eg. `default_provider`, ..., but we don't check the YAML content basically. 
We can control the fields existence via how we set pydantic models (not using `Optional` when we want field to be there).

Some tests were updated because of how config was refactored.

The rest is just changing imports.

If config is **not set** during the tests (it is in most cases in setup), the tests are manipulating with global config, that persists between tests.
It is not ideal. It is trade-off between easy of use on the app side (I'm open for ideas - initializing config in every module where it is needed, like `logger`, might be a way too).
We do have some sort of `setup` for most of the tests, where we set config in specific way. It doesn't matter in other tests (or they may fail if we change defaults in config models).

tests/integration/test_feedback.py
```diff
@@ -17,7 +18,7 @@
 def _setup():
     """Setups the test client."""
     global client
-    config.init_config("tests/config/valid_config.yaml")
+    config.reload_from_yaml_file("tests/config/valid_config.yaml")
```
bparees, line 21: why did you change the scope in test_authorized.py, but not here?

tests/unit/llms/providers/test_azure_openai.py
```diff
@@ -27,10 +27,13 @@ def provider_config():
     )
 
 
+def setup_module(module):
```
bparees, line 30: shouldn't this have a @pytest.fixture(scope="module") decorator?  or is "setup_module" a magic name?

also the previous behavior was invoking creating a new config for each test, now it's only going to happen for the module?

bparees (commented):
a lot of good cleanup in here, thanks.  couple small questions.

/approve

tests/integration/test_feedback.py
onmete: You mean scope -> `function`? Because this fixture loads defined config, but there are other fixtures `_with_disabled_feedback` and `_with_enabled_feedback` that controls the config in each test.

But it is probably better to have it defined as a `function` by default - these tests are not affected by the change.

Changed.

tests/unit/llms/providers/test_azure_openai.py
onmete: Similar reason as above. The `setup_module` is magic, but it is executed only once per module. It works here, as these tests are not changing config values.
These tests actually doesn't depend on config at all, the config setting was here (as in many other tests) because we didn't had defaults for subconfigs before.

~~Again, to be on safe side, I changed it to function-scoped fixture to ensure the config is initialized fresh for each test.~~

I added `conftest.py` into unit and integration package (we could have just one contest in the tests root, but we are running tests per package) with fixture that is executed **before every test** and sets the config as empty.
With this, we can ensure that if some test make changes in the config, it stay "isolated" as each test gets a fresh new one by default.

So I cleaned test a bit from the extra config setting that is not needed with this change. And there was just one test that was failing as was depending on the config state from the other test - so we are good now.

PR #945, opened 2024-05-20
Makefile target to run bandit + updated documentation + updated PDM commands

tisnik:
## Description

Makefile target to run bandit + updated documentation + updated PDM commands

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [x] Documentation Update
- [x] Configuration Update


Makefile
```diff
@@ -68,6 +68,9 @@ coverage-report:	test-unit ## Export unit test coverage report into interactive
 check-types: ## Checks type hints in sources
 	mypy --explicit-package-bases --disallow-untyped-calls --disallow-untyped-defs --disallow-incomplete-defs ols/
 
+security-check: ## Check the project for security issues
+	bandit -r ols
```
bparees, line 72: @tisnik most likely no one is going to run this on their own.  Is it stable/reliable enough that it makes sense to run this as part of the `verify` target so it gets run against all PRs and we don't regress?

PR #951, opened 2024-05-21
OLS-631: Use buildin dict in typing

onmete:
## Description

Use buildin dict in typing

## Type of change

- [x] Refactor


tisnik (commented):
ok

PR #953, opened 2024-05-21
OLS-622: provider-specific configuration

tisnik:
## Description

Provider-specific configuration

Please note that discriminated unions are not used in the end. It would be possible to use them in following manner:

```python
class AzureOpenAIConfig(BaseModel):
    url: str
    api_key: str


class WatsonxConfig(BaseModel):
    url: str


class ProviderSpecificConfig(RootModel):
    root: Union[AzureOpenAIConfig, WatsonxConfig]


class ProviderConfig(BaseModel):
    """LLM provider configuration."""

    name: Optional[str] = None
    type: Optional[str] = None
    provider_config: Optional[ProviderSpecificConfig] = None
```

but in the end we'll need to re-cast everytime config is used, no proper checks against type could be done and the type hints will become unusable at the same time. So I prefer to be dumb&specific, not overthink it;)

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Related Issue #[OLS-622](https://issues.redhat.com//browse/OLS-622)

## Testing

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -214,6 +214,31 @@ class AuthenticationConfig(BaseModel):
                 )
 
 
+class OpenAIConfig(BaseModel):
+    """Configuration specific to OpenAI provider."""
+
+    url: AnyHttpUrl
+
+
+class AzureOpenAIConfig(BaseModel):
+    """Configuration specific to Azure OpenAI provider."""
+
+    url: AnyHttpUrl
+    api_key: str
+
+
+class WatsonxConfig(BaseModel):
+    """Configuration specific to Watsonx provider."""
+
+    url: AnyHttpUrl
+
+
+class BAMConfig(BaseModel):
+    """Configuration specific to BAM provider."""
+
+    url: AnyHttpUrl
+
+
 class ProviderConfig(BaseModel):
     """LLM provider configuration."""
 
@@ -224,6 +249,10 @@ class ProviderConfig(BaseModel):
     project_id: Optional[str] = None
     models: dict[str, ModelConfig] = {}
     deployment_name: Optional[str] = None
+    openai_config: Optional[OpenAIConfig] = None
+    azure_config: Optional[AzureOpenAIConfig] = None
+    watsonx_config: Optional[WatsonxConfig] = None
+    bam_config: Optional[BAMConfig] = None
 
     def __init__(self, data: Optional[dict] = None) -> None:
         """Initialize configuration and perform basic validation."""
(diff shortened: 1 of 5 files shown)


ols/app/models/config.py
```diff
@@ -277,6 +308,51 @@ def setup_models_config(self, data: dict) -> None:
             model = ModelConfig(m, self.type)
             self.models[m["name"]] = model
 
+    def set_provider_specific_configuration(self, data: dict) -> None:
+        """Set the provider-specific configuration."""
+        # compute how many provider-specific configurations are
+        # found in config file
+        found = 0
+        for x in constants.SUPPORTED_PROVIDER_TYPES:
+            cfg_name = x + "_config"
```
onmete, line 317: Maybe `x.lower()` just in case?

ols/app/models/config.py
```diff
@@ -277,6 +308,51 @@ def setup_models_config(self, data: dict) -> None:
...
+                case constants.PROVIDER_OPENAI:
+                    self.openai_config = data.get("openai_config", None)
+                    self.check_provider_config(self.openai_config)
+                case constants.PROVIDER_BAM:
+                    self.watsonx_config = data.get("bam_config", None)
+                    self.check_provider_config(self.bam_config)
+                case constants.PROVIDER_WATSONX:
+                    self.watsonx_config = data.get("watsonx_config", None)
+                    self.check_provider_config(self.watsonx_config)
+
```
onmete, line 347: Isn't some raise here missing in case we don't match?

ols/app/models/config.py
tisnik: indeed

ols/app/models/config.py
tisnik: we match for sure, because the code to retrieve provider type does this check. But ok, I can add `_` case there

ols/app/models/config.py
tisnik: updated @onmete

ols/app/models/config.py
tisnik: updated @onmete

onmete (commented):
What happens if there is an extra field in the provider-specific config?

tisnik:
> What happens if there is an extra field in the provider-specific config?

The same what happens if more fields are in model, etc. It will be detected & sanitized later in `llms/proviers/provider.py`, but in another PR/story. Right now we need to have 'something' for operator to start using it.

PR #954, opened 2024-05-21
OLS-631: Switch ReferencedDocument from BaseModel to dataclass

onmete:
## Description

**For consideration.**

Change `ReferencedDocument` definition to pydantic dataclass (standard dataclass but with pydantic typing).
IMHO it is a better fit. You can also create an instance with direct args/kwargs instead of feeding it with dict.

## Type of change

- [x] Refactor

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -653,29 +653,19 @@
             "ReferencedDocument": {
                 "properties": {
                     "docs_url": {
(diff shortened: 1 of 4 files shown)


tisnik (commented):
that would be even better!

tisnik (commented):
it looks ok

PR #963, opened 2024-05-22
OLS-653: remove folders before getting rag content

asamal4:
## Description
Remove folders before getting rag content.


tisnik (commented):
cool

PR #966, opened 2024-05-22
OLS-492: Combine response evaluation with e2e

asamal4:
## Description

Include response evaluation for every PR. This will run after current e2e test cases utilising same API.
Update existing QnA and cut-off score..


tests/scripts/utils.sh
```diff
@@ -110,14 +110,10 @@ function run_suite() {
...
-  if [ -z ${RESPONSE_EVALUATION:-} ]; then  
-    SUITE_ID=$1 TEST_TAGS=$2 MODEL=$8 ARTIFACT_DIR=$ARTIFACT_DIR make test-e2e
-  else
-    export SCENARIO="${SCENARIO:-with_rag}"
-    PROVIDER=$3 MODEL=$8 SCENARIO=$SCENARIO make response-quality-check
-  fi
+  # Run e2e tests
+  SUITE_ID=$1 TEST_TAGS=$2 MODEL=$8 ARTIFACT_DIR=$ARTIFACT_DIR make test-e2e
+  # Run response evaluation
+  PROVIDER=$3 MODEL=$8 make response-quality-check
```
bparees, line 116: this is an ok first step, but what i'd really like to see is for the response quality check tests to be part of `make test-e2e`.  What are the obstacles to doing that?

also perhaps we should rename it to "response-consistency-check" because right now the intent is just to ensure that we are getting the same responses we expect, not necessarily what the quality of those responses is, right?

bparees: > If I have understood correctly, Do you want to run both pytest and response evaluation python scripts under same make target test-e2e ?

yeah, that's the idea.  I want a test to run as part of test-e2e that checks whether the response values have changed significantly from what we expect (hence "consistency" and not "quality"), for the 10 or so sample Q+A pairs you pick out.

If you want to continue running a check across more Q+A pairs as part of a periodic job, i'm fine with that, but I want a minimal set to be tested on every PR so we have an easy way to see if a change has significantly changed the responses we're seeing, which will be cause for further investigation.

> BTW it was response-sanity-check earlier, woke check changed it to response-quality-check.

ah yes.

tests/e2e/test_api.py
```diff
@@ -89,6 +89,7 @@ def teardown_module(module):
     """Clean up the environment after all tests are executed."""
     if on_cluster:
         must_gather()
+        cluster_utils.delete_user("test-user")
```
bparees, line 92: if you're going to cleanup the things that were done in setup, there's more than this to address (there is also the metrics-test-user, and some rolebindings).

it doesn't generally matter since they just get reused as is so it's ok if they don't get cleaned up, but it would be better if we were thorough lest someone see this and think there's a specific reason we are cleaning up some objects but not others.

tests/scripts/utils.sh
bparees: sorry I guess i still wasn't clear.  What i really expected is that 
```
python -m pytest tests/e2e --durations=0 -o junit_suite_name="${SUITE_ID}" -m "${TEST_TAGS}" --junit-prefix="${SUITE_ID}" --junit-xml="${ARTIFACT_DIR}/junit_e2e_${SUITE_ID}.xml" --eval_model "${MODEL}"
```

(the command currently invoked by `make test-e2e`)

would execute some new tests that did the response consistency check.  So they would look like just another test that is part of our e2e suite.  Not that we'd run two separate commands under a single make target.  I want the consistency tests to be less "special" and just be part of our pytest e2e suite.

so basically convert "validate_response.py" into a pytest test.

does that make sense?  (both "do you understand it" and "is it reasonable from an implementation perspective"?

bparees: > You perhaps want this as another test case for consistency purpose.

yes, but not consistency for consistencies sake...  the desire is consistency with a goal of making it easier for the team as a whole to see/understand/contribute to these tests vs them being a separate special case that people need to know about and learn.  I may be over-valuing that cost, especially when it's just 2 right now, but the more "unique" ways we have to test/run tests, the harder it becomes for people to track down a failing one or understand how the test may work.

it's also because i want these results to appear in the junit output so they easily appear in the job results for PR authors (this is also sort of related to consistency).  That said, there are other ways to achieve that, like what i did here: 

https://github.com/openshift/lightspeed-service/blob/5f05c24812ac52df69e7542a410197d15177907a/tests/scripts/utils.sh#L1
https://github.com/openshift/lightspeed-service/blob/5f05c24812ac52df69e7542a410197d15177907a/tests/scripts/utils.sh#L106-L107

so at a minimum if we change nothing else i'd ask you do something to ensure failures show up as explicit junit failures of a particular test name to make it easy to trace back to the failing test/code.

> But I feel like response evaluation does enough tasks in terms of code/logic and justified to be a stand alone process/script. -> reading the file, iterate over QnA and call api, compare the texts and calculate score And finally save the result.

you and i may have a different perspective on how complex a test should reasonably be.  In OCP we had very complex e2e tests that were defined as ginkgo tests just like simple ones.  They would make significant changes to clusters, wait for the changes to take effect, take additional actions, verify behavior/changes, wait long periods of time for node reboots, etc.  So just because a test may have significant logic in it doesn't justify making it run separately in my mind.  You are welcome to create new modules that the test code invokes, if you want to keep the test code itself short, and also keep the overall test suite file short.

> For few e2e [test cases](https://github.com/openshift/lightspeed-service/blob/main/tests/e2e/test_api.py#L343) we are already calling core score calculation so that we are not doing just the sub-string match..

yes and that was another reason i was pushing to move these tests in so they would be aligned with those, i assumed they were overall very similar to those tests.  In fact i'd be interested in a brief explanation of what is actually different about how/what each one tests, or should we be removing the existing tests in favor of this new larger/more configurable set?

> And I will modify if my reasons don't make sense and you still strongly feel that we should add this as another test case.

I appreciate your reasons and I don't want to force this one way or another right now.  If you address the junit reporting aspect of it, then i think we can merge this as is (reverting to 2 make targets is fine.  I never intended to have you create one make target that looks as it currently does, it was always my intention to fold the tests themselves into the existing pytest invocation).

then we can continue the hopefully friendly debate about the cost/benefit of consistency and how complex a test should be :)

PR #967, opened 2024-05-22
OLS-622: Provider specific base class

tisnik:
## Description

- Provider-specific base class
- Common attributes moved there
- Credentials-loading code
- Updated unit tests

- [x] Refactor

* Related Issue #[OLS-622](https://issues.redhat.com//browse/OLS-622)

* [x] I have performed a self-review of my code.
* [x] PR has passed all pre-merge test jobs.
* [x] If it is a core feature, I have added thorough tests.

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -55,8 +55,11 @@ class ModelConfig(BaseModel):
     """Model configuration."""
 
     name: Optional[str] = None
+
+    # TODO: OLS-656 Switch OLS operator to use provider-specific configuration parameters
     url: Optional[AnyHttpUrl] = None
     credentials: Optional[str] = None
+
     context_window_size: int = -1  # need to be set later, based on model
     response_token_limit: int = -1  # need to be set later, based on model
     options: Optional[dict[str, Any]] = None
@@ -69,6 +72,8 @@ class ModelConfig(BaseModel):
         if data is None:
             return
         self.name = data.get("name", None)
+
+        # TODO: OLS-656 Switch OLS operator to use provider-specific configuration parameters
         self.url = data.get("url", None)
         self.credentials = _get_attribute_from_file(data, "credentials_path")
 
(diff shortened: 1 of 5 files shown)


ols/app/models/config.py
```diff
@@ -214,35 +214,42 @@ def validate_yaml(self) -> None:
...
+
+    credentials_path: str  # required attribute
+
+
+class AzureOpenAIConfig(ProviderSpecificConfig):
     """Configuration specific to Azure OpenAI provider."""
 
-    url: AnyHttpUrl  # required attribute
     deployment_name: str  # required attribute
+    credentials_path: Optional[str] = None
```
onmete, line 235: So because it is optional here, we can't have `credentials_path` in parent, right?

tests/unit/app/models/test_config.py
```diff
@@ -312,9 +312,175 @@ def test_provider_config():
...
+    with pytest.raises(ValidationError, match="url"):
+        ProviderConfig(
+            {
+                "name": "test_name",
+                "type": "bam",
+                "url": "test_url",
+                "credentials_path": "tests/config/secret.txt",
+                "project_id": "test_project_id",
+                "bam_config": {
+                    "credentials_path": "tests/config/secret.txt",
```
onmete, line 375: I'm confused a bit what should be the right way to provide `credentials_path` then? Should it be on the provider root or <provider>_config?

tisnik: in future, the provider config will be correct way. ATM we are going to support "old" style as well, because it needs to be coordinated with OLS operator?

ols/app/models/config.py
tisnik: exactly

tests/unit/app/models/test_config.py
onmete: I see, so is there some todo with jira about "disabling" the old way?

tisnik: added @onmete

PR #989, opened 2024-05-28
OLS-488: Refactor /liveness endpoint

onmete:
## Description

Refactor /liveness endpoint

## Type of change

- [x] Refactor


tisnik (commented):
it looks ok, but please make sure it won't break UI

PR #990, opened 2024-05-28
Trivial: missing return

tisnik:
## Description

Trivial: missing return


ols/src/query_helpers/docs_summarizer.py
```diff
@@ -27,6 +27,7 @@ def _get_model_options(
         if provider_config is not None:
             model_config = provider_config.models.get(self.model)
             return model_config.options
+        return None
```
onmete, line 30: I like it is explicitly here, but didn't you removed it from somewhere as "not needed"?

tisnik: AFAIK not for function returning "something" (as opposed to function returning "nothing" aka None)

PR #992, opened 2024-05-28
OLS-670: Readiness probe - cache check

onmete:
## Description

Refactor /readiness probe
- Add cache database check in readiness probe
- Remove readiness HEAD method.
  - AFAIK the readiness HEAD provides the same information as the liveness probe. It just tells if the endpoint is reachable. It seems redundant.

## Type of change

- [x] Refactor
- [x] New feature

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -196,26 +196,8 @@
                         "content": {
                             "application/json": {
                                 "schema": {
-                                    "$ref": "#/components/schemas/HealthResponse"
+                                    "$ref": "#/components/schemas/ReadinessResponse"
                                 }
-                            }
-                        }
-                    }
-                }
-            },
-            "head": {
-                "tags": [
-                    "health"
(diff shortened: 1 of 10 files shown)


onmete:
@xrajesh @bparees Would you know if the HEAD method for /readiness is requirement from k8s side? AFAIK it would serve the same purpose as /liveness, while the actual GET will tell you if the service is ready (and not just reachable).

onmete:
stage still down, trying again

/retest

xrajesh:
> @xrajesh @bparees Would you know if the HEAD method for /readiness is requirement from k8s side? AFAIK it would serve the same purpose as /liveness, while the actual GET will tell you if the service is ready (and not just reachable).

No need for HEAD - HTTP probes use GET

onmete:
Closing for now. We'll redo with OLS-670

PR #1001, opened 2024-05-29
Report portal enabling

JoaoFula:
On boarding report portal

I created the dependency in pyproject.toml for the report portal pytest integration
in the conftest I readied the reading of the tags during the test calling, reading the envs, preparing the archive based on the xml report (that's saved in ARTIFACT_DIR) and uploading it to the s3 bucket to be redirected to RP.

I also added the curling of the binary file to upload to the s3 bucket and added the options during the call in the makefile.


pyproject.toml
```diff
@@ -74,6 +74,7 @@ dependencies = [
     "kubernetes==29.0.0",
     "pytest-asyncio==0.23.6",
     "psycopg2-binary==2.9.9",
+    "pytest-reportportal==5.4.1",
```
tisnik, line 77: it is dev dependency, is not it?

JoaoFula: what do you mean by that? is this the wrong place for the testing dependencies?

tisnik: yes

tisnik: if you move it to dev deps, the tests jobs should start to pass

JoaoFula: Hi @tisnik sorry to bother but where are the dev dependencies defined?

pyproject.toml
```diff
@@ -109,8 +112,6 @@ test-unit = "pdm run make test-unit"
 test-integration = "pdm run make test-integration"
 test-e2e = "pdm run make test-e2e"
 coverage-report = "pdm run make coverage-report"
-unit-tests-coverage-report = "pdm run make unit-test-coverage-report"
```
tisnik, line 112: I'd keep these targets, I use them from time to time

tisnik (commented):
It looks ok, just please don't delete those two aliases

pyproject.toml
JoaoFula: for sure. This was not intended, good catch.

PR #1002, opened 2024-05-29
OLS-641: Access token retrieval code

tisnik:
## Description

- Access token retrieval code
- Backed by unit tests

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #[OLS-641](https://issues.redhat.com//browse/OLS-641)

### DIFF
--- a/ols/src/llms/providers/azure_openai.py
+++ b/ols/src/llms/providers/azure_openai.py
@@ -3,10 +3,13 @@
 import logging
 from typing import Any, Optional
 
+from azure.core.credentials import AccessToken
+from azure.identity import ClientSecretCredential
 from langchain.llms.base import LLM
 from langchain_openai import AzureChatOpenAI
 
 from ols import constants
+from ols.app.models.config import AzureOpenAIConfig
 from ols.src.llms.providers.provider import LLMProvider
 from ols.src.llms.providers.registry import register_llm_provider_as
 
@@ -26,17 +29,16 @@ class AzureOpenAI(LLMProvider):
         self.url = str(self.provider_config.url or self.url)
         self.credentials = self.provider_config.credentials
         deployment_name = self.provider_config.deployment_name
+        azure_config = self.provider_config.azure_config
 
         # provider-specific configuration has precendence over regular configuration
-        if self.provider_config.azure_config is not None:
-            azure_config = self.provider_config.azure_config
+        if azure_config is not None:
             self.url = str(azure_config.url)
             deployment_name = azure_config.deployment_name
             self.credentials = azure_config.api_key
 
-        return {
+        default_parameters = {
             "azure_endpoint": self.url,
-            "api_key": self.credentials,
             "api_version": "2024-02-01",
             "deployment_name": deployment_name,
             "model": self.model,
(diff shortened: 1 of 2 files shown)


ols/src/llms/providers/azure_openai.py
```diff
@@ -52,6 +54,47 @@ def default_params(self) -> dict[str, Any]:
...
         }
 
+        if self.credentials is not None:
+            # if credentials with API key is set, use it to call Azure OpenAI endpoints
+            default_parameters["api_key"] = self.credentials
+        else:
+            # credentials for API key is not set
+            # -> it is needed to retrieve token first
+            access_token = self.retrieve_access_token(azure_config)
+            default_parameters["azure_ad"] = access_token.token
```
onmete, line 64: What `azure_ad` means?

ols/src/llms/providers/azure_openai.py
```diff
@@ -52,6 +54,47 @@ def default_params(self) -> dict[str, Any]:
...
+        )
+        return credential.get_token("https://cognitiveservices.azure.com/.default")
+
+
+def raise_missing_attribute_error(attribute_name: str) -> None:
+    """Raise exception when some attribute is missing in configuration."""
+    raise ValueError(
+        f"{attribute_name} should be set in azure_openai_config in order to retrieve "
+        "access token."
+        ""
```
onmete, line 99: `""`?

ols/src/llms/providers/azure_openai.py
```diff
@@ -52,6 +54,47 @@ def default_params(self) -> dict[str, Any]:
...
+        if azure_config.client_id is None:
+            raise_missing_attribute_error("client_id")
+        if azure_config.client_secret is None:
+            raise_missing_attribute_error("client_secret")
+
+        # everything is there, try to retrieve credential
+        credential = ClientSecretCredential(
+            azure_config.tenant_id, azure_config.client_id, azure_config.client_secret
+        )
+        return credential.get_token("https://cognitiveservices.azure.com/.default")
```
onmete, line 91: Could it fail? Should we catch an exception in that case?

tisnik: it can fail and will fail, is it a problem?

ols/src/llms/providers/azure_openai.py
tisnik: ah some black leftover, will remove it

ols/src/llms/providers/azure_openai.py
tisnik: I've updated PR with better info (and realized there are two such parameters, switched to longer one)

ols/src/llms/providers/azure_openai.py
onmete: Maybe logging something like the service is not available instead of hard crash?

tisnik: ok will change it. It will fail later, but that's another story

tisnik: updated @onmete

ols/src/llms/providers/azure_openai.py
```diff
@@ -52,6 +54,55 @@ def default_params(self) -> dict[str, Any]:
...
+            raise_missing_attribute_error("client_secret")
+
+        # everything is there, try to retrieve credential
+        try:
+            credential = ClientSecretCredential(
+                azure_config.tenant_id,
+                azure_config.client_id,
+                azure_config.client_secret,
+            )
+            return credential.get_token("https://cognitiveservices.azure.com/.default")
```
bparees, line 97: @tisnik does this access token ever expire such that OLS might need to retrieve a new one?

tisnik: yes that expire after some time. it is configurable, but definitely there. Fortunately each request basically refresh the token

bparees: > Fortunately each request basically refresh the token

meaning each request retrieves a new token?  or each request extends the expiration of the token?

bparees: are we getting a new access token every time because that is part of loading the llm and we load the llm on every OLS request?

tisnik: yes, basically

tisnik: of course in "normal" REST API I'd optimize it probably. but in case of LLM calls that are pretty costly itself (time+CPU time), dunno if it's worth

bparees: probably something to mark as a TODO followup.  I agree it won't be a bottleneck but it's one more request flow that can go wrong if something is down and we really shouldn't have to be getting a new token on every request.  I assume Azure would prefer we not get a new token every time also, but i guess it also somewhat depends on the expiration period of the token.

PR #1004, opened 2024-05-29
rename & organize e2e utils

asamal4:
## Description

Move all e2e utils files to e2e/utils folder.
rename test_decorators* file as this doesn't contain any test case..
rename response evaluation .py file; [OLS-492](https://issues.redhat.com//browse/OLS-492)


tests/e2e/test_api.py
```diff
@@ -16,26 +16,25 @@
...
     BASIC_ENDPOINTS_TIMEOUT,
     CONVERSATION_ID,
     LLM_REST_API_TIMEOUT,
     NON_LLM_REST_API_TIMEOUT,
 )
-from tests.scripts.must_gather import must_gather
-from tests.scripts.validate_response import ResponseEvaluation
-
-from .postgres_utils import (
+from tests.e2e.utils.decorators_utils import retry
```
tisnik, line 30: we don't have to repeat "utils" two times probably. For example in Go ecosystem it is not allowed.

tisnik (commented):
look ok in overall, have one nitpick

tests/e2e/test_api.py
tisnik: nice!

tisnik (commented):
looks good!

PR #1007, opened 2024-05-30
OLS-488: Readiness probe - index loading check

onmete:
## Description

Add index check to readiness probe.

## Type of change

- [x] Refactor
- [x] New feature

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -224,24 +224,6 @@
                         }
                     }
                 }
-            },
-            "head": {
-                "tags": [
-                    "health"
-                ],
-                "summary": "Liveness Probe Head Method",
-                "description": "Live status of service.",
-                "operationId": "liveness_probe_head_method_liveness_head",
-                "responses": {
-                    "200": {
-                        "description": "Successful Response",
-                        "content": {
-                            "application/json": {
-                                "schema": {}
-                            }
-                        }
-                    }
-                }
             }
         },
         "/metrics": {
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -196,26 +196,8 @@
                         "content": {
                             "application/json": {
                                 "schema": {
-                                    "$ref": "#/components/schemas/HealthResponse"
+                                    "$ref": "#/components/schemas/ReadinessResponse"
                                 }
-                            }
-                        }
-                    }
-                }
-            },
-            "head": {
-                "tags": [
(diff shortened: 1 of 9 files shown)


ols/app/endpoints/health.py
```diff
@@ -7,28 +7,27 @@
...
-#     "status": "unhealthy",
-#     "version": "1.0.0",
-#     "dependencies": {
-#         "database": "healthy",
-#         "externalApi": "unhealthy"
-#     }
-# }
+def index_is_ready() -> bool:
+    """Check if the index is loaded."""
+    if config._rag_index is None:
```
xrajesh, line 18: @onmete When there is no RAG content  configured , the server can proceed without RAG content . In that case the readiness can return false .

ols/app/models/models.py
```diff
@@ -342,6 +342,35 @@ class HealthResponse(BaseModel):
     }
 
 
+class ReadinessResponse(BaseModel):
+    """Model representing a response to a readiness request.
+
+    Attributes:
+        ready: The readiness of the service.
+        reason: The reason for the readiness.
```
xrajesh, line 350: @onmete do we want the reason field ? just thinking if we can keep in sync with liveness, unless there is a use for this.

onmete: Telling why service is not ready sounds like a good reason to me. But I didn't find any required schema for these two endpoints, so I guess it is on our needs. Do you have concerns with this schema?

ols/app/endpoints/health.py
onmete: Right, updated.

docs/openapi.json
```diff
@@ -196,30 +196,12 @@
...
                             "application/json": {
                                 "schema": {
-                                    "$ref": "#/components/schemas/HealthResponse"
+                                    "$ref": "#/components/schemas/ReadinessResponse"
                                 }
                             }
                         }
                     }
                 }
-            },
```
tisnik, line 205: are not HEAD HTTP methods used regularly, especially for liveness and readiness probes? @JoaoFula ?

tisnik (commented):
in overall it looks ok

docs/openapi.json
JoaoFula: yeah, I believe so. Don't normally use it myself but that's their purpose.

PR #1011, opened 2024-05-31
add questions pool for reference

asamal4:
## Description

Adding few questions just for future reference. [OLS-491](https://issues.redhat.com//browse/OLS-491)

- keeping questions with typos.
- keeping similar questions with minor variation.
- Discarded multiple lines questions.


tisnik (commented):
it looks corect!

PR #1015, opened 2024-06-03
OLS-673: Fix the error "error: Too many arguments for "SummarizerResponse" [call-arg]"

tisnik:
## Description

[OLS-673](https://issues.redhat.com//browse/OLS-673): Fix the error "error: Too many arguments for "SummarizerResponse"  [call-arg]"

Fixes: [OLS-673](https://issues.redhat.com//browse/OLS-673)

### DIFF
--- a/ols/app/models/models.py
+++ b/ols/app/models/models.py
@@ -1,9 +1,14 @@
 """Data models representing payloads for REST API calls."""
 
-from typing import Any, Optional, Self
+from typing import TYPE_CHECKING, Any, Optional, Self
 
 from pydantic import BaseModel, field_validator, model_validator
-from pydantic.dataclasses import dataclass
+
+# fixes OLS-673: Type checker reports "too many arguments" when the dataclass is constructed
+if TYPE_CHECKING:
+    from dataclasses import dataclass
+else:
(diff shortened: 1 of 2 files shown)


ols/app/models/models.py
```diff
@@ -1,9 +1,14 @@
...
+from typing import TYPE_CHECKING, Any, Optional, Self
 
 from pydantic import BaseModel, field_validator, model_validator
-from pydantic.dataclasses import dataclass
+
+# fixes OLS-673: Type checker reports "too many arguments" when the dataclass is constructed
+if TYPE_CHECKING:
+    from dataclasses import dataclass
+else:
+    from pydantic.dataclasses import dataclass
```
onmete, line 11: Even though the dataclass from pydantic should have the same interface as Python's native one, I would rather ignore the error instead of importing/using something "potentially else". WDYT?

tisnik: that would be ok. There are TODOs already added there so we won't forget about those.

bparees: yeah i'm inclined to agree with @onmete here

tisnik:
closing as it is no longer needed -> the proper fix took place

PR #1018, opened 2024-06-03
OLS-673: Enable Pydantic plugin for Mypy

tisnik:
## Description

See also: https://docs.pydantic.dev/latest/integrations/mypy/

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Related Issue #OLS-673
- Closes #


bparees:
would this mean we can/should remove the "make check-types" target because "make verify" is doing the same thing?

tisnik:
> would this mean we can/should remove the "make check-types" target because "make verify" is doing the same thing?

nope ATM. I need to figure out howto use the plugin correctly and especially find out why CI behaves differently than local run (with exactly the same packages+versions installed)

bparees:
>> would this mean we can/should remove the "make check-types" target because "make verify" is doing the same thing?
> nope ATM. I need to figure out howto use the plugin correctly and especially find out why CI behaves differently than local run (with exactly the same packages+versions installed)

but ultimately is that where we end up?  or you envision us continuing to have both invocations/mechanisms?

tisnik:
> > > would this mean we can/should remove the "make check-types" target because "make verify" is doing the same thing?
> > > nope ATM. I need to figure out howto use the plugin correctly and especially find out why CI behaves differently than local run (with exactly the same packages+versions installed)
> 
> but ultimately is that where we end up? or you envision us continuing to have both invocations/mechanisms?

locally I use both, based on requirements so my preference will be to keep them separate.

PR #1030, opened 2024-06-04
OLS-624: Disable additional properties + basic REST API tests with attachments

tisnik:
## Description

Disable additional properties + basic REST API tests with attachments

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [x] e2e tests

- Related Issue #[OLS-624](https://issues.redhat.com//browse/OLS-624)
- Closes #

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -981,6 +981,175 @@ def _perform_query(client, conversation_id, query):
     print(vars(response))
 
 
+@retry(max_attempts=3, wait_between_runs=10)
+def test_valid_question_with_empty_attachment_list() -> None:
+    """Check the REST API /v1/query with POST HTTP method using empty attachment list."""
+    endpoint = "/v1/query"
+
+    with metrics_utils.RestAPICallCounterChecker(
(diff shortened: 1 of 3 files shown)


tests/e2e/test_api.py
```diff
@@ -981,6 +981,95 @@ def _perform_query(client, conversation_id, query):
...
+        response = client.post(
+            endpoint,
+            json={
+                "conversation_id": "",
+                "query": "what is kubernetes?",
+                "attachments": [
+                    {
+                        "attachment_type": "log",
+                        "content_type": "text/plain",
+                        "content": "this is attachment",
```
bparees, line 1025: i know the implementation isn't in place yet but ultimately we're going to need a test that actually ensures the attachment content gets used in the prompt to the llm, so when you add the implementation part that actually adds the attachment content to the prompt, please keep that in mind (probably best to just update this test at that point, rather than adding another one)

tisnik: yes, it was/is my plan

PR #1035, opened 2024-06-05
OLS-663: Bug: When reference config is missing in config, use None instead of empty…

onmete:
## Description

When reference config is missing in config, use None instead of empty object.

This behavior resulted in the following problem:

When config entry is missing, the empty `ReferenceContent` is used (we put it there in the `__init__`) instead of `None` as specified as the default value in the `OLSConfig`. The app started without RAG - that is valid, but the /readiness checks that `ols_config.reference_content` is explicitly `None` (as it should be) - resulting in the endpoint reporting false as the index is not ready (mistake).

## Type of change

- [x] Bug fix

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -859,7 +859,8 @@ class OLSConfig(BaseModel):
             data.get("conversation_cache", None)
         )
         self.logging_config = LoggingConfig(data.get("logging_config", None))
-        self.reference_content = ReferenceContent(data.get("reference_content", None))
+        if data.get("reference_content") is not None:
+            self.reference_content = ReferenceContent(data.get("reference_content"))
         self.default_provider = data.get("default_provider", None)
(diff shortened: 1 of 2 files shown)


ols/app/models/config.py
```diff
@@ -859,7 +859,8 @@ def __init__(self, data: Optional[dict] = None) -> None:
             data.get("conversation_cache", None)
         )
         self.logging_config = LoggingConfig(data.get("logging_config", None))
-        self.reference_content = ReferenceContent(data.get("reference_content", None))
+        if data.get("reference_content"):
```
tisnik, line 862: I'd add `is not None`, lgtm otherwise

onmete:
cc @bparees This wouldn't have happened if we had done the changes from https://issues.redhat.com/browse/OLS-259
So maybe the time when we can tackle it is closing in?:)

PR #1044, opened 2024-06-06
Adding fake llm provider

vishnuchalla:
## Description

Adding fake LLM Provider to use it for load testing OLS standalone.

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

- Related Story: https://issues.redhat.com/browse/OLS-687

## Testing
Tested in local using `make test-unit` also verified the same with operator. Was able to fetch the pre-configured response.
```
vchalla@vchalla-thinkpadp1gen2:~$ oc rsh pod/lightspeed-app-server-66bc495fb4-h97nb
Defaulted container "lightspeed-service-api" out of: lightspeed-service-api, lightspeed-service-user-data-collector
sh-5.1$ curl -X 'POST' 'http://0.0.0.0:8080/v1/query' -H 'accept: application/json' -H 'Content-Type: application/json' -H "Authorization: Bearer TOKEN" -d '{"query": "write a deployment yaml for the mongodb image"}'
{"conversation_id":"c1bd68f1-bab2-406a-9a25-e7c8d51d8c29","response":"This is a preconfigured fake response.","referenced_documents":[{"docs_url":"https://docs.openshift.com/container-platform/4.15/getting_started/openshift-cli.html","title":"Creating and building an application using the CLI"}],"truncated":false}sh-5.1$ 

```


ols/app/models/config.py
```diff
@@ -1001,13 +1001,27 @@ def __eq__(self, other: object) -> bool:
...
+        if selected_default_provider == constants.PROVIDER_FAKE:
+            model_config = ModelConfig()
+            model_config.name = constants.FAKE_MODEL
+            model_config.context_window_size = constants.DEFAULT_CONTEXT_WINDOW_SIZE
+            provider_config = ProviderConfig()
+            provider_config.name = constants.PROVIDER_FAKE
+            provider_config.type = constants.PROVIDER_FAKE
+            provider_config.models[constants.FAKE_MODEL] = model_config
+            self.ols_config.default_model = constants.FAKE_MODEL
+            self.llm_providers.providers[constants.PROVIDER_FAKE] = provider_config
```
bparees, line 1020: this is a validation function, it should not be setting config values for the system, only checking them.

i'm also not clear why we need to treat the fake provider differently than any of the real providers.

ols/src/llms/llm_loader.py
```diff
@@ -43,7 +43,7 @@ def _resolve_provider_config(
 
     provider_config = providers_config.providers.get(provider)
 
-    if model not in provider_config.models:
+    if provider != constants.PROVIDER_FAKE and model not in provider_config.models:
```
bparees, line 46: why do we need to special case this?  Can we not just provide a full fake provider/model configuration?

or at least during config loading, if we see the requested provider is the fake one, populate the rest of the configuration as if it was a normal provider?  (using appropriately fake values, of course)?  Which sort of seems to be what you're trying to do with the validation logic you added, but i don't think such logic belongs in "validation"

tisnik (changes requested):
I agree with @bparees 
The fake provider can be implemented as yet another provider w/o the need to have 'special cases' spread over the sources. If you look at (for example) `BAM` provider, it would be possible to mimic it's behaviour almost completely - just the one line that returns it's instance will return some mock (I expect the mock will be HTTP server instance, right?)

ols/app/models/config.py
vishnuchalla: Earlier I was thinking not to even let a user specify fake provider configuration in the provider list itself. But looks like I made it complicated and its a bit odd with the current flow. So removing this snippet to keep it simple.

ols/src/llms/llm_loader.py
vishnuchalla: Removed this special case according to the reason stated above.

vishnuchalla:
@tisnik Restructured the code to avoid spreading across. This mock (i.e [fakellm](https://api.python.langchain.com/en/latest/llms/langchain_community.llms.fake.FakeListLLM.html)) is provided by langchain itself which is solely used for testing. Its not a http endpoint actually, but something that returns a instant preconfigured response.

tisnik (commented):
it looks much better, thank you

PR #1045, opened 2024-06-06
OLS-680: Read API token from default file if just directory is specified

tisnik:
## Description

Read API token from default file if just directory is specified

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #OLS-680
- Closes #


bparees:
/hold

sorry i have a review i'm still working on

ols/app/models/config.py
```diff
@@ -361,8 +380,10 @@ def set_provider_specific_configuration(self, data: dict) -> None:
                     azure_config = data.get("azure_openai_config")
                     self.check_provider_config(azure_config)
                     if azure_config is not None:
-                        azure_config["client_secret"] = _get_attribute_from_file(
-                            azure_config, "client_secret_path"
+                        azure_config["client_secret"] = _read_secret(
+                            azure_config,
+                            "client_secret_path",
```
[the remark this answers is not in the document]
xrajesh: +1

ols/app/models/config.py
```diff
@@ -36,6 +36,21 @@ def _get_attribute_from_file(data: dict, file_name_key: str) -> Optional[str]:
     return None
 
 
+def _read_secret(data: dict, path_key: str, default_filename: str) -> Optional[str]:
+    """Read secret from file on given path or from filename if path points do directory."""
```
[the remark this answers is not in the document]
tisnik: fixed @bparees

ols/app/models/config.py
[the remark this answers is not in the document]
tisnik: refactored too @bparees @xrajesh

ols/app/models/config.py
```diff
@@ -361,8 +391,23 @@ def set_provider_specific_configuration(self, data: dict) -> None:
...
+                            azure_config,
+                            constants.CREDENTIALS_PATH_SELECTOR,
+                            constants.AZURE_CLIENT_ID_FILENAME,
+                            raise_on_error=False,
+                        )
+                        azure_config["client_secret"] = _read_secret(
+                            azure_config,
+                            constants.CREDENTIALS_PATH_SELECTOR,
+                            constants.AZURE_CLIENT_SECRET_FILENAME,
+                            raise_on_error=False,
```
bparees, line 410: i think there is a flaw here.....if someone sets "credentials_path" to "/foo/bar/somefilename", then "tenant_id", "client_id", and "client_secret" are all going to get populated w/ the value of whatever is in "somefilename", no?

and then "api_key" will also get set to the filename value.  Which i guess means we'll prefer to use the "api_key" value as the credentials, so maybe it doesn't matter that much?  But it still seems bad that we'd populate the other dictionary keys from that file.

tisnik: indeed, you are right. Jmm ok, it would be possible to have keyword flag "must_be_directory" perhaps (just for Azure OpenAI-related secret).

ols/app/models/config.py
```diff
@@ -36,6 +36,32 @@ def _get_attribute_from_file(data: dict, file_name_key: str) -> Optional[str]:
...
+    try:
+        with open(filename, mode="r", encoding="utf-8") as f:
+            return f.read().rstrip()
+    except IOError as e:
+        # some files with secret must exist, so for such cases it is time
+        # to inform about improper configuration
+        if raise_on_error:
+            raise
+        # no logging configured yet
+        print("Problem reading secret from file {filename}:", e)
```
xrajesh, line 61: If the user has a typo on the name of the secret filed `apitoken` we will get this error. Can we also the write the log saying `Verify the provider secret contains` `apitoken`

ols/app/models/config.py
xrajesh: `azure_openai_config` section is new and OLS can expect  `credentials_path` to be a directory 
(1) Fail the validation if it IS a file 
else (2) check the directory for the presence of `apitoken` file , if `apitoken`  is present use that as credential 
else (3) If `apitoken` check for `tenant_id` `client_id` and `client_secret` - use them if all 3 are present.
else (4) Fail the credential validation. 
From the operator stand point , we are only setting a directory - and the previous versions of operator we did not have this `azure_openai_config` section.

ols/src/llms/providers/azure_openai.py
```diff
@@ -87,7 +87,7 @@ def retrieve_access_token(
         if azure_config.client_id is None:
             raise_missing_attribute_error("client_id")
         if azure_config.client_secret is None:
-            raise_missing_attribute_error("client_secret")
+            raise_missing_attribute_error("credentials_path")
```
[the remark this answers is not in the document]
tisnik: updated accordingly @xrajesh

ols/app/models/config.py
tisnik: added there @xrajesh

ols/app/models/config.py
tisnik: @bparees the flaw should be fixed now.
@xrajesh I guess it's there now (I guess that by [3] you mean if `not apitoken`

PR #1047, opened 2024-06-07
Add resource limits to E2E testing

onmete:
## Description

Add resource limits to E2E testing as defined by our operator.

## Type of change

- [x] Refactor


bparees:
/lgtm
/approve

but we should lower these limits once we resolve the underlying cause of the high memory usage

PR #1048, opened 2024-06-07
OLS-533: Fix diagram path in data collection docs

onmete:
## Description

Fix diagram path in data collection docs

## Type of change

- [x] Bug fix


tisnik (commented):
looks sane

PR #1051, opened 2024-06-07
OLS-692: OOM fix

tisnik:
## Description

OOM fix

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/user_data_collection/data_collector.py
+++ b/ols/user_data_collection/data_collector.py
@@ -342,6 +342,9 @@ def gather_ols_user_data(data_path: str) -> None:
                 logger.error(
                     f"{e.__class__.__name__} - upload and data removal canceled"
                 )
+
+            # close the tarball to release mem
+            tarball.close()
     else:
         logger.info(f"'{data_path}' contains no data, nothing to do...")
 


ols/user_data_collection/data_collector.py
```diff
@@ -342,6 +342,9 @@ def gather_ols_user_data(data_path: str) -> None:
                 logger.error(
                     f"{e.__class__.__name__} - upload and data removal canceled"
                 )
+
+            # close the tarball to release mem
+            tarball.close()
```
bparees, line 347: this seems like a good change but i'm not convinced this was the cause of the memory utilization.  did you do some tests that confirmed it?

the collector only runs every 2 hours (the first run is something like 5 mins after startup).  So it's hard to see how this could cause a rapid runaway of memory usage in the first few minutes after startup.

also the archives themselves are pretty small, so even if we were leaking the size of the tarfile, that would not explain hundreds of megs of usage, i don't think.

bparees: (also i'd sort of expect python to close the tarball when the variable goes out of scope, but that may not be a valid assumption)

bparees:
/lgtm
/approve

we certainly want this fix, i just am not convinced it's the only fix needed to resolve the bug.

PR #1052, opened 2024-06-09
OLS-581: Move the program logic to wait for OLS 

Nehanth:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing


tests/e2e/test_api.py
```diff
@@ -34,10 +35,14 @@
 from tests.e2e.utils.response_evaluation import ResponseEvaluation
 from tests.scripts.must_gather import must_gather
 
+sys.path.append(Path(__file__).parent.parent.parent.as_posix())
```
bparees, line 38: not clear to me why this is needed?  we are already importing other `test.scripts` content without this

tests/e2e/test_api.py
```diff
@@ -76,6 +81,14 @@ def setup_module(module):
 
         client = client_utils.get_http_client(ols_url, token)
         metrics_client = client_utils.get_http_client(ols_url, metrics_token)
+
+        # Wait for OLS to be ready
+        success = wait_for_ols(ols_url)
+        if not success:
+            generate_junit_report(os.getenv("SUITE_ID"), False)
```
bparees, line 88: does the pytest.fail below not already guarantee a junit test entry being produced?  (put another way, what does the junit + test job output look like if you don't explicitly do this?)

tests/scripts/utils.sh
```diff
@@ -100,13 +85,15 @@ function run_suite() {
...
   install_ols "$1" "$3" "$4" "$5" "$6" "$7" "$8" "$9"
 
-  # wait for the ols api server to come up
-  wait_for_ols "$OLS_URL"
-  if [ $? -ne 0 ]; then
-    echo "Timed out waiting for OLS to become available"
-    echo "${JUNIT_TEMPLATE}" | sed "s/SUITE_ID/${1}/g" > $ARTIFACT_DIR/junit_setup_${1}.xml
+ # Determine the hostname for the ols route
+  export OLS_URL=https://$(oc get route ols -o jsonpath='{.spec.host}')
+  echo "OLS_URL set to $OLS_URL"
```
bparees, line 90: to be fair this PR does what the ticket asked, but would it be terribly difficult to also move this logic into the setup while you're doing the refactoring?

tests/scripts/wait_for_ols.py
bparees, line 1: instead of treating this as a standalone "script", i would just move it into tests.e2e.utils (either in one of the existing files or a new one).  

Is there an expectation this will get invoked as a standalone script, as compared to just using the utility function(s) it provides?

tests/scripts/wait_for_ols.py
```diff
@@ -0,0 +1,101 @@
...
+        failure = ElementTree.SubElement(
+            testcase, "failure", message="OLS failed to start up"
+        )
+        failure.text = "OLS did not become available in time"
+    tree = ElementTree.ElementTree(testsuite)
+    artifact_dir = os.getenv("ARTIFACT_DIR", ".")
+    tree.write(f"{artifact_dir}/junit_setup_{suite_id}.xml")
+
+
+if __name__ == "__main__":
```
bparees, line 83: similar to the comment above about moving this whole thing into/under our tests.e2e.utils module, does it actually need logic to be invokable as a standalone script?  It doesn't look like this PR is doing so.  I'm guessing you just created this while you were doing some debugging?

bparees (commented):
a couple structural questions.

also multiple commits is fine while you're working on the PR but please be prepared to squash your commits prior to us merging this.  (We only retain separate commits in a PR if there is a good reason to keep certain changes in their own commits, but if it takes you 10 commits to iterate on a solution, we don't keep those separate when we merge)

tests/e2e/test_api.py
Nehanth: it fails in CI If I do not do that for some reason

bparees: @tisnik maybe you can help understand/explain why this is needed?

tests/e2e/test_api.py
Nehanth: oh okay, Will just remove it then

tests/e2e/utils/wait_for_ols.py
```diff
@@ -0,0 +1,54 @@
...
+import sys
+import time
+import warnings
+from pathlib import Path
+
+import requests
+from requests.exceptions import SSLError
+from urllib3.exceptions import InsecureRequestWarning
+
+sys.path.append(Path(__file__).resolve().parents[1].as_posix())
```
bparees, line 15: another thing i would not expect to be needed, especially now that this is a utility/helper, not a standalone script invocation.

tests/e2e/test_api.py
```diff
@@ -74,8 +77,24 @@ def setup_module(module):
         else:
             print("Setting up for standalone test execution\n")
 
+        # Determine the hostname for the OLS route
+        ols_url = os.popen("oc get route ols -o jsonpath='{.spec.host}'").read().strip()
```
bparees, line 81: we've got some helper functions for interacting with `oc` under "cluster.py", please take a look at that so we can do this in a consistent way

tests/scripts/utils.sh
```diff
@@ -100,15 +85,10 @@ function run_suite() {
   
   install_ols "$1" "$3" "$4" "$5" "$6" "$7" "$8" "$9"
 
-  # wait for the ols api server to come up
-  wait_for_ols "$OLS_URL"
-  if [ $? -ne 0 ]; then
-    echo "Timed out waiting for OLS to become available"
-    echo "${JUNIT_TEMPLATE}" | sed "s/SUITE_ID/${1}/g" > $ARTIFACT_DIR/junit_setup_${1}.xml
```
bparees, line 107: JUNIT_TEMPLATE(actually apparently it was JUNIT_SETUP_TEMPLATE) isn't used anymore, so you can delete it.

But fundamentally also the junit failure that JUNIT_TEMPLATE was used to generate, needs to be replicated by your new logic (i.e. we should still see a junit test failure representing the fact that OLS didn't become ready)

ironically also i think there was a bug here because it should have been JUNIT_SETUP_TEMPLATE, not JUNIT_TEMPLATE.  In any case the template is not needed now because replacing that template/generated junit test result is part of what your work is intended to replace.

tests/e2e/test_api.py
```diff
@@ -55,14 +57,24 @@
...
 OLS_COLLECTOR_DISABLING_FILE = OLS_USER_DATA_PATH + "/disable_collector"
 
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
+)
+logger = logging.getLogger(__name__)
+
 
+# ruff: noqa: S605, S607
```
tisnik, line 60: it is usually better to put these comments after lines with "problematic" commands

tests/e2e/test_api.py
tisnik: it might be saver to use `subprocess.Popen` with all the error checking etc.

tests/e2e/test_api.py
```diff
@@ -72,13 +84,41 @@ def setup_module(module):
...
+            logger.info("Setting up for standalone test execution")
             print("Setting up for standalone test execution\n")
 
+        # Determine the hostname for the OLS route
+        ols_url = os.popen("oc get route ols -o jsonpath='{.spec.host}'").read().strip()
+        if not ols_url.startswith("http"):
+            ols_url = f"https://{ols_url}"
+
+        logger.info(f"OLS_URL set to {ols_url}")
+        print(f"OLS_URL set to {ols_url}")
```
tisnik, line 85: log/print the same thingy - needed?

tests/e2e/test_api.py
```diff
@@ -72,13 +84,41 @@ def setup_module(module):
...
+        os.environ["OLS_URL"] = ols_url
+
         client = client_utils.get_http_client(ols_url, token)
         metrics_client = client_utils.get_http_client(ols_url, metrics_token)
+
+        # Wait for OLS to be ready
+        logger.info(f"Waiting for OLS to be ready on provider: {provider}...")
+        print(f"Waiting for OLS to be ready on provider: {provider}...")
```
tisnik, line 92: dtto

tests/e2e/test_api.py
```diff
@@ -72,13 +84,41 @@ def setup_module(module):
...
+        success = wait_for_ols(ols_url)
+        if not success:
+            failure_message = (
+                f"OLS did not become available in time on provider: {provider}"
+            )
+            logger.error(failure_message)
+            print(failure_message)
```
tisnik: dtto

tests/e2e/test_api.py
```diff
@@ -72,13 +84,41 @@ def setup_module(module):
...
+            must_gather()
+            pytest.exit(f"OLS did not become available in time on provider: {provider}")
+        logger.info("OLS is ready")
+        print("OLS is ready")
```
tisnik, line 98: dtto

tests/e2e/test_api.py
```diff
@@ -72,13 +84,41 @@ def setup_module(module):
...
+
     except Exception as e:
-        print(f"Failed to setup ols access: {e}")
-        sys.exit(1)
+        failure_message = f"Failed to setup ols access: {e}"
+        logger.error(failure_message)
+        print(failure_message)
```
tisnik: dtto

tests/e2e/utils/wait_for_ols.py
tisnik: hmm +1 tbh

tests/e2e/utils/wait_for_ols.py
```diff
@@ -0,0 +1,54 @@
...
+    attempts = int(timeout / interval)
+    for attempt in range(1, attempts + 1):
+        print(f"Checking OLS readiness, attempt {attempt} of {attempts}")
+        try:
+            response = requests.get(f"{url}/readiness", verify=True, timeout=5)
+            if response.status_code == requests.codes.ok:
+                print("OLS is ready")
+                return True
+        except SSLError:
+            print("SSL error detected, retrying without SSL verification")
```
tisnik, line 42: so does it mean that even for https it can skip SSL?

tisnik (commented):
some comments

tests/e2e/conftest.py
```diff
@@ -1,17 +1,55 @@
...
+    # If OLS did come up cleanly during setup, then just generate normal test reports for all tests
+    return TestReport.from_item_and_call(item, call)
+
 
 def pytest_addoption(parser):
     """Argument parser for pytest."""
     parser.addoption(
         "--eval_provider",
-        default="openai",
+        default="watsonx",
```
bparees, line 46: not sure where this change came from but i don't think it should be part of this PR

tests/e2e/conftest.py
```diff
@@ -1,17 +1,55 @@
...
         type=str,
         help="Provider name, currently used only to form output file name.",
     )
     parser.addoption(
         "--eval_model",
-        default="gpt",
+        default="ibm/granite-13b-chat-v2",
```
bparees, line 52: not sure where this change came from but i don't think it should be part of this PR

tests/e2e/conftest.py
```diff
@@ -26,12 +64,6 @@ def pytest_addoption(parser):
         default=None,
         help="Ids of questions to be validated. Check json file for valid ids.",
     )
-    parser.addoption(
-        "--eval_threshold",
-        type=float,
-        default=None,
-        help="Threshold value to be used for similarity score.",
-    )
```
bparees, line 34: not sure where this change came from but i don't think it should be part of this PR

tests/e2e/pytest.ini
```diff
@@ -8,4 +8,4 @@ markers =
 	any
 	cluster
 	rag
-	response_evaluation
+	response_evaluation
```
bparees, line 11: these look like inadvertent whitespace changes, should not be part of this PR

tests/e2e/test_api.py
```diff
@@ -98,11 +123,12 @@ def check_content_type(response, content_type, message=""):
     assert response.headers["content-type"].startswith(content_type), message
 
 
+@retry(max_attempts=3, wait_between_runs=10)
 def test_readiness():
     """Test handler for /readiness REST API endpoint."""
     endpoint = "/readiness"
     with metrics_utils.RestAPICallCounterChecker(metrics_client, endpoint):
-        response = client.get(endpoint, timeout=BASIC_ENDPOINTS_TIMEOUT)
+        response = client.get(endpoint, timeout=LLM_REST_API_TIMEOUT)
```
bparees, line 131: looks like you need to rebase your branch to get rid of some of these spurious changes.

tests/e2e/utils/wait_for_ols.py
bparees: I still think this can be removed and should be

tests/e2e/utils/wait_for_ols.py
bparees: it means it can skip verifying the certificate.  i don't think it's an issue for testing purposes, the OLS cert is signed by the cluster CA, so if we wanted to verify it we'd need to fetch the CA from the cluster and feed it into the test process.

tests/scripts/utils.sh
```diff
@@ -100,15 +85,10 @@ function run_suite() {
...
-    echo "Timed out waiting for OLS to become available"
-    echo "${JUNIT_SETUP_TEMPLATE}" | sed "s/SUITE_ID/${1}/g" > $ARTIFACT_DIR/junit_setup_${1}.xml
-
-    ARTIFACT_DIR=$ARTIFACT_DIR SUITE_ID=$1 python tests/scripts/must_gather.py
-    return 1
-  fi
+  # Determine the hostname for the ols route
+  #export OLS_URL=https://$(oc get route ols -o jsonpath='{.spec.host}')
+  #echo "OLS_URL set to $OLS_URL"
+  export SUITE_ID=$1
```
bparees, line 91: what's the purpose of this?  SUITE_ID is set and provided to the make invocation below.

bparees (commented):
please rebase this on top of the latest `main` branch to get rid of some of the weird deltas that are showing up, and remove any other changes this PR introduces that are not specifically about migrating the wait_for_ols logic.

Nehanth:
Close

PR #1058, opened 2024-06-10
OLS-559: Handle smaller chunk size

asamal4:
## Description

- Include top 5 chunks as chunk size has been reduced by https://github.com/openshift/lightspeed-rag-content/pull/41.
- De-dup referenced docs details for api response.
- Update predefined answer for existing questions as per latest changes.
- Include new QnAs.


bparees:
> @syedriko @bparees PTAL.
> Please tag [6a238245ec7d](https://quay.io/repository/openshift-lightspeed/lightspeed-rag-content/manifest/sha256:6a238245ec7da232bf64af0be95b8f5a3eede8127f4f49a123d5271e35667fc9)

tagged as `6a238245ec7d`

bparees:
@asamal4 one thing i think this is missing:  When we send the list of referenced content back in the response, we should dedupe the list (if we include multiple chunks from the same document, we should only list that document/title a single time in the list of references).  Otherwise the UI is going to display the same doc link multiple times.

make sense?

(that is what my /hold is for)

bparees:
yeah i don't think we want to include the chunk in our api response until/unless we have a product reason to do so (i.e. a way it is needed to help the user/implement a use case).  Just putting it there for CI purposes doesn't feel right, we can solve that need in other ways.

Also de-duping in the backend prevents every client from having to reimplement it.

Lastly, if we did want to include the actual chunk in the response, we could structure the api such that we list all the chunks from a given doc, rather than having the same doc entry repeated once for each chunk.

bparees:
(I was also originally thinking we'd do this in the UI and i may have even asked you to create https://issues.redhat.com/browse/OLS-713 but looking at your PR w/ fresh eyes it seemed to me it made more sense to just do it in the backend and leave the client/UI out of it)

xrajesh:

PR #1067, opened 2024-06-11
OLS-694: Improve memory usage in collector script by lazily loading llama_index

onmete:
## Description

With this, the memory consumption for the sidecar falls from 500MiB under 100MiB.

## Type of change

- [x] Optimization

### DIFF
--- a/ols/src/rag_index/index_loader.py
+++ b/ols/src/rag_index/index_loader.py
@@ -1,36 +1,44 @@
+# type: ignore
 """Module for loading index."""
 
 import logging
-from typing import TYPE_CHECKING, Any, Optional
-
-from llama_index.core import Settings, StorageContext, load_index_from_storage
-from llama_index.core.embeddings.utils import EmbedType
-from llama_index.core.indices.base import BaseIndex
-from llama_index.core.llms.utils import resolve_llm
-from llama_index.vector_stores.faiss import FaissVectorStore
+from typing import Optional
 
 from ols.app.models.config import ReferenceContent
-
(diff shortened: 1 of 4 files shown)


bparees:
/approve

/assign @tisnik 

@tisnik probably best for you to assess this from a python practices perspective, but the direction makes sense to me.

Makefile
```diff
@@ -65,7 +65,7 @@ coverage-report:	test-unit ## Export unit test coverage report into interactive
 	coverage html --data-file="${ARTIFACT_DIR}/.coverage.unit"
 
 check-types: ## Checks type hints in sources
-	mypy --explicit-package-bases --disallow-untyped-calls --disallow-untyped-defs --disallow-incomplete-defs ols/
+	mypy --explicit-package-bases --disallow-untyped-calls --disallow-untyped-defs --disallow-incomplete-defs --config-file pyproject.toml ols/
```
tisnik, line 68: this should not be needed, really

ols/src/rag_index/index_loader.py
```diff
@@ -88,7 +95,7 @@ def _load_index(self) -> None:
                 logger.exception(f"Error loading vector index:\n{err}")
 
     @property
-    def vector_index(self) -> Optional[BaseIndex]:
+    def vector_index(self):
```
tisnik, line 98: not sure if it's good approach tbh, but ok (it will kick us later)

tisnik (commented):
the idea is correct, just having some doubts (as usual)

PR #1075, opened 2024-06-12
Update example olsconfig.yaml to better suit for local OLS run

onmete:
## Description

Update example olsconfig.yaml to better suit for local OLS run.

## Type of change

- [x] Refactor


onmete:
/assign @bparees 

I've also added more defaults under dev part of the config, let me know if I should keep it there or not.

tisnik (commented):
yes that would be better

PR #1078, opened 2024-06-12
OLS-642: Redacted content should never be displayed in logs

Nehanth:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -421,21 +421,20 @@ def test_rag_question() -> None:
 
 
 def test_query_filter() -> None:
-    """Ensure responses does not include filtered words."""
+    """Ensure responses does not include filtered words and redacted words are not logged."""
     endpoint = "/v1/query"
     with metrics_utils.RestAPICallCounterChecker(metrics_client, endpoint):
+        query = "what is foo in bar?"
         response = client.post(
             endpoint,
-            json={"query": "what is foo in bar?"},
+            json={"query": query},
             timeout=LLM_REST_API_TIMEOUT,
         )
         assert response.status_code == requests.codes.ok
         check_content_type(response, "application/json")
-
         print(vars(response))
         json_response = response.json()
         assert "conversation_id" in json_response
-
         # values to be filtered and replaced are defined in:
         # tests/config/singleprovider.e2e.template.config.yaml
         response_text = json_response["response"].lower()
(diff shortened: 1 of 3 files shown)


tests/e2e/test_api.py
```diff
@@ -420,29 +421,33 @@ def test_rag_question() -> None:
         assert json_response["referenced_documents"][0]["title"]
 
 
-def test_query_filter() -> None:
-    """Ensure responses does not include filtered words."""
+def test_query_filter(caplog) -> None:
```
bparees, line 424: this is not the correct way to check the logs in an e2e test.  This might work in an integration test, but for an e2e test the test code runs completely external to the OLS process and cannot directly capture the logs.  See https://github.com/openshift/lightspeed-service/blob/aaff42e2700020db057eb1e00712e69401aa4fdb/tests/e2e/test_api.py#L1336 for an example of how to check the OLS logs in an e2e test.

Nehanth: made the changes. thanks ben

tisnik (commented):
it looks ok

tests/e2e/test_api.py
```diff
@@ -444,6 +443,17 @@ def test_query_filter() -> None:
...
+        data_collection_container_name = "ols-sidecar-user-data-collector"
+        pod_name = cluster_utils.get_single_existing_pod_name()
+
+        container_log = cluster_utils.get_container_log(
+            pod_name, data_collection_container_name
+        )
+
+        # Check if redacted words are not logged in the container log
+        assert "foo" not in container_log
+        assert "bar" not in container_log
```
bparees, line 455: please do a "positive" test as well, meaning you should make sure the expected(substituted) content (in this case text like `deploy` and `openshift`) is in the log.  Otherwise this test can pass just based on us (for some reason) not getting any log data back.

tests/e2e/test_api.py
```diff
@@ -444,6 +443,19 @@ def test_query_filter() -> None:
         assert "foo" not in response_text
         assert "bar" not in response_text
 
+        data_collection_container_name = "ols-sidecar-user-data-collector"
```
Nehanth, line 446: Container is empty

tests/e2e/test_api.py
bparees: looks like you're going to need a more exact match here because the test is currently failing by matching other log entries that contain "foo":

```
{assert 'foo' not in '2024-06-18 ...ed":false}\n'
  
  'foo' is contained here:
    ter(name='foo_filter', pattern='\\b(?:foo)\\b', replace_with='deployment'), QueryFilter(name='bar_filter', pattern='\\b(?:bar)\\b', replace_with='openshift')]    
```

PR #1081, opened 2024-06-13
OLS-671: /readiness - LLM check

onmete:
## Description

Add LLM check for /readiness probe.

## Type of change

- [x] New feature

PR #1086, opened 2024-06-15
fix readiness e2e test case

asamal4:
## Description

As we have added LLM call as part of readiness check (https://github.com/openshift/lightspeed-service/pull/1081). It will take more time to finish processing..
watsonx e2e readiness test case is failing as readiness call timeout value is very less. (failing for https://github.com/openshift/lightspeed-service/pull/1082 and other PR as well).

We could increase the timeout or do multiple retry. This PR contains both (we can remove one later, if we want)


tisnik (commented):
yes, now the BASIC_ENDPOINTS_TIMEOUT is not enough there

PR #1114, opened 2024-06-20
OLS-490: Remove deprecated Extra and replace it with literals

onmete:
## Description

Remove deprecated Extra and replace it with literals.

In pydantic:
```python
@deprecated(
    "Extra is deprecated. Use literal values instead (e.g. `extra='allow'`)", category=PydanticDeprecatedSince20
)
```

## Type of change

- [x] Refactor


tisnik (commented):
it looks ok

PR #1119, opened 2024-06-21
OLS-679: Azure token refresh

onmete:
## Description

Add cache for azure AD token depending on expiration

## Type of change

- [x] New feature

### DIFF
--- a/ols/src/llms/providers/azure_openai.py
+++ b/ols/src/llms/providers/azure_openai.py
@@ -1,6 +1,7 @@
 """Azure OpenAI provider implementation."""
 
 import logging
+import time
 from typing import Any, Optional
 
 from azure.core.credentials import AccessToken
@@ -14,6 +15,25 @@ from ols.src.llms.providers.provider import LLMProvider
 from ols.src.llms.providers.registry import register_llm_provider_as
 
 logger = logging.getLogger(__name__)
+
+
+class TokenCache:
+    """Store access token and its expiration time.
+
+    Attributes:
+        access_token: access token to be stored
+        expires_on: unix time when the token expires
+    """
+
+    access_token: Optional[str]
+    expires_on: int = 0
+
+
+def token_is_expired() -> bool:
+    """Check if the cached token is still valid."""
+    if TokenCache.expires_on == 0:
+        return True
+    return time.time() > TokenCache.expires_on
 
 
 @register_llm_provider_as(constants.PROVIDER_AZURE_OPENAI)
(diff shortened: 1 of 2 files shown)


ols/src/llms/providers/azure_openai.py
```diff
@@ -16,6 +17,19 @@
 logger = logging.getLogger(__name__)
 
 
+token_cache: dict[str, Any] = {
```
tisnik, line 20: don't you wanna to use class instead? (just question, it would be ok to use dict there)

ols/src/llms/providers/azure_openai.py
```diff
@@ -54,24 +68,37 @@ def default_params(self) -> dict[str, Any]:
...
         return AzureChatOpenAI(**self.params)
 
+    def resolve_access_token(self, azure_config: AzureOpenAIConfig) -> Optional[str]:
+        """Resolve access token to call Azure OpenAI."""
+        if token_is_expired():
+            access_token = self.retrieve_access_token(azure_config)
+            # set up active directory token to access Azure services, including OpenAI one
+            if access_token is None:
+                # we are unable to retrieve access token
+                token_cache["access_token"] = access_token  # None
```
tisnik, line 93: you can basically set up this one before line 91 (with `if`), becasuse this set operation does not depend whethere token is `None` or not

ols/src/llms/providers/azure_openai.py
```diff
@@ -54,24 +68,37 @@ def default_params(self) -> dict[str, Any]:
...
+            else:
+                token_cache["access_token"] = access_token.token
+                # Update expires_on to current time + expires_in,
+                # minus a small leeway. Default expiration seems to
+                # be 1 hour.
+                token_cache["expires_on"] = time.time() + access_token.expires_on - 30
```
tisnik, line 99: will it work in non-UTC environment?

tisnik: is the `expires_on` some relative time (duration)?

tisnik (commented):
two questions, one nitpick, but look ok in overall

ols/src/llms/providers/azure_openai.py
onmete: might be better from typing perspective indeed

ols/src/llms/providers/azure_openai.py
onmete: It is an int - `The token's expiration time in Unix time.`
I should probably explicitly make timestamps UTC.

tisnik:

PR #1124, opened 2024-06-21
OLS-688: install ols using operator for e2e

bparees:
## Description

attempting to refactor the e2e ols installer logic to install using the latest OLS
operator instead of manually creating resources

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/tests/config/operator_install/catalog.yaml
+++ b/tests/config/operator_install/catalog.yaml
@@ -1,0 +1,12 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: lightspeed-operator-catalog
+  namespace: openshift-marketplace
+spec:
+  displayName: LightSpeed Operator
+  sourceType: grpc
+  image: quay.io/openshift-lightspeed/lightspeed-catalog:latest
+  updateStrategy:
+    registryPoll:
+      interval: 10m
--- a/tests/config/operator_install/olsconfig.crd.azure_openai.yaml
+++ b/tests/config/operator_install/olsconfig.crd.azure_openai.yaml
@@ -1,0 +1,35 @@
+apiVersion: ols.openshift.io/v1alpha1
+kind: OLSConfig
+metadata:
+  name: cluster
+  labels:
+    app.kubernetes.io/created-by: lightspeed-operator
+    app.kubernetes.io/instance: olsconfig-sample
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: olsconfig
+    app.kubernetes.io/part-of: lightspeed-operator
+spec:
+  llm:
+    providers:
+      - credentialsSecretRef:
+          name: llmcreds
+        deploymentName: gpt-35-turbo-16k
+        models:
+          - name: gpt-3.5-turbo
+        name: azure_openai
+        type: azure_openai
+        url: 'https://ols-test.openai.azure.com/'
+  ols:
+    defaultModel: gpt-3.5-turbo
+    defaultProvider: azure_openai
+    deployment:
+      replicas: 1
+    disableAuth: false
+    logLevel: DEBUG
+    queryFilters:
+      - name: foo_filter
+        pattern: '\b(?:foo)\b'
+        replaceWith: "deployment"
+      - name: bar_filter
+        pattern: '\b(?:bar)\b'
+        replaceWith: "openshift"
--- a/tests/config/operator_install/olsconfig.crd.bam.yaml
+++ b/tests/config/operator_install/olsconfig.crd.bam.yaml
@@ -1,0 +1,33 @@
+apiVersion: ols.openshift.io/v1alpha1
+kind: OLSConfig
+metadata:
+  name: cluster
+  labels:
+    app.kubernetes.io/created-by: lightspeed-operator
+    app.kubernetes.io/instance: olsconfig-sample
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: olsconfig
+    app.kubernetes.io/part-of: lightspeed-operator
+spec:
+  llm:
+    providers:
+      - credentialsSecretRef:
+          name: llmcreds
+        models:
+          - name: ibm/granite-13b-chat-v2
+        name: bam
+        type: bam
+  ols:
+    defaultModel: ibm/granite-13b-chat-v2
+    defaultProvider: bam
+    deployment:
+      replicas: 1
+    disableAuth: false
+    logLevel: DEBUG
+    queryFilters:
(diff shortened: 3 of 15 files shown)


bparees:
@asamal4 ptal at the `watsonx.test_api: test_model_evaluation` test failure in this job run:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/1124/pull-ci-openshift-lightspeed-service-main-e2e-ols-cluster/1808989305238982656

based on the overall logs:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/1124/pull-ci-openshift-lightspeed-service-main-e2e-ols-cluster/1808989305238982656/build-log.txt

It looks like the test is evaluating all the model/provider pairs, even though the config is specifically watsonx/granite:

evidence:

1) in the test failure output "bam" appears even though the provider is watsonx:
```
question = 'what is kubernetes?', provider = 'bam'
```
2) the eval args seen in the overall log show the provider+model set to watsonx/granite as expected:
```
Response evaluation arguments: Namespace(keyword='', markexpr='', maxfail=0, continue_on_collection_errors=False, confcutdir=None, noconftest=False, keepduplicates=False, collect_in_virtualenv=False, importmode='prepend', basetemp=None, durations=0, durations_min=0.005, showfixtures=False, show_fixtures_per_test=False, version=0, plugins=[], traceconfig=False, verbose=0, no_header=False, no_summary=False, reportchars='fE', disable_warnings=False, showlocals=False, tbstyle='auto', showcapture='all', fulltrace=False, color='auto', code_highlight='yes', capture='no', runxfail=False, pastebin=None, assertmode='rewrite', xmlpath='/logs/artifacts/junit_e2e_watsonx.xml', junitprefix='watsonx', doctestmodules=False, doctestreport='udiff', doctestglob=[], doctest_ignore_import_errors=False, doctest_continue_on_failure=False, last_failed_no_failures='all', stepwise=False, stepwise_skip=False, logger_disable=[], asyncio_mode=None, cov_source=[], cov_report={}, cov_config='.coveragerc', no_cov_on_fail=False, no_cov=False, cov_append=False, cov_branch=None, eval_provider='watsonx', eval_model='ibm/granite-13b-chat-v2', eval_out_dir='/logs/artifacts', eval_query_ids=None, eval_scenario='with_rag'
```
3) that same log shows this output right after the eval args:  `Model evaluation for bam+ibm/granite-13b-chat-v2`

in fact that is the only output regarding model evaluation that i see in the entire build log (maybe because it on'y failed when running against watsonx, so we only captured the output for that run?  i'm not sure, pytest does some weird things to stdout unfortuantely).

it looks like maybe the messages themselves are misleading about what is actually being done (it is just gathering the test questions/expected responses for bam and the other models, vs actually attempting evaluation for bam), so maybe all that's needed is some cleanup/clearer messages, but it's definitely confusing right now.

bparees:
> Only scenario, where we will attempt to call api with different provider/model for model_evaluation is when one of the earlier test suite failed for response_evaluation and didn't create the CSV file.. And we will attempt to call API again, but it will fail as OLS is not installed with all provider/model.

@asamal4 thanks for the explanation.  debating what, if anything, we could do to improve the debuggability of this behavior.  It seems like if we know the provider/model is not configured/enabled, we could throw a cleaner error that basically explains what you just explained to me.  ("earlier invocations of the provider/model failed, so we don't have response data to compare to, current suite invocation does not include the provider/model configuration so we cannot retry the invocation")

but in any case i understand the observed behavior now, thank you

bparees:
@xrajesh @onmete in the latest run the only failure was for transcript storage checking, specifically the test expected to find rag chunks in the transcript and did not:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/1124/pull-ci-openshift-lightspeed-service-main-e2e-ols-cluster/1809262114733297664

```
        assert transcript["metadata"]  # just check if it is not empty
        assert transcript["redacted_query"] == "what is kubernetes?"
        # we don't want llm response influence this test
        assert "query_is_valid" in transcript
        assert "llm_response" in transcript
        assert "rag_chunks" in transcript
        assert isinstance(transcript["rag_chunks"], list)
>       assert len(transcript["rag_chunks"])
E       assert 0
E        +  where 0 = len([])

tests/e2e/test_api.py:1179: AssertionError}
```

i don't think that has to do w/ my changes here, but might indicate some flakiness around whether or not we consistently determine that there are sufficiently high enough scoring chunks to include them, with the query that's being used in this test case?

Either that or something went wrong actually storing the RAG metadata w/ the transcript.

/retest

bparees:
@xrajesh @tisnik ok i think this is ready.  Assuming you agree w/ the general concept, you can merge this as is or you're welcome to cherry-pick it into another PR if you want to make significant changes prior to merging.

Note that this currently uses the catalog/operator image from quay.io which means it's only as current as the last mirror time.  An optimization would be to use the CI catalog+operator image (built from the last merge to the operator repo), which would be a more current image than waiting for the mirror.  But as mirroring happens hourly, it should not be a huge issue unless you are trying to coordinate changes to the operator+operand.

The primary benefit to this change is that the e2e tests OLS as it would be installed/configured by the operator (modulo the modifications that the install/setup logic does such as changing the data collector sidecar container for stage operation), so we know that we're using all the same RBAC rules, operand deployment definition, etc.

it also ensures that the config yaml OLS is consuming is one that was generated by the operator, based on the config CRD.

bparees:
@tisnik @xrajesh bump (if you guys are busy w/ release stuff/other priorities no worries, please just give me an ack that you've seen this and will get to it later)

tisnik (commented):
it looks ok! Thank you @bparees

xrajesh:
Thank you @bparees . Sorry - was really focused on the release out . We did it ! Thank you @bparees 
This PR -  other than quay.io mirror may be using previous image - The catalog image itself needs to be built after the bundle is built in a PROW run. Or you may be using the catalog which points to the previous version of bundle. 
Keep this concern away - I think we will merge this PR and update it - by installing the bundle rather than the catalog.
Ben FYI - Even in Konflux release , we have 2 pipelines (1) Releases , operand and bundle container image (2) adds the bundle to public RH operator index (its a container ) . If you really look at it ,as long as we are consistently testing using the bundle - that is good , thats what is shipped to customers.

xrajesh:

xrajesh:

PR #1131, opened 2024-06-24
Remove tests for gradio ui

onmete:
## Description

For consideration.
Remove tests for gradio UI as it is not "real" UI, this is just for development.
There is no point in having tests for the dev component, we don't really care what happens there.

## Type of change

- [x] Refactor


bparees:
> Remove tests for gradio UI as it is not "real" UI, this is just for development.
> There is no point in having tests for the dev component, we don't really care what happens there.

my 2 cents:  if the tests are flaky or a maintenance burden, i'd support getting rid of them, but if they are working and not causing us extra work to maintain, then it seems worth keeping them as long as we still expect developers to use the gradio UI for devel/debugging purposes.  We wouldn't want it to become broken if we expect developers to be using it as part of their test/debug/devel workflow. (Which, as long as we reference it in our docs, is the case)

onmete:
I don't have enough faith in these tests to actually find out if there is something wrong with GradioUI, as we are patching the actual post there.

I would ask differently. If there would be no tests for GradioUI, would we dedicate some time to properly test it?

bparees:
> I don't have enough faith in these tests to actually find out if there is something wrong with GradioUI, as we are patching the actual post there.
> I would ask differently. If there would be no tests for GradioUI, would we dedicate some time to properly test it?

I'm going to defer this to @tisnik as he created these tests in the first place.  But I agree they don't appear to be adding a ton of value in their current form and that we probably would not prioritize making them better.  This is, imho, one of the consequences of focusing on test coverage as a metric.

That said, if we remove the existing ones, there is even less likelihood we'd ever consider improving them/writing better ones.

tisnik:
How are you planning to test all the sources in `ols/src/ui/gradio_ui.py`? Are you going to say "it is not our code" or something like it?

onmete:
My thought process is that this is purely a dev thing. If there is a code change that breaks these tests, it is not worth the time to update these tests. But the more likely scenario is that these tests won't catch the breaking change (like changing schema) as the post is mocked.
I would be fine with this: if gradio is broken, we fix it when someone actually finds out it is broken.

PR #1132, opened 2024-06-24
Fix user warning in model config

onmete:
## Description

Without this, you can see warning at the app start
```
/home/ometelka/projects/lightspeed-service/.venv/lib/python3.11/site-packages/pydantic/_internal/_fields.py:160: UserWarning: Field "model_parameters" has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.
  warnings.warn(
```

## Type of change

- [x] Bug fix

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -109,7 +109,7 @@ class ModelConfig(BaseModel):
     credentials: Optional[str] = None
 
     context_window_size: PositiveInt = constants.DEFAULT_CONTEXT_WINDOW_SIZE
-    model_parameters: ModelParameters = ModelParameters()
+    parameters: ModelParameters = ModelParameters()
 
     options: Optional[dict[str, Any]] = None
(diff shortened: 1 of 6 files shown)


ols/app/models/config.py
```diff
@@ -113,6 +114,11 @@ class ModelConfig(BaseModel):
 
     options: Optional[dict[str, Any]] = None
 
+    # model_* is protected namespace by default, as we want to use
+    # `model_parameters`, we need to remove it from protected namespaces
```
bparees, line 118: given the class name is ModelConfig, could we just rename the field to `parameters` instead?  

i don't think it matters a whole lot, but erasing all these warnings seems like overkill and may cause confusion in the future if we shadow/override something that matters?

onmete: Updated.

PR #1136, opened 2024-06-24
OLS-734: Add attachments to transcript

onmete:
## Description

Add attachments to transcript.

## Type of change

- [x] New feature

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -103,7 +103,9 @@ def conversation_request(
     # Redact all attachments
     attachments = redact_attachments(conversation_id, attachments)
 
-    # All attachments should be appended to query
+    # All attachments should be appended to query - but store original
+    # query for later use in transcript storage
+    query_without_attachments = llm_request.query
     llm_request.query = append_attachments_to_query(llm_request.query, attachments)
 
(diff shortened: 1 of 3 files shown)


ols/app/endpoints/ols.py
```diff
@@ -103,7 +103,9 @@ def conversation_request(
     # Redact all attachments
     attachments = redact_attachments(conversation_id, attachments)
 
-    # All attachments should be appended to query
+    # All attachments should be appended to query - but store original
+    # query for later use in transcript storage
+    query_without_attachments = llm_request.query
```
onmete, line 108: I guess this is TBD, but I don't really know if we should store the query with attachments AND attachments.

It might be good to have the actual full query to replicate the answers, but it should be possible to construct it anyway (from the `attachments` field in the transcript).

So this is a question of data duplication vs. some use case (?)

tisnik: IMHO we should start first to store + retrieve attachments, not vice versa. IMHO this ^ is not fully solvable

tisnik (commented):
it looks ok

tisnik:

tisnik:

PR #1137, opened 2024-06-24
Usefixtures in integration tests who activelly does not use the fixture

tisnik:
## Description

Usefixtures in integration tests

Fixtures can be injected either by passing them as parameters to the test
function, or by using the `@pytest.mark.usefixtures` decorator.

If the test function depends on the fixture being activated, but does not
use it in the test body or otherwise rely on its return value, prefer
the `@pytest.mark.usefixtures` decorator, to make the dependency explicit
and avoid the confusion caused by unused arguments.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/tests/integration/test_authorized.py
+++ b/tests/integration/test_authorized.py
@@ -39,7 +39,8 @@ def _enabled_auth():
     config.dev_config.disable_auth = False
 
 
-def test_post_authorized_disabled(_disabled_auth):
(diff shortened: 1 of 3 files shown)


tisnik:

PR #1147, opened 2024-06-25
OLS-431: evaluation using groundtruth

asamal4:
## Description

- Capability to evaluate different provider+models using same pre-defined answer (currently similarity score)Save result with all model responses.
- Save summary score which can be used to compare different models (currently average of similarity score is used)
- Enable run from CI (currently not used to gate PRs)
- Combined with existing consistency check, just to avoid duplicate llm calls. If response is not available then only call api to get response.

Note: groundtruth answer is not relevant for this PR. Currently BAM+Granite result is used as ground-truth to test the approach.

[Sample summary result - json](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/1147/pull-ci-openshift-lightspeed-service-main-e2e-ols-cluster/1805586570246361088/artifacts/e2e-ols-cluster/e2e/artifacts/model_evaluation_summary.json)

[sample full result - csv](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/1147/pull-ci-openshift-lightspeed-service-main-e2e-ols-cluster/1805586570246361088/artifacts/e2e-ols-cluster/e2e/artifacts/model_evaluation_summary.json)

### DIFF
--- a/tests/e2e/utils/response_evaluation.py
+++ b/tests/e2e/utils/response_evaluation.py
@@ -3,10 +3,11 @@
 import json
 import os
 from collections import defaultdict
+from datetime import UTC, datetime
 
 import requests
 from llama_index.embeddings.huggingface import HuggingFaceEmbedding
-from pandas import DataFrame
+from pandas import DataFrame, read_csv
 from scipy.spatial.distance import cosine, euclidean
 
 from tests.e2e.utils.constants import EVAL_THRESHOLD, LLM_REST_API_TIMEOUT
@@ -28,6 +29,38 @@ class ResponseEvaluation:
 
         with open("tests/test_data/question_answer_pair.json") as qna_f:
             self._qa_pairs = json.load(qna_f)["evaluation"]
+
+    def _get_api_response(self, question, provider, model):
+        """Get api response for a question/query."""
+        response = self._api_client.post(
+            "/v1/query",
+            json={
+                "query": question,
+                "provider": provider,
+                "model": model,
+            },
+            timeout=LLM_REST_API_TIMEOUT,
+        )
+        if response.status_code != requests.codes.ok:
+            raise Exception(response)
+
+        print(f"API request is successful for {provider}+{model}; Query: {question}")
+        return response.json()["response"].strip()
+
+    def _get_recent_response(self, recent_resp_df, question, provider, model):
+        """Get llm response from the stored data, if available."""
+        if recent_resp_df is not None:
+            try:
+                return recent_resp_df[recent_resp_df.question == question][
+                    "llm_response"
+                ].iloc[0]
+            except IndexError:
+                print(
+                    "Recent response for query is not found in the file. "
+                    "Separate api call is required to get response."
+                )
+        # Recent response is not found, call api to get response
+        return self._get_api_response(question, provider, model)
 
     def _similarity_score(self, response, answer):
         """Calculate similarity score between two strings."""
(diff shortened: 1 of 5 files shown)


tests/e2e/utils/response_evaluation.py
```diff
@@ -29,6 +30,23 @@ def __init__(self, eval_args, api_client):
...
+        response = self._api_client.post(
+            "/v1/query",
+            json={
+                "query": question,
+                "provider": provider,
+                "model": model,
+            },
+            timeout=LLM_REST_API_TIMEOUT,
+        )
+        if response.status_code != requests.codes.ok:
```
tisnik, line 44: `assert response.status_code == requests.codes.ok`

tests/e2e/utils/response_evaluation.py
```diff
@@ -143,3 +151,68 @@ def validate_response(self):
             print("No result. Nothing to process.")
 
         return True
+
+    def evaluate_models(self):
+        """Evaluate models against groundtruth answer."""
+        print("Running model evaluation using groundtruth...")
+        inscope_models = [
```
tisnik, line 158: if you use list of lists, no splitting will be needed later. that `+` might be problematic in future when some provider/model choose to use that symbol somewhere (remember c++ :)

tests/e2e/utils/response_evaluation.py
```diff
@@ -143,3 +151,68 @@ def validate_response(self):
...
+        for provider_model in inscope_models:
+            result_dict = defaultdict(list)
+            print(f"Model evaluation for {provider_model}")
+            provider, model = provider_model.split("+")
+            try:
+                prev_resp = read_csv(
+                    f"{self._args.eval_out_dir}/response_evaluation_result-"
+                    f"{provider_model.replace('/', '-')}+with_rag.csv"
+                )
+            except FileNotFoundError:
```
tisnik, line 176: no logging/anything?

tests/e2e/utils/response_evaluation.py
```diff
@@ -143,3 +151,68 @@ def validate_response(self):
...
+                prev_resp = None
+
+            for query_id in self._qa_pairs.keys():
+                question = self._qa_pairs[query_id]["question"]
+                answer_data = self._qa_pairs[query_id]["answer"][answer_id]
+                answer = answer_data["text"][0]
+
+                if prev_resp is not None:
```
tisnik, line 184: might be worth to comment what this logic does exactly (at least for me it is not clear). Or better - that might be ideal block to be refactored into it's own function.

tests/e2e/utils/response_evaluation.py
```diff
@@ -143,3 +151,68 @@ def validate_response(self):
...
+        summary_dict = {}
+        for m in inscope_models:
+            summary_dict[m] = result_df[f"{m}_score"].mean()
+        summary_dict = {
+            "timestamp": str(datetime.now(UTC)),
+            "eval_set": result_df.eval_id.unique().tolist(),
+            "avg_similarity_score": summary_dict,
+        }
+        with open(f"{self._args.eval_out_dir}/model_evaluation_summary.json", "w") as f:
+            json.dump(summary_dict, f)
```
tisnik, line 218: prob. I'm lost, but does it mean that every test that call this function, will spit some JSONs?

tests/e2e/utils/response_evaluation.py
tisnik: you mean that something from `tests/e2e` will be used from other code?

tests/e2e/utils/response_evaluation.py
```diff
@@ -143,3 +166,68 @@ def validate_response(self):
...
+
+        result_df.to_csv(
+            f"{self._args.eval_out_dir}/model_evaluation_result.csv", index=False
+        )
+        summary_dict = {}
+        for p_m in inscope_models:
+            provider_model = "+".join(p_m)
+            summary_dict[provider_model] = result_df[f"{provider_model}_score"].mean()
+        summary_dict = {
+            "timestamp": str(datetime.now(UTC)),
```
tisnik, line 228: `datetime.utcnow()`

tests/e2e/utils/response_evaluation.py
tisnik: you are right, because it is possible to turn off assertions, it is not a good idea to have them in real code (just in tests). If that's ^ your reason I'm ok with the code ;)

tests/e2e/utils/response_evaluation.py
tisnik: in 3.12, indeed

tisnik:

PR #1149, opened 2024-06-25
Fix JSONDecodeError in collector

onmete:
## Description

Log JSONDecodeError in collector

## Type of change

- [x] Fix

### DIFF
--- a/ols/user_data_collection/data_collector.py
+++ b/ols/user_data_collection/data_collector.py
@@ -267,8 +267,9 @@ def upload_data_to_ingress(tarball: io.BytesIO) -> requests.Response:
         )
 
     if response.status_code != requests.codes.accepted:
+        logger.error(f"posting payload failed, full response: {response}")
         raise requests.exceptions.HTTPError(
-            f"data upload failed with response: {response.json()}"
+            f"data upload failed with response code: {response.status_code}"
         )
 
     request_id = response.json()["request_id"]


ols/user_data_collection/data_collector.py
```diff
@@ -257,14 +257,18 @@ def upload_data_to_ingress(tarball: io.BytesIO) -> requests.Response:
...
+        with requests.Session() as s:
+            s.headers = headers
+            logger.debug(f"posting payload to {url}")
+            response = s.post(
+                url=url,
+                files=payload,
+                timeout=INGRESS_TIMEOUT,
+            )
+    except json.JSONDecodeError:
+        logger.error(f"can't decode JSON content, response: {response}")
```
bparees, line 270: please put a little more context in this error message about what OLS was trying to do/what failed.  Keep in mind this might be the only message a user sees in the log and they will have no idea what the implication is when they see it.

bparees:
> some test(s) perhaps?

maybe i'm not imaginative enough but i can't think of  a useful test here.....   the only test i can think of would amount to mocking the post call to force a json decoder error, at which point all you're testing is that "yes, the try/catch block works to catch the specific exception we mocked the code to throw"

I suppose it would ensure no one later removed the try/catch logic, but does it add any other value?  or is there some other test you have in mind that would accomplish more?

tisnik:
> > some test(s) perhaps?
> 
> maybe i'm not imaginative enough but i can't think of a useful test here..... the only test i can think of would amount to mocking the post call to force a json decoder error, at which point all you're testing is that "yes, the try/catch block works to catch the specific exception we mocked the code to throw"
> 
> I suppose it would ensure no one later removed the try/catch logic, but does it add any other value? or is there some other test you have in mind that would accomplish more?

anything that can happen in code. If the `response` will be renamed or the logic will completely change, the f-string won't be filled in and yet another runtime exception will be raised this time. Very probably there are other cases, but having uncovered lines of code in Python (and especially in Python, that is not compiled, and the importing is different from checking runtime behaviour) is just yet another bug waiting to occur in prod.

(btw we already had problem with exceptions raised in catch block because of improper f-string)

tisnik (commented):
now it looks ok, just please try to edit the title

tisnik:

PR #1156, opened 2024-06-26
Don't test for 'any' exception, we should be more specific

tisnik:
## Description

Don't test for 'any' exception, we should be more specific


tisnik:

PR #1159, opened 2024-06-26
OLS-697: Add CacheEntry model

onmete:
## Description

Add CacheEntry model.
The history format we send in the prompt is unchanged.
The follow-up is adding the attachments to the model.
The insert_or_append method can be used to store only single cache entry - which is exactly what is happening in the ols endpoint.

Question:
- Should Redis cache be updated? Or finally, removed?
   - not now

## Type of change

- [x] New feature

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -6,7 +6,7 @@ import logging
 import os
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Literal, Optional
+from typing import Any, Optional
 
 import pytz
 from fastapi import APIRouter, Depends, HTTPException, status
@@ -15,6 +15,7 @@ from ols import config, constants
 from ols.app import metrics
 from ols.app.models.models import (
     Attachment,
+    CacheEntry,
     ErrorResponse,
     ForbiddenResponse,
     LLMRequest,
@@ -185,12 +186,10 @@ def retrieve_conversation_id(llm_request: LLMRequest) -> str:
     return conversation_id
 
 
-def retrieve_previous_input(
-    user_id: str, llm_request: LLMRequest
-) -> list[dict[Literal["type", "content"], str]]:
+def retrieve_previous_input(user_id: str, llm_request: LLMRequest) -> list[CacheEntry]:
     """Retrieve previous user input, if exists."""
     try:
-        previous_input: list[dict[Literal["type", "content"], str]] = []
+        previous_input = []
         if llm_request.conversation_id:
             cache_content = config.conversation_cache.get(
                 user_id, llm_request.conversation_id
@@ -245,7 +244,7 @@ def retrieve_attachments(llm_request: LLMRequest) -> list[Attachment]:
 def generate_response(
     conversation_id: str,
     llm_request: LLMRequest,
-    previous_input: list[dict[Literal["type", "content"], str]],
+    previous_input: list[CacheEntry],
 ) -> SummarizerResponse:
     """Generate response based on validation result, previous input, and model output."""
     # Summarize documentation
@@ -253,11 +252,7 @@ def generate_response(
         docs_summarizer = DocsSummarizer(
             provider=llm_request.provider, model=llm_request.model
         )
-        history = [
-            conversation["type"] + ": " + conversation["content"].strip()
-            for conversation in previous_input
-            if conversation
-        ]
+        history = CacheEntry.cache_entries_to_history(previous_input)
         return docs_summarizer.summarize(
             conversation_id, llm_request.query, config.rag_index, history
         )
@@ -282,16 +277,6 @@ def generate_response(
         )
 
 
-def human_msg(content: str) -> dict[Literal["type", "content"], str]:
-    """Create a human message dictionary."""
-    return {"type": "human", "content": content}
-
-
-def ai_msg(content: str) -> dict[Literal["type", "content"], str]:
-    """Create an AI message dictionary."""
-    return {"type": "ai", "content": content}
-
-
 def store_conversation_history(
     user_id: str, conversation_id: str, llm_request: LLMRequest, response: Optional[str]
 ) -> None:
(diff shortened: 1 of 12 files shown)


ols/src/cache/cache.py
```diff
@@ -39,30 +40,28 @@ def construct_key(user_id: str, conversation_id: str) -> str:
         return f"{user_id}{Cache.COMPOUND_KEY_SEPARATOR}{conversation_id}"
 
     @abstractmethod
-    def get(
-        self, user_id: str, conversation_id: str
-    ) -> Optional[list[dict[Literal["type", "content"], str]]]:
+    def get(self, user_id: str, conversation_id: str) -> Optional[list[CacheEntry]]:
```
tisnik, line 43: if we are updating the code - does it make sense to still use Optional there?

ols/app/models/models.py
```diff
@@ -490,3 +490,49 @@ class SummarizerResponse:
     response: str
     rag_chunks: list[RagChunk]
     history_truncated: bool
+
+
+class CacheEntry(BaseModel):
```
tisnik, line 495: it's ok, just the name is a bit misleading, as the real cache entry also contains compound keys etc.

ols/src/cache/postgres_cache.py
```diff
@@ -104,9 +105,7 @@ def initialize_cache(self) -> None:
         cur.close()
         self.conn.commit()
 
-    def get(
-        self, user_id: str, conversation_id: str
-    ) -> Optional[list[dict[Literal["type", "content"], str]]]:
+    def get(self, user_id: str, conversation_id: str) -> Optional[list[CacheEntry]]:
```
tisnik, line 108: it is not optional in the code, right?

tests/unit/cache/test_postgres_cache.py
```diff
@@ -166,62 +159,10 @@ def test_insert_or_append_operation_first_item(mock_connect):
     mock_cursor.execute.assert_has_calls(calls, any_order=True)
 
 
-@patch("psycopg2.connect")
```
tisnik, line 169: not needed anymore?

ols/app/endpoints/ols.py
```diff
@@ -85,8 +86,6 @@ def conversation_request(
 
     conversation_id = retrieve_conversation_id(llm_request)
 
-    # Redacts the Query then logs
```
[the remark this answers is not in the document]
onmete: The comment is confusing. I read it as do the query redaction first and then you can redact logs. What about this change? https://github.com/openshift/lightspeed-service/pull/1174

ols/src/cache/cache.py
onmete: Right, it can't return `None`, but empty list. so just `-> list[CacheEntry]`?

ols/app/models/models.py
onmete: Any better suggestion? :)

ols/src/cache/postgres_cache.py
onmete: same as here https://github.com/openshift/lightspeed-service/pull/1159#discussion_r1657157876

ols/app/endpoints/ols.py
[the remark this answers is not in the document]
tisnik: yup it might be confusing. Your note is better. Do you plan to add it via this PR?

ols/src/cache/cache.py
tisnik: yes please

tests/unit/cache/test_postgres_cache.py
onmete: I hope I didn't understood the test wrongly, but it seemed like it is a test for adding multiple entires via single `insert_or_append`, which is not possible anymore - only single entry can be added via `insert_or_append`.

ols/src/cache/postgres_cache.py
tisnik: (Q&A above)

ols/app/endpoints/ols.py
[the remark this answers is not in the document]
onmete: Since that PR already exists, I'll rebase once merged.

tests/unit/cache/test_postgres_cache.py
tisnik: it checks if `insert_or_append` appends new Q&A into the thread. It must call `SELECT` followed by `INSERT`

tisnik: https://app.codecov.io/gh/openshift/lightspeed-service/pull/1159/blob/ols/src/cache/postgres_cache.py

onmete: added back

ols/app/models/models.py
onmete: I assume no better name for the moment :)

ols/app/endpoints/ols.py
```diff
@@ -245,19 +244,15 @@ def retrieve_attachments(llm_request: LLMRequest) -> list[Attachment]:
...
     try:
         docs_summarizer = DocsSummarizer(
             provider=llm_request.provider, model=llm_request.model
         )
-        history = [
-            conversation["type"] + ": " + conversation["content"].strip()
-            for conversation in previous_input
-            if conversation
-        ]
+        history = CacheEntry.cache_entries_to_history(previous_input)
```
xrajesh, line 255: @onmete Previously we were removing the whitespaces from the "content" - dont see that happening with this changes - may be I wam missing to see ?

ols/app/models/models.py
```diff
@@ -490,3 +490,49 @@ class SummarizerResponse:
...
+        )
+
+    @staticmethod
+    def cache_entries_to_history(cache_entries: list["CacheEntry"]) -> list[str]:
+        """Convert cache entries to a history."""
+        history = []
+        for entry in cache_entries:
+            history.append(f"human: {entry.query}")
+            if entry.response != "":
+                history.append(f"ai: {entry.response}")
```
xrajesh, line 536: @onmete When we did not have a response - I would think we want to let the the LLM know that question we dont have an answer - Trying to understand why we just add the question part of history and dont say anything about the answer.

ols/app/endpoints/ols.py
onmete: Right, I've added that to `cache_entries_to_history` and extra test for that.

ols/app/models/models.py
onmete: Good catch. It is probably leftover from my testing. Changed and added test.

PR #1172, opened 2024-06-27
Remove duplicated tests for health endpoints

onmete:
## Description

Remove duplicated tests for health endpoints. There are basically the same tests at the top (also with metrics).

## Type of change

- [x] Refactor


tisnik (commented):
yup, it looks ok

PR #1174, opened 2024-06-27
Improve note about query redaction

onmete:
## Description

Improve note about query redaction

## Type of change

- [x] Refactor


tisnik (commented):
it is good

PR #1177, opened 2024-06-28
[WIP] prompt - update instruction for yaml & commands

asamal4:
## Description

Currently gpt is declining to respond stating below reasons. (worked for granite)
1. Not allowed to execute command. This impacts queries where ask requires command execution which not a capability.
Ex: What pods are in the openshift-lightspeed namespace?

2. With slight variation, it may consider a query to be non-ocp. Anything related to kubernetes/deployment should be considered as ocp related. Here  `image red.ht/accelerators:2016` reason for query to be considered as non-ocp. 
Ex: show me an example YAML deployment for the image red.ht/accelerators:2016
If we have any reference in RAG chunk, then also it may work. But we don't get any reference to red.ht/accelerators:2016.
So prompt modification is required.

Response changes for existing eval queries are majorly within cut-off. But all pre-defined answers were modified to highlight how much deviation from initial answers.
Manually verified responses, where variance is beyond cut-off. Difference is not that significant.

Above 2 queries are added to eval set.

### DIFF
--- a/ols/src/prompts/prompts.py
+++ b/ols/src/prompts/prompts.py
@@ -14,7 +14,7 @@ from ols.constants import SUBJECT_ALLOWED, SUBJECT_REJECTED
 # but that is not done as granite was adding role tags like `Human:` in the response.
 # With PromptTemplate, we have more control how we want to structure the prompt.
 
-QUERY_SYSTEM_INSTRUCTION = """
+QUERY_SYSTEM_INSTRUCTION_OLD = """
 You are OpenShift Lightspeed - an intelligent assistant for question-answering tasks \
 related to the OpenShift container orchestration platform.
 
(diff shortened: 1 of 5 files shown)


ols/src/prompts/prompts.py
```diff
@@ -23,7 +23,10 @@
 Refuse to assume any other identity or to speak as if you are someone else.
 If the context of the question is not clear, consider it to be OpenShift.
 Never include URLs in your replies.
-Refuse to answer questions or execute commands not about OpenShift.
+Refuse to answer questions that are not about OpenShift.
+When query is related to Kubernetes, Deployment & cluster, then answer with reference to OpenShift.
```
syedriko, line 27: What does "Deployment & cluster" refer to?

ols/src/prompts/prompts.py
```diff
@@ -23,7 +23,10 @@
...
+You can not execute any command directly, so respond with the relevant command & instruction with placeholder name.
```
syedriko, line 28: What does "& instruction with placeholder name" mean?

ols/src/prompts/prompts.py
syedriko: I'm not sure this adds clarity. 
I was hoping to emphasize the relationship between OpenShift and Kubernetes with the
"OpenShift is a distribution of Kubernetes..."
line below. May be the "basic facts" section needs to precede the instructions...

PR #1179, opened 2024-06-28
Add some config default values and load those in data collector

onmete:
## Description

Add some config default values and load those in data collector.
Only subconfigs that are needed for data collector are initialized here. The proper approach should be to have defaults for all fields (subconfigs with its defaults instead of None's), but more changes are required in order to make it happen.

## Type of change

- [x] Refactor
- [x] New feature
- [x] Bug fix

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -847,7 +847,7 @@ class OLSConfig(BaseModel):
     conversation_cache: Optional[ConversationCacheConfig] = None
     logging_config: Optional[LoggingConfig] = None
     reference_content: Optional[ReferenceContent] = None
-    authentication_config: Optional[AuthenticationConfig] = None
+    authentication_config: AuthenticationConfig = AuthenticationConfig()
     tls_config: Optional[TLSConfig] = None
 
     default_provider: Optional[str] = None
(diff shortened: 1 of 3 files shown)


ols/app/models/config.py
```diff
@@ -847,15 +847,15 @@ class OLSConfig(BaseModel):
     conversation_cache: Optional[ConversationCacheConfig] = None
     logging_config: Optional[LoggingConfig] = None
     reference_content: Optional[ReferenceContent] = None
-    authentication_config: Optional[AuthenticationConfig] = None
+    authentication_config: Optional[AuthenticationConfig] = AuthenticationConfig()
```
tisnik, line 850: not optional anymore

ols/app/models/config.py
```diff
@@ -847,15 +847,15 @@ class OLSConfig(BaseModel):
...
     tls_config: Optional[TLSConfig] = None
 
     default_provider: Optional[str] = None
     default_model: Optional[str] = None
     query_filters: Optional[list[QueryFilter]] = None
     query_validation_method: Optional[str] = constants.QueryValidationMethod.LLM
 
-    user_data_collection: Optional[UserDataCollection] = None
+    user_data_collection: Optional[UserDataCollection] = UserDataCollection()
```
tisnik, line 858: not optional anymore

tisnik (commented):
ok

PR #1181, opened 2024-06-28
Use provider type, not provider name, as metrics label

bparees:
## Description

provider name can be whatever the admin wants.  provider type has to be one of our explicit known values.  For purposes of us understanding which providers customers are using, we want the type(azure_openai, openai, watsonx, etc), not their arbitrary name for the provider config.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -136,7 +136,7 @@ class DocsSummarizer(QueryHelper):
 
         with TokenMetricUpdater(
             llm=bare_llm,
-            provider=self.provider,
+            provider=provider_config.type,
             model=self.model,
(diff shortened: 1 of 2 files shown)


bparees:
@tisnik ptal

does not break existing tests because all our existing configs use the same value for `name` and `type`.  Which is probably something to fix in the future to make sure we are always using the right field in the right spots.  When both values are the same it masks cases where we are using the wrong value.

tisnik (commented):
yes, that's indeed better

PR #1198, opened 2024-07-08
No global skips for Bandit security checks

tisnik:
## Description

- No global skips for Bandit security checks
- Explicit skips in the code on few places only

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -191,7 +191,7 @@ ATTACHMENT_CONTENT_TYPES = frozenset(
 )
 
 # Default name of file containing API token
-API_TOKEN_FILENAME = "apitoken"  # noqa: S105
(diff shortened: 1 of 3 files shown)


onmete:
Or maybe disable the rule? It won't prevent pushing the commit with a secret.
It would just fail in CI. If we remove the rule, we move the responsibility to spot a secret to the reviewer.

tisnik:
> Or maybe disable the rule? It won't prevent pushing the commit with a secret. It would just fail in CI. If we remove the rule, we move the responsibility to spot a secret to the reviewer.

the rule is useful to catch these situation. We just have some "artifical" secrets in the code (ideally it won't be the case)

PR #1208, opened 2024-07-09
Update query in test_transcripts_storing_cluster e2e test

onmete:
## Description

Update query in the test. This test is sometimes failing due to missing rag in the transcript. The query was too simple so the rag was not always pulled.

## Type of change

- [x] Refactor
- [x] Bug fix


tisnik (commented):
it looks ok too

onmete:
As it is simple PR, I can reopen if needed. Closing for now.

PR #1219, opened 2024-07-09
Switch to azure gpt35-16k for ci e2e

asamal4:
## Description

Switch to azure gpt35-16k for ci e2e


tisnik (commented):
it seems to be ok. I'll just wait till CI finishes

PR #1220, opened 2024-07-09
OLS-697: Add attachments to conversation cache

onmete:
## Description

Add attachments to conversation cache

## Type of change

- [x] New feature


ols/app/models/models.py
```diff
@@ -502,6 +502,7 @@ class CacheEntry(BaseModel):
 
     query: str
     response: Optional[str] = ""
+    attachments: list[Attachment] = Field(default_factory=list)
```
tisnik, line 505: I don't try it, but `= []` is not sufficient?

tisnik (commented):
it looks ok. If the one nitpick can not be solved, I'll lgtm ASAP

ols/app/models/models.py
onmete: It should work too, pydantic is smart in this way. https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values

I'll change.

tisnik:
please squash commits and we are done ;)

PR #1223, opened 2024-07-10
Evaluation - Add 30 per topic QnAs

asamal4:
## Description

Add per topic QnAs (parquet file)
Move json file to eval folder
Use absolute json path


scripts/evaluation/response_evaluation.py
```diff
@@ -41,10 +41,14 @@ def __init__(self, eval_args, api_client):
         )
         self._rouge_scorer = RougeScorer(["rougeL"], use_stemmer=True)
 
-        with open("tests/test_data/question_answer_pair.json") as qna_f:
+        eval_dir = os.path.dirname(__file__)
+        self._input_dir = os.path.join(eval_dir, "eval_data")
+        with open(f"{self._input_dir}/question_answer_pair.json") as qna_f:
```
tisnik, line 46: I'd use os.join there too, a bit less fragile than f-string

tisnik (commented):
one nitpick, but look ok

tisnik (commented):
it looks ok

PR #1227, opened 2024-07-11
Add script to get rag chunks for spot verification

asamal4:
## Description

Sometime it is essential to verify retrieved RAG chunks.
- to analyze any change in response during evaluation due to RAG content
- to check what exactly we are sending to LLM.
- to check order and score for the chunks.

We can run this script on adhoc basis to do above analysis.


tisnik (commented):
yup it looks correct

PR #1264, opened 2024-07-17
Don't override credentials when they are not specified in provider specific config

onmete:
## Description

Don't override credentials when they are not specified in provider specific config.

## Type of change

- [x] Bug fix


tisnik (commented):
it looks ok

tisnik:

PR #1292, opened 2024-07-19
Adding upload_artifact_s3

JoaoFula:
Adding function to upload artifacts to s3 so they can be send to report portal. This is related to [OLS-592](https://issues.redhat.com/browse/OLS-592)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,8 @@ dev = [
     "ruff==0.5.2",
     "bandit==1.7.9",
     "types-requests==2.32.0.20240622",
-    "gradio==4.37.2"
+    "gradio==4.37.2",
+    "boto3==1.34.145",
 ]
 
 [[tool.pdm.source]]
--- a/scripts/upload_artifact_s3.py
+++ b/scripts/upload_artifact_s3.py
@@ -1,0 +1,54 @@
+"""Upload artifact containing the pytest results and configuration to an s3 bucket.
+
+This will in turn get picked by narberachka service,
+which will either send it to ibutsu or report portal, based on the
+prefix of the file name.
+A dictionary containing the credentials of the S3 bucket must be specified, containing the keys:
+AWS_BUCKET
+AWS_REGION
+AWS_ACCESS_KEY_ID
+AWS_SECRET_ACCESS_KEY
+
+"""
+
+import os
+
+import boto3
+from botocore.exceptions import ClientError
+
+path = "."
+extension = ".tar.gz"
+
+
+def upload_artifact_s3(aws_env):
+    """Upload artifact to the specified S3 bucket.
+
+    Returns:
+        True if upload successful.
+        False otherwise.
+    """
+    s3_client = boto3.client(
+        service_name="sqs",
+        region_name=aws_env["AWS_REGION"],
+        aws_access_key_id=aws_env["AWS_ACCESS_KEY_ID"],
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -47,14 +47,29 @@ dev = [
...
+    "gradio==4.37.2",
+    "boto3==1.34.145",
 ]
 
 [[tool.pdm.source]]
 type = "find_links"
 url = "https://download.pytorch.org/whl/cpu/torch_stable.html"
 name = "torch"
 
+
```
tisnik, line 59: why it was moved?

scripts/upload_artifact_s3.py
```diff
@@ -0,0 +1,38 @@
+import os
+import boto3
+import datetime
+from botocore.exceptions import ClientError
+
```
tisnik, line 5: just please specify which environment variables should be set

like in https://github.com/openshift/lightspeed-service/blob/main/scripts/data_collection/download_ols_archive.py#L3

scripts/upload_artifact_s3.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    s3_client = boto3.client(service_name="sqs",
+                             region_name=aws_env["AWS_REGION"],
+                             aws_access_key_id=aws_env["AWS_ACCESS_KEY_ID"],
+                             aws_secret_access_key=aws_env["AWS_SECRET_ACCESS_KEY"])
+
+    for file in os.listdir(path):
+        if file.endswith(extension):
+            file_name = file
```
tisnik, line 25: so it will upload just one file. is it expected?

scripts/upload_artifact_s3.py
```diff
@@ -0,0 +1,38 @@
...
+    print("found a file to upload: %s\n", file_name)
```
tisnik, line 26: f-string

tisnik (commented):
some comments, the main one about the loop + uploading just one file

scripts/upload_artifact_s3.py
```diff
@@ -0,0 +1,38 @@
+import os
```
tisnik, line 1: if you add shebang, you can make this script executable

pyproject.toml
JoaoFula: Don't know. when I run pdm self update, it moves this.
I can move it back, no problem

scripts/upload_artifact_s3.py
JoaoFula: 100%, thank you!

scripts/upload_artifact_s3.py
JoaoFula: The file is a .tar.gz so yes

pyproject.toml
tisnik: please do, let's keep the changes minimal. TY

tisnik (commented):
ok :)

tisnik:
`make verify` + `make format`

scripts/upload_artifact_s3.py
JoaoFula: To be honest I don't know if we want it executable. Since we're expected to only use this in CI, I think we can just be calling it from within our code and we avoid using subprocess which I feel is very unsafe to be running.

tisnik: that's ok then

PR #1332, opened 2024-07-30
Removed redundant open modes

tisnik:
## Description

Redundant `open` mode parameters are unnecessary and should be removed to
avoid confusion.

## Example
```python
with open("foo.txt", "r") as f:
    ...
```

Use instead:
```python
with open("foo.txt") as f:
    ...
```

## References
- [Python documentation: `open`](https://docs.python.org/3/library/functions.html#open)


tisnik:

PR #1372, opened 2024-08-07
Remove unused manifest and configmap

onmete:
## Description

Remove unused manifest and configmap.
We are using operator in e2e.

## Type of change

- [x] Refactor


tests/e2e/test_api.py
```diff
@@ -237,6 +237,15 @@ def install_ols() -> tuple[str, str, str]:
         ]
     )
 
+    # one of our libs logs a secrets in debug mode which causes the pod
+    # logs beying redacted/removed completely - we need log at info level
+    configmap_yaml = cluster_utils.run_oc(["get", "cm/olsconfig", "-o", "yaml"]).stdout
+    modified_yaml = configmap_yaml.replace(
```
tisnik, line 243: can it this be done via YAML manipulation, not on string level (it can fail when whitespaces will be different etc.)

tests/e2e/utils/cluster.py
```diff
@@ -4,21 +4,25 @@
...
             check=True,
+            input=input,
         )
         return res
     except subprocess.CalledProcessError as e:
-        print(
-            f"Error running oc command {args}: {e}, stdout: {e.output}, stderr: {e.stderr}"
-        )
-        raise
+        if "AlreadyExists" in e.stderr:
```
tisnik, line 19: IMHO it need to be checked outside that function. As this one just runs any command with any output, where "AlreadyExists" (raise) might even be expected outcode

tisnik (commented):
looks ok!

tisnik:

PR #1430, opened 2024-08-19
model specific dynamic prompt creation

asamal4:
## Description
Every model/provider has some recommendation or specific way of adding special tags, which help models to understand instruction better. Also azure uses tags/roles to do the content/jailbreak filter.

- Modified the logic to have special tags included in the prompt for granite based on this [documentation](https://dataplatform.cloud.ibm.com/docs/content/wsj/analyze-data/fm-models-ibm-chat.html?context=wx#chat).
- Modified for GPT prompt to have correct role. This will helps us to enable azure ai content filter.

This change will potentially resolve below:

- We can enable azure jailbreak filter as we will have separate role for system instruction and azure won't flag our system instruction as jailbreak. [OLS-776](https://issues.redhat.com//browse/OLS-776) 
- This will also enable us to have separate prompt structure with special tags for granite & gpt. [OLS-525](https://issues.redhat.com//browse/OLS-525).
Tested with few examples for granite, it is not adding some unwanted text/tags like [Explanation] in the response & tends to understand history better.

We will update prompt structure for any new provider/model accordingly, when we start supporting.

### DIFF
--- a/ols/src/prompts/prompt_generator.py
+++ b/ols/src/prompts/prompt_generator.py
@@ -1,80 +1,125 @@
 """Prompt generator based on model / context."""
 
-from collections import namedtuple
-from typing import Any, Optional
-
-from langchain.prompts import PromptTemplate
-
-from ols.constants import (
-    GPT35_TURBO,
-    GRANITE_13B_CHAT_V2,
-    PROVIDER_AZURE_OPENAI,
-    PROVIDER_BAM,
-    PROVIDER_OPENAI,
-    PROVIDER_RHOAI_VLLM,
-    PROVIDER_WATSONX,
+from langchain_core.messages import AIMessage, HumanMessage
+from langchain_core.prompts import (
+    ChatPromptTemplate,
+    HumanMessagePromptTemplate,
+    MessagesPlaceholder,
+    PromptTemplate,
+    SystemMessagePromptTemplate,
 )
 
+from ols.constants import ModelFamily
+
 from .prompts import (
-    CONTEXT_PLACEHOLDER,
-    HISTORY_PLACEHOLDER,
-    QUERY_PLACEHOLDER,
     QUERY_SYSTEM_INSTRUCTION,
     USE_CONTEXT_INSTRUCTION,
     USE_HISTORY_INSTRUCTION,
 )
 
-PromptConfiguration = namedtuple("PromptConfiguration", "provider model")
 
-# This dictionary might be expanded in the future, if some combination of model+provider
-# requires specifically updated prompt.
-prompt_configurations = {
-    PromptConfiguration(PROVIDER_BAM, GRANITE_13B_CHAT_V2): QUERY_SYSTEM_INSTRUCTION,
-    PromptConfiguration(PROVIDER_OPENAI, GPT35_TURBO): QUERY_SYSTEM_INSTRUCTION,
-    PromptConfiguration(
-        PROVIDER_WATSONX, GRANITE_13B_CHAT_V2
-    ): QUERY_SYSTEM_INSTRUCTION,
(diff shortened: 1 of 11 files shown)


tests/unit/utils/test_token_handler.py
```diff
@@ -109,46 +109,59 @@ def test_available_tokens_with_buffer_weight(self):
...
 
     @mock.patch("ols.utils.token_handler.TOKEN_BUFFER_WEIGHT", 1.05)
+    @mock.patch("ols.utils.token_handler.MINIMUM_CONTEXT_TOKEN_LIMIT", 1)
     @mock.patch("ols.utils.token_handler.RAG_SIMILARITY_CUTOFF", 0.4)
     def test_token_handler(self):
         """Test token handler for context."""
         retrieved_nodes = self._mock_retrieved_obj[:3]
         rag_chunks, available_tokens = self._token_handler_obj.truncate_rag_context(
-            retrieved_nodes
+            retrieved_nodes, "gpt"
```
tisnik, line 118: probably put "gpt" into constant?

ols/src/prompts/prompt_generator.py
```diff
@@ -1,78 +1,124 @@
...
+            prompt_message = prompt_message + "\n{chat_history}"
+
+        prompt_message = prompt_message + "\n<|user|>\n{query}\n<|assistant|>\n"
+        return PromptTemplate.from_template(prompt_message), llm_input_values
+
+    def generate_prompt(
+        self, model: str
+    ) -> tuple[ChatPromptTemplate | PromptTemplate, dict]:
+        """Generate prompt."""
+        if "granite" in model:
```
tisnik, line 121: don't we want to use constants instead? Like model type or whatever...

tisnik (commented):
it looks ok in overall!!
just have few nitpicks about having "granite" and "gpt" as constants. WDYT?

ols/src/prompts/prompt_generator.py
```diff
@@ -1,78 +1,126 @@
...
+    """Restructure history."""
+    if text.startswith("human: "):
+        if ModelFamily.GRANITE in model:
+            return "\n<|user|>\n" + text.removeprefix("human: ")
+    else:
+        if ModelFamily.GRANITE in model:
+            return "\n<|assistant|>\n" + text.removeprefix("ai: ")
+
+    # No processing required here for gpt.
+    return text
```
xrajesh, line 47: @asamal4 If the model is not Granite , we don't have to do any of this checks right , can we avoid those checks something like this ?
```
def restructure_history(text: str, model: str) -> str:
    """Restructure history."""
    # if model is not granite, then no processing required.
    if ModelFamily.GRANITE not in model:
        return text

    if text.startswith("human: "):
            return "\n<|user|>\n" + text.removeprefix("human: ")
    else:
        if text.startswith("ai: "):
            return "\n<|assistant|>\n" + text.removeprefix("ai: ")
    return text
```

ols/src/prompts/prompt_generator.py
```diff
@@ -1,78 +1,126 @@
...
+
+    def _generate_prompt_gpt(self) -> tuple[ChatPromptTemplate, dict]:
+        """Generate prompt for GPT."""
+        prompt_message = []
+        sys_intruction = self._sys_instruction.strip() + "\n"
+        llm_input_values: dict = {"query": self._query}
+
+        if len(self._rag_context) > 0:
+            llm_input_values["context"] = "".join(self._rag_context)
+            sys_intruction = sys_intruction + "\n" + USE_CONTEXT_INSTRUCTION.strip()
```
xrajesh, line 74: @asamal4 Do we need this \n here ? we already have in the line 69 - may be needed , trying to understand .

tisnik (commented):
it looks ok

tisnik:

tisnik:

PR #1446, opened 2024-08-20
update openai version to 1.41

asamal4:
## Description

update openai version to 1.41
[OLS-924](https://issues.redhat.com//browse/OLS-924)


tisnik:
it was superseded already

PR #1486, opened 2024-08-27
OLS-941: disable query val by default

asamal4:
## Description

- Disable query val by default.
- Necessary modification to test cases
- Validate provider/model sent through request payload separately to have proper error message.


tisnik (commented):
it looks good!

tisnik:

PR #1512, opened 2024-08-31
increase max resp token default to 1024

asamal4:
## Description

Increase max resp token default to 1024. YAML in response may take more tokens, so increasing from 512.

This is just a default value; it is recommended to use correct context window and reasonable response token based on model & use.


tisnik (commented):
it looks ok

PR #1540, opened 2024-09-05
Removed f-strings from logging

tisnik:
## Description

The `logging` module provides a mechanism for passing additional values to
be logged using the `extra` keyword argument. This is more consistent, more
efficient, and less error-prone than formatting the string directly.

Using f-strings to format a logging message requires that Python eagerly
format the string, even if the logging statement is never executed (e.g.,
if the log level is above the level of the logging statement), whereas
using the `extra` keyword argument defers formatting until required.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update


JoaoFula:
Also, interesting to learn this, thanks @tisnik

PR #1550, opened 2024-09-06
Adding pyroscope and memray dependencies

vishnuchalla:
## Description

Adding pyroscope and memray dependencies

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update

## Closes
* https://issues.redhat.com/browse/OLS-964
* https://issues.redhat.com/browse/OLS-965

## Testing
Tested and verified in local. Both pyroscope and memray get triggered as a part of OLS's startup activity.


tisnik (commented):
would it be possible to add these dependencies only for "devel mode"? The same way as GradIO. It will keep number of packages on lowest level possible + images will be smaller + (most important) less CVE-prone

vishnuchalla:
> would it be possible to add these dependencies only for "devel mode"? The same way as GradIO. It will keep number of packages on lowest level possible + images will be smaller + (most important) less CVE-prone

We will need these dependencies when OLS is running through the operator so that we can capture profiles while load testing. Move them to them "dev' group will make it impossible.

tisnik:
> > would it be possible to add these dependencies only for "devel mode"? The same way as GradIO. It will keep number of packages on lowest level possible + images will be smaller + (most important) less CVE-prone
> 
> We will need these dependencies when OLS is running through the operator so that we can capture profiles while load testing. Moving them to them `dev` group will make it impossible.

ok, but at least loading should be performed if really needed (and configured in config.yaml) right?

vishnuchalla:
> > > would it be possible to add these dependencies only for "devel mode"? The same way as GradIO. It will keep number of packages on lowest level possible + images will be smaller + (most important) less CVE-prone
> > 
> > 
> > We will need these dependencies when OLS is running through the operator so that we can capture profiles while load testing. Moving them to them `dev` group will make it impossible.
> 
> ok, but at least loading should be performed if really needed (and configured in config.yaml) right?

When we say loading, are you referring to `import` statement in specific?

tisnik:
> > > > would it be possible to add these dependencies only for "devel mode"? The same way as GradIO. It will keep number of packages on lowest level possible + images will be smaller + (most important) less CVE-prone
> > > 
> > > 
> > > We will need these dependencies when OLS is running through the operator so that we can capture profiles while load testing. Moving them to them `dev` group will make it impossible.
> > 
> > 
> > ok, but at least loading should be performed if really needed (and configured in config.yaml) right?
> 
> When we say loading, are you referring to `import` statement in specific?

Yes, in the similar way we "optimized" Gradio packages loading

vishnuchalla:
> > > > would it be possible to add these dependencies only for "devel mode"? The same way as GradIO. It will keep number of packages on lowest level possible + images will be smaller + (most important) less CVE-prone
> > > 
> > > 
> > > We will need these dependencies when OLS is running through the operator so that we can capture profiles while load testing. Moving them to them `dev` group will make it impossible.
> > 
> > 
> > ok, but at least loading should be performed if really needed (and configured in config.yaml) right?
> 
> When we say loading, are you referring to `import` statement in specific?

I just conducted some experiments to re-evaluate your suggestion. I think moving them to `dev` makes sense. 
And profiling under load doesn't add any value because of huge amount of data that it spits and is impossible to interpret. Example memory profile from `memray`
```
-rw-r--r--. 1 1000700000 root       5.2G Sep 11 02:59 output.bin
```
And pyroscope also captured information that is impossible to interpret. So In order to get better understand on profiling, keeping them in `dev` group and capturing profiles against one single profile is sufficient enough to take a look at things in detail.
cc: @syedriko @xrajesh

Makefile
```diff
@@ -48,6 +48,16 @@ update-deps: ## Check pyproject.toml for changes, update the lock file if needed
 run: ## Run the service locally
 	python runner.py
 
+memray-run: ## Run the service locally using memray
+	python -m memray run -o ./memory_profile.bin runner.py
+
+memray-flamegraph: ## Extract flamegraph from memray output
```
tisnik, line 54: you can make it as dependency for memray-run, right

runner.py
```diff
@@ -148,10 +149,47 @@ def start_uvicorn():
...
-    # create and start the rag_index_thread - allows loading index in
-    # parallel with starting the Uvicorn server
-    rag_index_thread = threading.Thread(target=load_index)
-    rag_index_thread.start()
+    if config.dev_config.pyroscope_url:
+        try:
+            import requests
+
+            response = requests.get(config.dev_config.pyroscope_url, timeout=60)
+            if response.status_code == 200:
```
tisnik, line 157: requests.codes.ok

Makefile
vishnuchalla: Changed as suggested.

vishnuchalla:
@tisnik can you take a look on this one as well please?

vishnuchalla:
> /hold

@tisnik Is there a reason to hold this PR?

vishnuchalla:
@tisnik @xrajesh @syedriko since we are post GA, can we unhold and take this forward?

vishnuchalla:
@xrajesh @tisnik 👋🏼  PTAL once you get some time. Thank you

pyproject.toml
```diff
@@ -56,6 +56,8 @@ dev = [
     "types-requests==2.32.0.20240622",
     "gradio==4.44.1",
     "boto3==1.34.145",
+    "pyroscope-io>=0.8.7",
```
tisnik, line 59: please lock version for both new packages

pyproject.toml
```diff
@@ -56,6 +56,8 @@ dev = [
...
+    "memray>=1.14.0",
```
tisnik, line 60: dtto

pdm.lock
```diff
@@ -4,8 +4,11 @@
 [metadata]
```
tisnik, line 4: was it generated on MacOS? seems a bit weird that so many changes were incorporated

runner.py
```diff
@@ -60,6 +59,43 @@ def load_index():
     # init loading of query redactor
     config.query_redactor
 
+    if config.dev_config.pyroscope_url:
```
tisnik, line 62: could you please move it somewhere into `ols/src/utils/pyroscope` or so? TYVM!

tisnik (changes requested):
Seems fine in overall, just have few comments/nitpicks. Please look at them

pyproject.toml
vishnuchalla: Done

pyproject.toml
vishnuchalla: locked as suggested.

pdm.lock
vishnuchalla: I used a fedora linux machine.

runner.py
vishnuchalla: Moved it to `ols/utils/pyroscope` as suggested.

vishnuchalla:
@xrajesh @tisnik looks like good to go. Went through all the CI tests

tisnik:
/lgtm
/approve

PR #1593, opened 2024-09-14
[Bug fix] Race condition while creating feedback directory

vishnuchalla:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/app/endpoints/feedback.py
+++ b/ols/app/endpoints/feedback.py
@@ -63,7 +63,7 @@ def store_feedback(user_id: str, feedback: dict) -> None:
     storage_path = Path(config.ols_config.user_data_collection.feedback_storage)
     if not storage_path.exists():
         logger.debug(f"creating feedback storage directories '{storage_path}'")
-        storage_path.mkdir(parents=True)
+        storage_path.mkdir(parents=True, exist_ok=True)
 
     current_time = str(datetime.utcnow())
     data_to_store = {"user_id": user_id, "timestamp": current_time, **feedback}


onmete:
Please put a comment into a code about why it is done. Otherwise, it might confuse someone in the future as from a code perspective without any context it doesn't make much sense (there is an `if` that checks this). The `if` removal would work too, but the comment would be better.

vishnuchalla:
> Please put a comment into a code about why it is done. Otherwise, it might confuse someone in the future as from a code perspective without any context it doesn't make much sense (there is an `if` that checks this). The `if` removal would work too, but the comment would be better.

Changed as suggested.

onmete:
Maybe something more specific?
```python
# Creates storage path only if it doesn't exist. The `exist_ok=True` prevents
# race conditions in case of multiple server instances trying to set up storage
# at the same location.
```
And does `transcripts` storage suffer the same issue?

vishnuchalla:
> Maybe something more specific?
> 
> ```python
> # Creates storage path only if it doesn't exist. The `exist_ok=True` prevents
> # race conditions in case of multiple server instances trying to set up storage
> # at the same location.
> ```
> 
> And does `transcripts` storage suffer the same issue?

Good point. Updated it too. Default config did not have `transcripts` configured so missed on it.

tisnik (commented):
it looks ok

tisnik:
/lgtm
/approve

PR #1622, opened 2024-09-18
feedback.py: Don't use f-strings in logger

tisnik:
## Description

`feedback.py`: Don't use f-strings in logger

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update

### DIFF
--- a/ols/app/endpoints/feedback.py
+++ b/ols/app/endpoints/feedback.py
@@ -73,7 +73,7 @@ def store_feedback(user_id: str, feedback: dict) -> None:
     with open(feedback_file_path, "w", encoding="utf-8") as feedback_file:
         json.dump(data_to_store, feedback_file)
 
-    logger.debug(f"feedback stored in '{feedback_file_path}'")
+    logger.debug("feedback stored in '%s'", feedback_file_path)
 
 
 @router.get("/status")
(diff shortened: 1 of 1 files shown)


onmete:
/hold

What is wrong with fstrings?

onmete:
I understand that, but using fstrings is generally more convenient and readable. The performance impact of evaluating fstring when logging is set to a higher level is less than negligible.

It is recommended by the logging, but the community is split on this topic. To me, readability has a better value here.

Either way, we should be consistent in the logging. I see we have both approaches in our code (for different levels). This PR should unify it across the whole codebase and not just small part of it - this is unnecessarily atomic change, especially with our problems/waiting times on CI.

tisnik:
> Either way, we should be consistent in the logging. I see we have both approaches in our code (for different levels). This PR should unify it across the whole codebase and not just small part of it - this is unnecessarily atomic change, especially with our problems/waiting times on CI.

we are fixing it everywhere as time constraints allow us

onmete:
Isn't it like 30 minutes at most to change it everywhere?

PR #1661, opened 2024-09-25
OLS-1067 config: add api_version required for azure

asamal4:
## Description

Add api_version to be configurable [Optional]. This is currently required for azure. Default version is used when it is not set for azure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ Depends on configuration, but usually it is not needed to generate or use API ke
 
    3. Specific configuration options for Azure OpenAI
 
+       - `api_version`: as specified in official documentation, if not set; by default `2024-02-15-preview` is used.
        - `deployment_name`: as specified in AzureAI project settings
 
    4. Default provider and default model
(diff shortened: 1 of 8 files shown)


tisnik (changes requested):
please update/add tests also into `tests/unit/app/models/test_config.py`

seems LGTM instead

tisnik:
> > please update/add tests also into `tests/unit/app/models/test_config.py`
> > seems LGTM instead
> 
> @tisnik Test cases are updated here https://github.com/openshift/lightspeed-service/pull/1661/files#diff-c491ee1688f00aa59e05f7efabde3fac2698cba49f5c046a582f6510947e4e5f for below 3 scenarios..
> 
>     * default value for azure when not set
> 
>     * same as config value when set for azure
> 
>     * None for other provider (watsonx)

sry I meant also `tests/unit/utils/test_config.py`. Or... I can do it later if you prefer fast merge...

tisnik:
/override "Red Hat Konflux / lightspeed-service-on-pull-request "

tisnik:

PR #1665, opened 2024-09-25
OLS-1072: Include OCP 4.17

onmete:
## Description
 include 4.17 ref


tisnik (commented):
it looks legit

tisnik:

PR #1712, opened 2024-10-07
Handle decode error

onmete:
## Description

Catch and log when there is something wrong in get of refresh token.


tisnik (commented):
it looks ok

tisnik:
/lgtm

PR #1727, opened 2024-10-08
Removing the get from the checks if the resources exist

JoaoFula:
We can just try to delete since we have the error catch underneath.


tisnik:

PR #1730, opened 2024-10-08
Sync with upstream

tisnik:
## Description

Sync with upstream


tisnik:

PR #1785, opened 2024-10-20
Disable Mypy temporarily

tisnik:
## Description

Disable Mypy temporarily


tisnik:

PR #1807, opened 2024-10-24
fix: history structure for granite prompt

asamal4:
## Description

Fix: Use formatted history/messages for granite prompt.


tisnik (commented):
it look correct

xrajesh:

PR #1813, opened 2024-10-24
Proper image version

tisnik:
## Description

Proper image version


xrajesh:

PR #1981, opened 2024-12-01
eval: clean up QnAs

asamal4:
## Description

Clean up eval QnAs in parquet file per review comments from experts.
Also did some basic clean up and removed some of the queries which are not very useful. Details can be found here [OLS-1207](https://issues.redhat.com//browse/OLS-1207)


tisnik (commented):
it looks ok!

PR #2000, opened 2024-12-03
OLS-39: Remove unused ReferencedDocument method

onmete:
## Description

Remove unused ReferencedDocument method

## Type of change

- [x] Refactor

### DIFF
--- a/tests/unit/app/models/test_models.py
+++ b/tests/unit/app/models/test_models.py
@@ -1,7 +1,5 @@
 """Unit tests for the API models."""
 
-import json
-
 import pytest
 from pydantic import ValidationError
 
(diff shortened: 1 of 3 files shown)


tests/unit/app/models/test_models.py
```diff
@@ -94,38 +93,6 @@ def test_llm_response():
...
-            conversation_id=conversation_id,
-            response=response,
-            referenced_documents=referenced_documents,
-            truncated=False,
-        )
-
-        assert llm_response.conversation_id == conversation_id
-        assert llm_response.response == response
-        assert llm_response.referenced_documents == referenced_documents
-        assert not llm_response.truncated
```
onmete, line 127: The removed method was used only in tests.
This test was testing that reference doc can be loaded from json.
We can do something like this for every model (dataclass), but I don't see value in doing this.
```python
def test_referenced_docs_from_json():
    """Test the ReferencedDocument model can be loaded from JSON."""
    ref_docs = ReferencedDocument("a", "b")
    ref_docs_json = json.dumps(ref_docs.__dict__)
    assert ReferencedDocument(**json.loads(ref_docs_json)) == ref_docs
```

tisnik (commented):
yes, it must be some leftover there, let's remove it

tisnik:

PR #2001, opened 2024-12-03
OLS-39: Ability to create ref docs from rag chunks

onmete:
## Description

Ability to create ref docs from rag chunks

## Type of change

- [x] Refactor


ols/app/models/models.py
```diff
@@ -135,6 +135,18 @@ class ReferencedDocument:
     docs_url: str
     title: str
 
+    @classmethod
```
tisnik, line 138: why it needs to be class method?

onmete: Can be static - changed.

onmete:
I accidentally pushed main head here - the PR closed itself. New one https://github.com/openshift/lightspeed-service/pull/2003

PR #2002, opened 2024-12-03
OLS-39: Remove condition for query validation if it is a follow-up question

onmete:
## Description

Remove condition for query validation if it is a follow-up question

## Type of change

- [x] Refactor
- [x] New feature

### DIFF
--- a/ols/app/endpoints/ols.py
+++ b/ols/app/endpoints/ols.py
@@ -120,12 +120,7 @@ def conversation_request(
     validate_requested_provider_model(llm_request)
 
     # Validate the query
-    if not previous_input:
-        valid = validate_question(conversation_id, llm_request)
-    else:
-        logger.debug("follow-up conversation - skipping question validation")
-        valid = True
-
+    valid = validate_question(conversation_id, llm_request)
     timestamps["validate question"] = time.time()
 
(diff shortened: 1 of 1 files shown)


ols/app/endpoints/ols.py
```diff
@@ -120,12 +120,7 @@ def conversation_request(
     validate_requested_provider_model(llm_request)
 
     # Validate the query
-    if not previous_input:
-        valid = validate_question(conversation_id, llm_request)
-    else:
-        logger.debug("follow-up conversation - skipping question validation")
-        valid = True
```
onmete, line 127: TBD - I don't see much sense in having it in code when we agreed we don't want to do that. If anything, this should be part of validation configuration (and not hardcoded) - if only initial question or also follow-ups should be validated.
So I would remove this and once there is a need/usecase for having question validation (besides system instructions), let's do proper configuration.

onmete: @tisnik @asamal4 WDYT?

xrajesh: This change does not modify the current behavior as we dont do any validation by default. 
/lgtm

tisnik:
/approve

PR #2003, opened 2024-12-03
OLS-39: Ability to create ref docs from rag chunks

onmete:
## Description

Ability to create ref docs from rag chunks

## Type of change

- [x] Refactor

### DIFF
--- a/ols/app/models/models.py
+++ b/ols/app/models/models.py
@@ -1,5 +1,6 @@
 """Data models representing payloads for REST API calls."""
 
+from collections import OrderedDict
 from typing import Optional, Self
 
 from pydantic import BaseModel, field_validator, model_validator
@@ -123,7 +124,7 @@ class LLMRequest(BaseModel):
         return self
 
 
-@dataclass
+@dataclass(frozen=True, unsafe_hash=False)
 class ReferencedDocument:
     """RAG referenced document.
 
@@ -134,6 +135,22 @@ class ReferencedDocument:
 
     docs_url: str
     title: str
+
+    @staticmethod
+    def from_rag_chunks(rag_chunks: list["RagChunk"]) -> list["ReferencedDocument"]:
+        """Create a list of ReferencedDocument from a list of rag_chunks.
+
+        Order of items is preserved.
+        """
+        return list(
+            OrderedDict(
+                (
+                    rag_chunk.doc_url,
+                    ReferencedDocument(rag_chunk.doc_url, rag_chunk.doc_title),
+                )
+                for rag_chunk in rag_chunks
+            ).values()
+        )
 
 
 class LLMResponse(BaseModel):
--- a/tests/unit/app/endpoints/test_ols.py
+++ b/tests/unit/app/endpoints/test_ols.py
@@ -665,38 +665,6 @@ def test_conversation_request(
 
 
 @pytest.mark.usefixtures("_load_config")
-@patch("ols.src.query_helpers.question_validator.QuestionValidator.validate_question")
-@patch("ols.src.query_helpers.docs_summarizer.DocsSummarizer.summarize")
-@patch("ols.config.conversation_cache.get")
-def test_conversation_request_dedup_ref_docs(
-    mock_conversation_cache_get,
-    mock_summarize,
-    mock_validate_question,
-    auth,
-):
-    """Test deduplication of referenced docs."""
-    mock_rag_chunk = [
-        RagChunk("text1", "url-b", "title-b"),
-        RagChunk("text2", "url-b", "title-b"),  # duplicate doc
-        RagChunk("text3", "url-a", "title-a"),
-    ]
-    mock_validate_question.return_value = True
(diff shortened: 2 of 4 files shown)


ols/app/models/models.py
```diff
@@ -135,6 +135,18 @@ class ReferencedDocument:
     docs_url: str
     title: str
 
+    @staticmethod
+    def from_rag_chunks(rag_chunks: list["RagChunk"]) -> list["ReferencedDocument"]:
+        """Create a list of ReferencedDocument from a list of rag_chunks."""
+        return list(
+            {
+                rag_chunk.doc_url: ReferencedDocument(
```
tisnik, line 143: it is in original code, I know. But why we create dict just to convert it to list in the next step?

onmete: To drop duplicates.

tisnik: `list(set(rag_chunks))`

onmete: Sure, but ReferencedDocument needs to be hashable and I need to modify the other test to not relly on the order.
I can do that, no problem.

onmete: updated

tisnik (commented):
it looks ok

ols/app/models/models.py
onmete: While dict kinda retains the order of keys as they are inserted, if it is crucial to retain the same order as is in the rag_chunks, I would do this then.

```python
list(
    OrderedDict(
        (rag_chunk.doc_url, ReferencedDocument(rag_chunk.doc_url, rag_chunk.doc_title))
        for rag_chunk in summarizer_response.rag_chunks
    ).values()
)
```

onmete: updated

tests/unit/app/endpoints/test_ols.py
```diff
@@ -664,38 +664,6 @@ def test_conversation_request(
...
-    mock_summarize.return_value = SummarizerResponse(
-        response="some response",
-        rag_chunks=mock_rag_chunk,
-        history_truncated=False,
-    )
-    llm_request = LLMRequest(query="some query")
-    response = ols.conversation_request(llm_request, auth)
-
-    assert len(response.referenced_documents) == 2
-    assert response.referenced_documents[0].docs_url == "url-b"
```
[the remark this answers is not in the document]
onmete: It is tested in the new test.

ols/app/models/models.py
```diff
@@ -135,6 +136,19 @@ class ReferencedDocument:
     docs_url: str
     title: str
 
+    @staticmethod
+    def from_rag_chunks(rag_chunks: list["RagChunk"]) -> list["ReferencedDocument"]:
```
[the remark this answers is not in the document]
onmete: updated docstring

tests/unit/app/endpoints/test_ols.py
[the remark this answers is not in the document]
onmete: What exactly do you think is missing, because the current test tests the order as well.

tests/unit/app/models/test_models.py
```diff
@@ -411,3 +412,20 @@ def test_cache_entries_to_history_no_response():
             "human: what?",
             "ai: ",
         ]
+
+
+def test_ref_docs_from_rag_chunks():
+    """Test the ReferencedDocument model method `from_rag_chunks`."""
+    rag_chunk_1 = RagChunk("bla1", "url1", "title1")
+    rag_chunk_2 = RagChunk("bla2", "url2", "title2")
```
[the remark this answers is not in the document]
onmete: this passes as well
```python
def test_ref_docs_from_rag_chunks():
    """Test the ReferencedDocument model method `from_rag_chunks`."""
    rag_chunk_1 = RagChunk("bla2", "url2", "title2")
    rag_chunk_2 = RagChunk("bla1", "url1", "title1")
    rag_chunk_3 = RagChunk("bla2", "url2", "title2")  # duplicated doc

    ref_docs = ReferencedDocument.from_rag_chunks(
        [rag_chunk_1, rag_chunk_2, rag_chunk_3]
    )
    expected = [
        ReferencedDocument(docs_url="url2", title="title2"),
        ReferencedDocument(docs_url="url1", title="title1"),
    ]

    assert ref_docs == expected
```

tests/unit/app/endpoints/test_ols.py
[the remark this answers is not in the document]
onmete: So you are afraid that it is passing because the input is already sorted?
This is not the case.

However if I add sorting into tested logic, the test will pass, but the behavior will be different. That is a problem with only two values/docs (the original test suffers from the same).

I've added a third to ensure there is no hidden sorting.

tisnik (commented):
ok

tisnik:

tisnik:

tisnik:

PR #2045, opened 2024-12-11
OLS-1290 - Added better method to check if containers are ready

JoaoFula:
## Type of change

- [ x] Test fix


xrajesh:
/lgtm
/approve

PR #2123, opened 2025-01-09
chore: remove unnecessary exact model name reference during test

asamal4:
## Description

Removed unnecessary exact model name reference in test modules.

**Background**:
Recently we moved to granite 3 model for CI e2e test. There were several files got modified just to remove any reference to granite 2 in test code. But we don't have any actual logic which checks the exact model name, so having exact name in test cases is not really needed. And this also creates confusion that actual model name is used in app code.

_So removing exact model name reference in unit/integration tests (actual e2e test requires correct/exact model names)_

### DIFF
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -58,16 +58,6 @@ class ModelFamily(StrEnum):
     GRANITE = "granite"
 
 
-# BAM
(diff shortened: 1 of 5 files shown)


tisnik:
Yes, it looks ok
/lgtm
/approve

PR #2126, opened 2025-01-10
fix(KONFLUX-6218): align repository ids to cpe mapping

ralphbean:
This update changes the rpm repository ids to match those found in Red Hat's [repository-to-cpe.json](https://security.access.redhat.com/data/meta/v1/repository-to-cpe.json) mapping file, used by third-party scanners.

In order for scanners like clair to understand what [CPE](https://cpe.mitre.org/) a Red Hat rpm is associated with, it needs to be able to find its repository in Red Hat's published mapping file.

Even though some "arch-less" repository ids currently appear in the mapping file, they are going to be removed soon and will be blocked by Konflux in the future in https://github.com/release-engineering/rhtap-ec-policy/pull/99.


ralphbean:
I made a mistake here with `ubi-9-codeready-builder-source`. I'm going to close this and resubmit. Apologies for the noise.

PR #2131, opened 2025-01-12
OLS-1228: readable token counter textual representation

tisnik:
## Description

[OLS-1228](https://issues.redhat.com//browse/OLS-1228): readable token counter textual representation

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #[OLS-1228](https://issues.redhat.com//browse/OLS-1228)

### DIFF
--- a/ols/app/metrics/token_counter.py
+++ b/ols/app/metrics/token_counter.py
@@ -101,6 +101,16 @@ class GenericTokenCounter(BaseCallbackHandler):
         """Compute tokens count for given input text."""
         return len(self.token_handler.text_to_tokens(text))
 
+    def __str__(self) -> str:
+        """Textual representation of GenericTokenCounter instance."""
+        return (
+            f"{self.__class__.__name__}: "
+            + f"input_tokens: {self.input_tokens} "
+            + f"output_tokens: {self.output_tokens} "
+            + f"counted: {self.input_tokens_counted} "
+            + f"LLM calls: {self.llm_calls}"
+        )
+
 
 class TokenMetricUpdater:
     """A context manager to update token metrics in a callback handler.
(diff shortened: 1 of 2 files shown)


ols/app/metrics/token_counter.py
```diff
@@ -101,6 +101,16 @@ def tokens_count(self, text: str) -> int:
         """Compute tokens count for given input text."""
         return len(self.token_handler.text_to_tokens(text))
 
+    def __str__(self) -> str:
+        """Textual representation of GenericTokenCounter instance."""
+        return (
+            f"{self.__class__.__name__}: "
+            + f"input_tokens: {self.input_tokens} "
```
JoaoFula, line 108: Out of curiosity, why do we use f strings here when we've been removing f strings from almost everywhere. Makes it more readable?

tisnik: ah yes, we removed them from logging statements. Why? Because if logging is disabled or set to value like FATAL, all those "lower level" logging like log.DEBUG will still be evaluated. Not this one, because it is evaluated just when somebody actually PRINTS the value. I mean - this statement is in method and it is not evaluated when the method is not called.

tisnik:
/approve

ols/app/metrics/token_counter.py
JoaoFula: ok, so for example in test cases, it's ok to continue using f strings since it's more readable no?
From what I remember reading, it's slightly slower tho right? Since it's further from the C integration?

PR #2154, opened 2025-01-16
Loosen deps

onmete:
## Description

Trying to loose deps


tisnik (commented):
why is the sections in pyproject.toml moved into new places? Please keep the PR minimal.

Also don't we need to provide at least minimal major.minor versions for all packages?

Also black and mypy will be problematic if not pinned.

onmete:
- The pyproject.toml wasn't manual job, I guess the new pdm version did the reorganization.
- minimal major.minor - then you need to periodically also increase the minimal version of all libs, no?
- black and mypy will be problematic if not pinned - can pin

tisnik:
> * The pyproject.toml wasn't manual job, I guess the new pdm version did the reorganization.

probably. Could you please minimize changes made there?
And check PDM version you use locally?

> 
>     * minimal major.minor - then you need to periodically also increase the minimal version of all libs, no?
no, just we'll keep it as is

> 
>     * black and mypy will be problematic if not pinned - can pin

onmete:
@tisnik 
Added minor versions.
Black and mypy pinned.
I've pinned the pdm to the previous version, but the pyproject.toml is still shuffled a bit, but I don't think it is that big of a deal. The changes are readable.

tisnik (changes requested):
can you please minimize changes made in `pyproject.toml`? Don't move/rename sections etc.

tisnik (changes requested):
Keep these lines there too please:
```
# The following section is needed only for torch[cpu] variant on Linux,
# because this variant need to be downloaded from external link, not from
# standard Python package index:
#
# [[tool.pdm.source]]
# type = "find_links"
# url = "https://download.pytorch.org/whl/cpu/torch_stable.html"
# name = "torch"
````

File: requirements-build.txt [hunk 1]
```diff
@@ -1,16 +1,12 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.10
```
JoaoFula, line 2: shouldn't we be generating things with Python 3.11 or above?
This is just a comment, I'm aware, but out of coherency?

onmete: fixed

JoaoFula:
Hi @onmete , this failure seems a bit suspicious https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/2154/pull-ci-openshift-lightspeed-service-main-4.17-e2e-ols-cluster/1887513017717886976/artifacts/e2e-ols-cluster/e2e/artifacts/azure_openai/cluster/podlogs/lightspeed-app-server-65b8c56756-bv7px-lightspeed-service-api.log

Let's see this next run but if the failure remains, we might have a dependency issue happening right now.

tisnik:
> Hi @onmete , this failure seems a bit suspicious https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_lightspeed-service/2154/pull-ci-openshift-lightspeed-service-main-4.17-e2e-ols-cluster/1887513017717886976/artifacts/e2e-ols-cluster/e2e/artifacts/azure_openai/cluster/podlogs/lightspeed-app-server-65b8c56756-bv7px-lightspeed-service-api.log
> 
> Let's see this next run but if the failure remains, we might have a dependency issue happening right now.

missing deps perhaps

tisnik:
`pyaml` seems to be missing, indeed

onmete:
Seems the problem is somewhere here `scripts/generate_packages_to_prefetch.py -a x86_64`.
The build is working with all the hashes (but not in konflux/cachi size limit).

onmete:
@tisnik @JoaoFula 
So what I did here was create a new lock for the platform (for the container file) and generate requirements from this lock.
We should keep the non-platform specific lock for dev environements.
```
pdm update
pdm lock --platform=manylinux_2_34_x86_64 --lockfile=pdm-manylinux.lock
pdm export -f requirements -o requirements.txt --lockfile=pdm-manylinux.lock
```
And then, one needs to manually revert the httpx and sqlalchemy in requirements.txt

WDYT?

pyproject.toml
```diff
@@ -142,27 +120,43 @@ license = {file = "LICENSE"}
...
+    "memray>=1.15.0",
+    "reportportal-client>=5.5.6",
+    "pytest-reportportal>=5.4.1",
+    "pytest-benchmark[histogram]>=4.0.0",
+    "typing-extensions>=4.12.2",
+    "pytest-subtests>=0.13.1",
+    "build>=1.2.2.post1",
+    "twine>=5.1.1",
+    "pylint>=3.3.2",
+    "matplotlib",
```
tisnik, line 161: minimal version?

tisnik (changes requested):

Makefile
```diff
@@ -123,9 +123,7 @@ schema:	## Generate OpenAPI schema file
 	python scripts/generate_openapi_schema.py docs/openapi.json
 
 requirements.txt:	pyproject.toml pdm.lock ## Generate requirements.txt file containing hashes for all non-devel packages
-	pip-compile --generate-hashes pyproject.toml
-	# please note that it is also possible to use the following command, but it generates sligthly less readable output file:
-	# pdm export --prod --format requirements --no-extras --output requirements.txt
+	pdm export --prod --format requirements --output requirements.txt --no-extra
```
tisnik, line 126: should be `--no-extras` as `pdm export --help` says

currently it seems, that `--no-extra` works, but it's probably PDM bug that might be fixed later :)

tisnik (changes requested):

tisnik:

PR #2216, opened 2025-02-07
OLS-1379: Add oc tools

onmete:
## Description

Adding oc tools.

## Type of change

- [x] New feature


ols/src/tools/oc_cli.py
```diff
@@ -0,0 +1,137 @@
...
+        oc get pod test-pod -o
+        custom-columns=CONTAINER:.spec.containers[0].name,IMAGE:.spec.containers[0].image
+
+        # List all replication controllers and services together in ps output format.
+        oc get rc,services
+
+        # List one or more resources by their type and names.
+        oc get rc/web service/frontend pods/web-pod-13je7
+    """
+    result = run_oc(["get", *sanitize_oc_args(args)])
```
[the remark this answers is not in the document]
onmete: yes, besides sanitization, we can also catch the case when llm returns not just the args, but the full command `oc get whatever` instead of just `whatever`.

onmete:
@asamal4 @tisnik I've pushed the oc tools. Check it out please, I'll add tests later (not e2e for tools).

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -150,18 +151,59 @@ def _invoke_llm(
...
+
+    def _invoke_llm_with_tools(
+        self,
+        messages: list,
+        llm_input_values: dict,
+        tools_map: dict,
+        is_final_round: bool,
+    ) -> tuple[AIMessage, TokenCounter]:
+        """Invoke LLM with or without tools based on conditions."""
+        llm = (
```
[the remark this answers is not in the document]
onmete: updated

onmete:
@asamal4 Updated the question, is this what you were after?

ols/src/tools/oc_cli.py
```diff
@@ -0,0 +1,238 @@
...
+    blocked_chars = {";", "&", "|", "`", "$", "(", ")", "<", ">", "\\"}
+
+    sanitized_args = []
+    for arg in args:
+        safe_part = []
+        for char in arg:
+            if char in blocked_chars:
+                # stop processing further characters in this argument
+                logger.warning(
+                    "Problematic character(s) found in oc tool argument '%s'", arg
```
tisnik, line 26: probably we should stop processing such input as is, not try to sanitize it further

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -131,37 +130,70 @@ def _prepare_prompt(
 
     def _invoke_llm(
         self,
-        llm: LLM | RunnableBinding,
-        prompt: PromptTemplate | ChatPromptTemplate,
-        prompt_input: dict,
+        messages: list,
```
tisnik, line 133: list of what? Any?

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -131,37 +130,70 @@ def _prepare_prompt(
...
+        llm_input_values: dict,
```
tisnik, line 134: dict of what?

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -131,37 +130,70 @@ def _prepare_prompt(
...
+            # langchain magic for chaining
+            chain = messages | llm  # type: ignore
             out = chain.invoke(
-                input=prompt_input,
+                input=llm_input_values,
                 config={"callbacks": [generic_token_counter]},
             )
         return out, generic_token_counter.token_counter
 
+    def _get_available_tools(self, user_token: Optional[str]) -> dict:
```
tisnik, line 158: dict[str, tool] perhaps

tisnik (changes requested):

tisnik:
/approve

@JoaoFula if you are ok with the sanitization function, please LGTM

tisnik:

ols/src/tools/oc_cli.py
```diff
@@ -0,0 +1,238 @@
...
+                )
+                break
```
JoaoFula, line 28: if we have one forbidden argument do we want to process all other arguments?

JoaoFula: I think I misunderstood. We're really filtering the forbidden chars from the argument list right?
I'm not entirely sure I like that. Also, how likely is it to happen that our tools would be called with an accidental character in there? I feel if we'd have strange chars they'd be on purpose and if we're still running the commands but just removing said chars, we might run into potentially unexpected situations no? wdyt?

onmete: There is a whole story dedicated to this: https://issues.redhat.com/browse/OLS-1443

JoaoFula: so this is just a placeholder for the time being while we discuss the proper approach with ProdSec?

onmete: This PR contains just the first iteration of the tools. We don't know what tools we need and what should be the args of them. Most likely, this some sort of sanitization of llm args will be needed, but I wouldn't stress too much about it at this specific moment. We definitely should have this revisited and backed once we want to release/announce OLS with oc tools capabilities.

JoaoFula: ok, so do you wanna just add a 
//TODO: OLS-1443
?

ols/src/tools/oc_cli.py
```diff
@@ -0,0 +1,239 @@
...
+        if safe_part:  # only add non-empty results
+            sanitized_args.append("".join(safe_part).strip())
+
+    return sanitized_args
+
+
+def run_oc(args: list[str]) -> subprocess.CompletedProcess:
+    """Run `oc` CLI with provided arguments and command."""
+    try:
+        res = subprocess.run(  # noqa: S603
```
xrajesh, line 41: @onmete We are not passing the token to the actual command that is suggested by the model - We need to execute the command with the user token . Example : `oc describe pod lightspeed-operator-controller-manager-bc4576dbc-wfbh8  -n default --token="e..."` -  Can be done with security story

ols/src/tools/oc_cli.py
```diff
@@ -0,0 +1,239 @@
...
+            args,
+            e,
+            e.stdout,
+            e.stderr,
+        )
+        raise
+
+
+# TODO: server address configurable via request?
+def log_to_oc(
```
xrajesh, line 61: @onmete Could you please explain why we need this login ? Why we cannot directly call the commands from model ?

ols/src/tools/oc_cli.py
onmete: For each user question, the token is checked and used to login. All tool calls within the user question are executed with this.
Yes, I see the problem that the service is logged to oc until another question hits.
We can spawn a dedicated shell for the user question execution, that will go away with the response.
Or as you say, sending the token and server for each command.
Related to the comment bellow - I guess there is technically nothing wrong if the tools fails because of wrong token, just in this specific case we can know in advance. But we can go with failures for now and solve it later.

PR #2218, opened 2025-02-07
feat: enable tool calling in a loop

asamal4:
## Description
Change Summary
* [OLS-1383](https://issues.redhat.com/browse/OLS-1383)
   * Add a flag to the config to enable introspection
* [OLS-1373](https://issues.redhat.com//browse/OLS-1373)
   * Add sample tool with mock output
   * Tool calling in a loop when introspection flag is enabled
   * Stop the loop when max iteration is reached or model gives final response.
   * Existing flow when flag is False.
   * Minor modification to prompt when tool calling is enabled
   * Modify watsonx model class to handle bind_tools
   * some refactoring
* [OLS-1288](https://issues.redhat.com/browse/OLS-1288)
   * Use langchain runnable instead of LLMChain (Didn't change query validator, as we need to remove the logic)
* e2e test case
   * Added different suite with different marker for tool calling e2e test case. Currently commented (till we have property in operator crd)

Note
* Goal is to have the flow (tool calling in a loop) ready, so that we can iterate. Optimization is not in scope.
* Prompt optimization is not in scope
* Actual tools are not in scope
* Optimizing for granite not in scope
* There are several other enhancement, will create follow up stories.

### DIFF
--- a/tests/e2e/test_api.py
+++ b/tests/e2e/test_api.py
@@ -98,6 +98,7 @@ def fixture_postgres_connection():
 
 
 @pytest.mark.smoketest
+@pytest.mark.introspection
 @retry(max_attempts=3, wait_between_runs=10)
 def test_readiness():
     """Test handler for /readiness REST API endpoint."""
@@ -110,6 +111,7 @@ def test_readiness():
 
 
 @pytest.mark.smoketest
+@pytest.mark.introspection
 def test_liveness():
     """Test handler for /liveness REST API endpoint."""
     endpoint = "/liveness"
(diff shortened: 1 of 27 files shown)


onmete (commented):
Looks ok for the next steps.

tests/e2e/test_api.py
```diff
@@ -98,6 +98,7 @@ def fixture_postgres_connection():
 
 
 @pytest.mark.smoketest
+@pytest.mark.introspection
```
onmete, line 101: Why are test_readiness, test_liveness and test_valid_question marked as introspection?

onmete: Make a note to remove the tags after OLS-1404 then. Anything else blocking merge?

JoaoFula: Updated decription in OLS-1404

tisnik (commented):
it looks ok. Do you think it is a good idea to call the tool (whatever is) within the service context? I mean, that would be quite hard to put into upstream if it will be done this way.

tisnik:
/lgtm
/approve

PR #2225, opened 2025-02-11
Enfore http/1.0 to reedice flakines

onmete:
## Description

Hopefully some improvement in tests flakiness

### DIFF
--- a/tests/e2e/test_streaming_query_endpoint.py
+++ b/tests/e2e/test_streaming_query_endpoint.py
@@ -18,6 +18,19 @@ from . import test_api
 STREAMING_QUERY_ENDPOINT = "/v1/streaming_query"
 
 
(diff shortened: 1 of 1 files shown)


tisnik (commented):
Chunked HTTP is supported from HTTP 1.1, so how to enforce HTTP 1.0?

onmete:
Yea, it is not enforcing the protocol, just the behavior.

> HTTP/1.0 does not support persistent connections by default, so each request-response pair typically results in the connection being closed automatically. Therefore, setting Connection: close in HTTP/1.1 mimics the behavior of HTTP/1.0 regarding connection management but does not change the HTTP version being used.

This is just a shot to try to reduce the streaming tests flakiness.

tisnik:
/lgtm
/approve

PR #2226, opened 2025-02-12
OLS-1384: Update requirements.txt

onmete:
## Description

Update requirements.txt from the current lib set.
Fix the version too.


tisnik (commented):
should be it part of #2154 ?

onmete:
Nope, that one is too hard to make happen.

xrajesh:
Looks good - but  I am missing the to understand , what is driving this change - could you please post what is driving this change ? Thank you @onmete

PR #2228, opened 2025-02-13
eval: fix prompt & save temp score

asamal4:
## Description

- Fix relevancy prompt for other models
- Store calculated score to avoid re-calculation if script fails between models.


xrajesh:

PR #2240, opened 2025-02-18
Update lock and requirements

onmete:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing


Containerfile
```diff
@@ -31,7 +31,7 @@ COPY --from=lightspeed-rag-content /rag/embeddings_model ./embeddings_model
 # (avoid accidental inclusion of local directories or env files or credentials)
 COPY runner.py requirements.txt ./
 
-RUN for a in 1 2 3 4 5; do pip3.11 install --no-cache-dir -r requirements.txt && break || sleep 15; done
+RUN pip3.11 install --no-cache-dir -r requirements.txt
```
JoaoFula, line 34: Hi @onmete , why is this being removed? We need this for CI since it fails rather often without it.
We potentially need to include the returning of the error in case of failure in the last attempt but remove it all together is going to make CI flaky again.

onmete: As we discussed before, there is something weird going on with this.
The build in this PR passed with the loop, but it is failing without it (same behavior I get locally).
The build error is not something a repeat will fix, yet it somehow passes with the loop.

JoaoFula: ok, I failed to understand previously that the build was passing in the loop, I thought it was failing but not returning any error. And that it stopped failing due to us not hitting the exact failing condition. That's why I suggested the `|| exit 1` at the end.

PR #2266, opened 2025-02-26
eval: change according to latest history format

asamal4:
## Description

- Fix error due to recent history format change.
- Remove deprecated LLMChain.


tisnik (commented):
yeah, it is better approach

tisnik:
/lgtm
/approve

PR #2270, opened 2025-02-27
Make json events stream as default application/json format for streaming

onmete:
## Description

Make json events stream as default application/json format for streaming

## Type of change

- [x] Refactor


xrajesh:
/lgtm
/approve

PR #2272, opened 2025-03-03
eval: latest eval score

asamal4:
## Description
Publish Latest eval result.

Note:
Not publishing result for openai/gpt-4o-mini, as in past we have seen similar performance.
Judge model is changed to llama instead of gpt to avoid any bais.

[OLS-1466](https://issues.redhat.com//browse/OLS-1466)


tisnik (commented):
looks ok

tisnik:
/lgtm
/approve

PR #2302, opened 2025-03-07
fix: e2e local run - missing attribute

asamal4:
## Description

fix undefined attribute/var issue for local e2e run.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -37,10 +37,9 @@ def pytest_sessionstart():
     # OLS_URL env only needs to be set when running against a local ols instance,
     # when ols is run against a cluster the url is retrieved from the cluster.
     ols_url = os.getenv("OLS_URL", "")
+    pytest.on_cluster = False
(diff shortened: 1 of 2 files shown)


tests/e2e/conftest.py
```diff
@@ -37,10 +37,9 @@ def pytest_sessionstart():
     # OLS_URL env only needs to be set when running against a local ols instance,
     # when ols is run against a cluster the url is retrieved from the cluster.
     ols_url = os.getenv("OLS_URL", "")
+    pytest.on_cluster = False
     if "localhost" not in ols_url:
         pytest.on_cluster = True
-
-    if pytest.on_cluster:
```
JoaoFula, line 43: Why do we remove this check? If we remove it, we should then not raise an error if the installation fails no?

JoaoFula: there's a few extra stuff necessary, like declaring the on_cluster variable as global which the pytest.on_cluster cannot do.
This was what was causing the sessionfinish to not run the must_gather

PR #2310, opened 2025-03-10
Ability to use token in noop auth (if needed)

onmete:
## Description

Ability to use token in noop auth (if needed). When this auth is set in config and user posts a query with token, noop is forwarding this token then instead of default value.

### DIFF
--- a/ols/src/auth/auth_dependency_interface.py
+++ b/ols/src/auth/auth_dependency_interface.py
@@ -2,7 +2,33 @@
 
 from abc import ABC, abstractmethod
 
-from fastapi import Request
+from fastapi import HTTPException, Request
+
+
+def extract_bearer_token(header: str) -> str:
+    """Extract the bearer token from an HTTP authorization header.
+
+    Args:
+        header: The authorization header containing the token.
+
+    Returns:
+        The extracted token if present, else an empty string.
+    """
+    try:
+        scheme, token = header.split(" ", 1)
+        return token if scheme.lower() == "bearer" else ""
+    except ValueError:
+        return ""
+
+
+def extract_token_from_request(request: Request) -> str:
+    """Extract the bearer token from a FastAPI request."""
+    authorization_header = request.headers.get("Authorization")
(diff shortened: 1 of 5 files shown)


onmete:
@tisnik WDYT?
I have this need to post the token to my local instance when testing the tools. I'm fine if there is another workaround.

tisnik:
/override "Red Hat Konflux / integration-tests / lightspeed-service"

ols/src/auth/auth_dependency_interface.py
```diff
@@ -2,7 +2,33 @@
...
+        return token if scheme.lower() == "bearer" else ""
+    except ValueError:
+        return ""
+
+
+def extract_token_from_request(request: Request) -> str:
+    """Extract the bearer token from a FastAPI request."""
+    authorization_header = request.headers.get("Authorization")
+    if not authorization_header:
+        raise HTTPException(
```
joshuawilson, line 28: What else could cause this to fail? I am concerned we pass along `no user token` for other failures.

onmete: It fail when there is no auth header. This is just moved logic from https://github.com/openshift/lightspeed-service/pull/2310/files#diff-c5c14c799ad92af71ed616b2187df94744b2986503a12ffad69f4a3fdc1e1ecdL266 to this function to be used in the noop - https://github.com/openshift/lightspeed-service/pull/2310/files#diff-3b24e1df8a702456e6bd68d43feab4a498270c7749577ab35a160cbf6b02fd36R40

joshuawilson: I know it fails when there is no auth header. I'm asking if anything else can make it fail, if so should we raise other exceptions that deal with that.

onmete: Don't see the case.

onmete:
closing in favor of https://github.com/openshift/lightspeed-service/pull/2441

PR #2389, opened 2025-03-31
load-test: temp fix for fake-llm output format

asamal4:
Temporary fix for fake-llm (load test) which gives output as string.
Currently every method that we use gives us proper output, except fake-llm.
We need to move to a different fake-llm (or custom fake-llm) which can handle streaming/non-streaming & tool calling and gives response not as string.
Even this temp fix will fail for tool calling. (load test can be run with tool calling set to False till we have a permanent fix)

[Discussion next-steps.](https://redhat-internal.slack.com/archives/C068JAU4Y0P/p1742968057080699?thread_ts=1742906110.101449&cid=C068JAU4Y0P)


tisnik:
/override "Red Hat Konflux / integration-tests / lightspeed-service"

tisnik:

PR #2393, opened 2025-04-01
OLS-1627: Fix PG issue

tisnik:
## Description

Fix PG issue
Conversion retrieved values from memoryview to string before deserialization

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

## Fixes:
#[ols-1627](https://issues.redhat.com//browse/ols-1627)

### DIFF
--- a/ols/src/cache/postgres_cache.py
+++ b/ols/src/cache/postgres_cache.py
@@ -131,12 +131,16 @@ class PostgresCache(Cache):
         Returns:
             The value associated with the key, or None if not found.
         """
+        # just check if user_id and conversation_id are UUIDs
+        super().construct_key(user_id, conversation_id, skip_user_id_check)
+
         with self.conn.cursor() as cursor:
             try:
                 value = PostgresCache._select(cursor, user_id, conversation_id)
                 if value is None:
                     return []
-                return [CacheEntry.from_dict(cache_entry) for cache_entry in value]
+                history = [CacheEntry.from_dict(cache_entry) for cache_entry in value]
+                return history
             except psycopg2.DatabaseError as e:
                 logger.error("PostgresCache.get %s", e)
                 raise CacheError("PostgresCache.get", e) from e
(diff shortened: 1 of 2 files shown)


ols/src/cache/postgres_cache.py
```diff
@@ -131,12 +131,16 @@ def get(
...
+        super().construct_key(user_id, conversation_id, skip_user_id_check)
+
         with self.conn.cursor() as cursor:
             try:
                 value = PostgresCache._select(cursor, user_id, conversation_id)
                 if value is None:
                     return []
-                return [CacheEntry.from_dict(cache_entry) for cache_entry in value]
+                history = [CacheEntry.from_dict(cache_entry) for cache_entry in value]
+                return history
```
onmete, line 143: It is easier to put debug points if it is not directly returned, right? ;)

tisnik: probably yes, but I'm going to add debug logs there (not in PR fix), so it will need be stored in localvar anyway

tests/e2e/xxx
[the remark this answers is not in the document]
tisnik: @asamal4 indeed, removed. TY

ols/src/cache/postgres_cache.py
tisnik: @onmete do you need to debug it? I can show/demo howto debug such things in running service (but my tooling might be a bit different from yours, dunno)

onmete: I prefer returning variables instead of doing computation as part of the return because it is easier to put debug lines to stay in the function and not checking the return. We had some conversation about this while ago, so I was just referring it here. :)

tisnik: ah ok, I thought to need to debug it, or help us with the release

tisnik:

PR #2399, opened 2025-04-02
change tool calling query

asamal4:
## Description
OpenAI tool calling test case started failing.
Surprisingly it works if we remove the `?` in query. GPT is not supposed to be so fragile though.
Perhaps there is a better solution, but for time being it fixes the issue.


tisnik:

tisnik:

tisnik:

PR #2411, opened 2025-04-07
eval: taxonomy - context & answer relevancy

asamal4:
## Description

- Context
  - Flag to check if context is valid as per the query.
  - Context relevancy score against the answer.
- Answer
  - Answer relevancy score against the query.
  - Answer ground-ness score against the context

Both custom & ragas implementation are done to check answer & context.
Script to automate the process with taxonomy file

[OLS-1368](https://issues.redhat.com//browse/OLS-1368)


tisnik (approved):
it looks ok

PR #2449, opened 2025-04-17
LCORE-22: Mcp in ols

onmete:
## Description

Mcp in ols

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -8,6 +8,7 @@ from langchain.globals import set_debug
 from langchain_core.messages import AIMessage, BaseMessage
 from langchain_core.messages.ai import AIMessageChunk
 from langchain_core.prompts import ChatPromptTemplate
+from langchain_mcp_adapters.client import MultiServerMCPClient
 from llama_index.core import VectorStoreIndex
 
 from ols import config
@@ -18,7 +19,7 @@ from ols.constants import MAX_ITERATIONS, RAG_CONTENT_LIMIT, GenericLLMParameter
 from ols.customize import reranker
 from ols.src.prompts.prompt_generator import GeneratePrompt
 from ols.src.query_helpers.query_helper import QueryHelper
-from ols.src.tools.tools import execute_oc_tool_calls, get_available_tools
+from ols.src.tools.tools import execute_tool_calls, get_available_tools
 from ols.utils.token_handler import TokenHandler
 
 logger = logging.getLogger(__name__)
@@ -63,6 +64,8 @@ class DocsSummarizer(QueryHelper):
         self._prepare_llm()
         self.verbose = config.ols_config.logging_config.app_log_level == logging.DEBUG
 
+        # TODO: remove introspection config and decide based if mcp_servers
+        # are provided
         # tools part
         self._introspection_enabled = config.ols_config.introspection_enabled
         self.tools_map = get_available_tools(self._introspection_enabled, user_token)
@@ -163,11 +166,29 @@ class DocsSummarizer(QueryHelper):
 
         return final_prompt, llm_input_values, rag_chunks, truncated
 
+    # TODO: better way
+    def mcp_config_to_client_dump(self):
+        """Unpacks config to MultiServerMCPClient expectations."""
+        servers_conf = {}
+        for server, server_conf in config.config.mcp_servers.servers.items():
+            servers_conf[server] = {
+                "transport": server_conf.transport,
+            }
+            if server_conf.stdio:
+                servers_conf[server].update(server_conf.stdio.model_dump())
+                # TODO: I'm not really happy about this solution, open
+                # for ideas
+                if server == "openshift":
+                    servers_conf[server]["env"] = {"OC_USER_TOKEN": self.user_token}
+            if server_conf.sse:
+                servers_conf[server].update(server_conf.sse.model_dump())
+        return servers_conf
+
     async def _invoke_llm(
         self,
         messages: ChatPromptTemplate,
         llm_input_values: dict[str, str],
-        tools_map: dict[str, Callable],
+        tools_map: list[Callable],
         is_final_round: bool,
         token_counter: GenericTokenCounter,
     ) -> AsyncGenerator[AIMessageChunk, None]:
@@ -187,7 +208,7 @@ class DocsSummarizer(QueryHelper):
         llm = (
             self.bare_llm
             if is_final_round or not tools_map
-            else self.bare_llm.bind_tools(tools_map.values())
+            else self.bare_llm.bind_tools(tools_map)
         )
 
         # create and execute the chain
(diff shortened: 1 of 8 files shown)


ols/src/query_helpers/docs_summarizer.py
```diff
@@ -218,68 +240,76 @@ async def iterate_with_tools(
...
-                # (load test can be run with tool calling set to False till we
-                # have a permanent fix)
-                if isinstance(chunk, str):
-                    yield StreamedChunk(type="text", text=chunk)
-                    break
-
-                # check if LLM has finished generating
-                if chunk.response_metadata.get("finish_reason") == "stop":  # type: ignore [attr-defined]
-                    return
+            async with asyncio.timeout(constants.TOOL_CALL_ROUND_TIMEOUT):
```
onmete, line 245: I've added this timeout because I forgot the
```python
if __name__ == "__main__":
    mcp.run(transport="stdio")
```
in the mcp/openshift tools and was scratching my head for a while why it is not doing anything and waiting forever.
But it is TBD. I see some value in limiting the rounds, but there is some line between beying stuck and executing many/long running tools.
I'm also not sure how that would work with human in the loop. So I'm fine with keeping this while it has value, or removing now.

mcp_local/openshift.py
```diff
@@ -1,95 +1,112 @@
...
-def token_works_for_oc(token: str) -> bool:
-    """Check if the token can be used with `oc` CLI.
+def is_secret_in_args(args: list[str]) -> list[str]:
+    """Check if any of the arguments contain 'secret' or 'secrets'."""
+    for arg in args:
+        if "secret" in arg or "secrets" in arg:
+            logger.error(
+                "'secret' or 'secrets' found in argument: %s",
+                arg,
+            )
```
asamal4, line 58:
```suggestion
    arg_str = " ".join(args)
    if "secret" in arg_str:
        logger.error("'secret' keyword found in argument: %s", arg_str)
```

tests/config/config_for_integration_tests.yaml
```diff
@@ -119,17 +119,6 @@ ols_config:
...
-  - name: foo
-    transport: stdio
-    stdio:
-      command: python
-      args:
-        - mcp_server_1.py
-  - name: bar
-    transport: sse
-    sse:
-      url: 127.0.0.1:8080
```
onmete, line 132: I've remove the mcp_servers from unit/integration configs. Those are not real servers so the mcp client hangs/timeouts trying to connect.

tests/config/operator_install/olsconfig.crd.azure_openai_introspection.yaml
```diff
@@ -34,3 +34,10 @@ spec:
       - name: bar_filter
         pattern: '\b(?:bar)\b'
         replaceWith: "openshift"
+mcp_servers:
```
JoaoFula, line 37: isn't this meant to be under spec? is the tabbing correct?

onmete: yes...

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -217,68 +284,81 @@ async def iterate_with_tools(
         Yields:
             StreamedChunk objects representing parts of the response
         """
+        mcp_servers = MCPConfigBuilder.mcp_config_to_client_dump(
+            config.mcp_servers.servers, self.user_token
+        )
```
asamal4, line 289: As we are still dealing with local mcp I feel like we can use customize to get config for the same..
We can discuss further on this.. but I think this approach is quick and still useful in future (we can still retain the logic later)
example
```
        mcp_servers = MCPConfigBuilder.mcp_config_to_client_dump(
            mcp from the config + mcp from customize , self.user_token
        )
```

asamal4:
cluster claim issue
/retest

ols/src/tools/tools.py
```diff
@@ -1,109 +1,59 @@
...
-
-    logger.info("User token provided; adding 'oc' tools")
-    return oc_tools
+def get_tool_by_name(tool_name: str, mcp_client: MultiServerMCPClient) -> Callable:
+    """Get a tool by its name from the MCP client."""
+    tool = [tool for tool in mcp_client.get_tools() if tool.name == tool_name]
+    if len(tool) == 0:
+        raise ValueError(f"Tool '{tool_name}' not found.")
+    if len(tool) > 1:
+        raise ValueError(f"Multiple tools found with name '{tool_name}'.")
```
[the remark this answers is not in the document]
onmete: yes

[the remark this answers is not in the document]
asamal4: Seems like an unlikely corner-case and yet a very fundamental problem, there should be some standard practise to solve this issue.
At least we shouldn't fail, because we won't be controlling the mcp server always.
We can do a workaround, but I think that can wait (we just have one mcp right now). I will create a follow up story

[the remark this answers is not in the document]
onmete: I've seen some issues in `langchain-mcp-adapters` itself about this scenario too. It is still early technology. That is why I put the check in. Until there is a better solution.

tests/config/operator_install/olsconfig.crd.openai_introspection.yaml
```diff
@@ -32,3 +32,10 @@ spec:
       - name: bar_filter
         pattern: '\b(?:bar)\b'
         replaceWith: "openshift"
+  mcp_servers:
+    - name: openshift
+      transport: stdio
+      stdio:
+        command: python
+        args:
+          - /app-root/mcp_local/openshift.py
```
asamal4, line 41: Please use relative path
```suggestion
  mcp_servers:
    - name: openshift
      transport: stdio
      stdio:
        command: python
        args:
          - mcp_local/openshift.py
```

tests/config/operator_install/olsconfig.crd.azure_openai_introspection.yaml
```diff
@@ -34,3 +34,10 @@ spec:
       - name: bar_filter
         pattern: '\b(?:bar)\b'
         replaceWith: "openshift"
+  mcp_servers:
+    - name: openshift
+      transport: stdio
+      stdio:
+        command: python
+        args:
+          - /app-root/mcp_local/openshift.py
```
asamal4, line 43: Please use relative path
```suggestion
          - mcp_local/openshift.py
```

onmete: Once I get the CI working, I'll test if it works. There might be some changes required given the place where we execute the code.

asamal4: It shouldn't be different than local `make run`. But okay lets wait for CI.
meanwhile if you want you can try temporarily hardcoding in the code.

Please let @xrajesh know about this.. as this will be added in operator.

asamal4:
@Ondrej I don't see containerfile change to put mcp_local..
Currently we only send ols/ and couple of other files..
https://github.com/openshift/lightspeed-service/blob/main/Containerfile#L57

This is the reason why it's gets stuck in CI. there is also error message in the log..
```
2025-04-30 07:28:12,645 [ols.src.query_helpers.docs_summarizer:docs_summarizer.py:290] INFO: Using mcp servers: ['openshift'] for tool calling
python3.11: can't open file '/app-root/mcp_local/openshift.py': [Errno 2] No such file or directory
```

ols/src/tools/tools.py
```diff
@@ -1,103 +1,69 @@
...
-
-
-# NOTE: Limit for a tool description is 1024 characters
-for tool in oc_tools.values():
-    check_tool_description_length(tool)
-
-
-def execute_oc_tool_calls(
-    tools_map: dict,
+def get_tool_by_name(tool_name: str, mcp_client: MultiServerMCPClient) -> Callable:
```
iNecas, line 12: Given you're calling `arun` later on the result of this method, the signature should probably be `-> BaseTool`?

onmete: It is actually `StructuredTool`. Updated.

tisnik (approved):
approved

PR #2450, opened 2025-04-18
OLS-1176: add option to configure CA certificate of HTTPS proxy

raptorsun:
## Description

Add a new option to configure CA certificate of HTTPS proxy. This allows the lightspeed backend to connect to external server ,such as LLM provider, through a TLS introspection HTTPS proxy, not only through a transparent proxy. 

This PR does not need synchronize with the operator side.
The operator side PR: https://github.com/openshift/lightspeed-operator/pull/702

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue # [OLS-1176](https://issues.redhat.com//browse/OLS-1176)
- Closes #

## Testing

### DIFF
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -9,6 +9,7 @@ from pydantic import (
     AnyHttpUrl,
     BaseModel,
     DirectoryPath,
+    Field,
     FilePath,
     PositiveInt,
     field_validator,
@@ -118,6 +119,44 @@ class TLSConfig(BaseModel):
                     "Can not enable TLS without ols_config.tls_config.tls_key_path"
                 )
             checks.file_check(self.tls_key_path, "OLS server certificate private key")
+
+
+class ProxyConfig(BaseModel):
+    """HTTPS Proxy configuration."""
+
+    proxy_url: Optional[AnyHttpUrl] = Field(
+        default_factory=lambda: os.getenv("https_proxy") or os.getenv("HTTPS_PROXY")
+    )  # type: ignore
+    proxy_ca_cert_path: Optional[FilePath] = None
+
+    def __init__(self, data: Optional[dict] = None) -> None:
+        """Initialize configuration and perform basic validation."""
+        super().__init__()
+        if not data:
+            return
+        self.proxy_url = data.get("proxy_url")
+        self.proxy_ca_cert_path = data.get("proxy_ca_cert_path")
+
+    def validate_yaml(self) -> None:
+        """Validate proxy config."""
+        if self.proxy_url is None and self.proxy_ca_cert_path:
+            raise checks.InvalidConfigurationError("Proxy URL is missing")
+        if self.proxy_url and not checks.is_valid_http_url(self.proxy_url):
+            raise checks.InvalidConfigurationError(
+                "Proxy URL is invalid, only http:// and https:// URLs are supported"
+            )
+        if self.proxy_ca_cert_path is not None:
+            checks.file_check(self.proxy_ca_cert_path, "Proxy CA certificate")
+
+    @model_validator(mode="before")
+    @classmethod
+    def set_default_proxy_url(cls, values: Any) -> Any:
+        """Set proxy URL from environment variable if not provided."""
+        if "proxy_url" not in values or values["proxy_url"] is None:
+            env_proxy = os.getenv("https_proxy") or os.getenv("HTTPS_PROXY")
+            if env_proxy:
+                values["proxy_url"] = env_proxy
+        return values
 
 
 class AuthenticationConfig(BaseModel):
@@ -1043,6 +1082,8 @@ class OLSConfig(BaseModel):
 
     quota_handlers: Optional[QuotaHandlersConfig] = None
 
+    proxy_config: Optional[ProxyConfig] = None
+
     def __init__(
         self, data: Optional[dict] = None, ignore_missing_certs: bool = False
     ) -> None:
@@ -1095,6 +1136,7 @@ class OLSConfig(BaseModel):
             data.get("tlsSecurityProfile", None)
         )
         self.quota_handlers = QuotaHandlersConfig(data.get("quota_handlers", None))
+        self.proxy_config = ProxyConfig(data.get("proxy_config"))
 
     def __eq__(self, other: object) -> bool:
         """Compare two objects for equality."""
@@ -1118,6 +1160,7 @@ class OLSConfig(BaseModel):
                 and self.expire_llm_is_ready_persistent_state
                 == other.expire_llm_is_ready_persistent_state
                 and self.quota_handlers == other.quota_handlers
+                and self.proxy_config == other.proxy_config
             )
         return False
 
@@ -1136,6 +1179,8 @@ class OLSConfig(BaseModel):
             self.tls_security_profile.validate_yaml()
         if self.authentication_config is not None:
             self.authentication_config.validate_yaml()
+        if self.proxy_config is not None:
+            self.proxy_config.validate_yaml()
 
         valid_query_validation_methods = list(constants.QueryValidationMethod)
         if self.query_validation_method not in valid_query_validation_methods:
--- a/ols/src/llms/providers/provider.py
+++ b/ols/src/llms/providers/provider.py
@@ -336,6 +336,22 @@ class LLMProvider(AbstractLLMProvider):
         self, use_custom_certificate_store: bool, use_async: bool
     ) -> httpx.Client | httpx.AsyncClient:
         """Construct HTTPX client instance to be used to communicate with LLM."""
+        # no proxy by default
+        proxy = None
+        # set up proxy if configured.
+        if config.ols_config.proxy_config and config.ols_config.proxy_config.proxy_url:
+            logger.debug(
+                "Proxy is configured. Proxy URL: %s Proxy CA cert: %s",
+                config.ols_config.proxy_config.proxy_url,
+                config.ols_config.proxy_config.proxy_ca_cert_path,
+            )
+            proxy_context = ssl.create_default_context(
+                cafile=config.ols_config.proxy_config.proxy_ca_cert_path
+            )
+            proxy = httpx.Proxy(
+                url=config.ols_config.proxy_config.proxy_url, ssl_context=proxy_context
+            )
+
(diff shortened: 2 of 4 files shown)


ols/app/models/config.py
```diff
@@ -121,6 +121,42 @@ def validate_yaml(self, disable_tls: bool = False) -> None:
...
+        """Validate proxy config."""
+        if self.proxy_url is None and self.proxy_ca_cert_path:
+            raise checks.InvalidConfigurationError("Proxy URL is missing")
+        if self.proxy_url and not checks.is_valid_http_url(self.proxy_url):
+            raise checks.InvalidConfigurationError(
+                "Proxy URL is invalid, only http:// and https:// URLs are supported"
+            )
+        if self.proxy_ca_cert_path is not None:
+            checks.file_check(self.proxy_ca_cert_path, "Proxy CA certificate")
+
```
onmete, line 158: Can you try something like this?

```python
class ProxyConfig(BaseModel):
    """HTTPS Proxy configuration."""

    proxy_url: Optional[AnyHttpUrl] = os.environ.get("https_proxy")
    proxy_ca_cert_path: Optional[FilePath] = None

# no additional code
```

ols/app/models/config.py
```diff
@@ -121,6 +121,42 @@ def validate_yaml(self, disable_tls: bool = False) -> None:
...
+            return
+        self.proxy_url = data.get("proxy_url", None)
+        if self.proxy_url is None:
+            # check if HTTPS_PROXY is set in the environment
+            # and use it as a proxy URL
+            os_proxy = os.environ.get("HTTPS_PROXY")
+            if os_proxy:
+                self.proxy_url = os_proxy
+            else:
+                os_proxy = os.environ.get("https_proxy")
```
onmete, line 143: Is there a reason not to follow the industry standard and check for a lowercase env variable? Is there env where user have no controll about this and it is set as lowercase?

tests/unit/app/models/test_config.py
```diff
@@ -4117,3 +4119,54 @@ def test_ols_config_with_quota_handlesr_missing_name():
...
         )
+
+
+def test_proxy_config_default_values():
+    """Test the ProxyConfig model default values."""
+    proxy_config = ProxyConfig()
+    os_proxy = os.getenv("https_proxy")
+    if os_proxy is None:
+        os_proxy = os.getenv("HTTPS_PROXY")
+    assert proxy_config.proxy_url == os_proxy
```
onmete, line 4130: Can you explicitly test the cases when the env is missing and contains a value?

ols/app/models/config.py
raptorsun: good point, removing uppercase env var.

ols/app/models/config.py
raptorsun: I keep the init function, just find it very difficult to code the unit test. to make this work, the env var must be set before importing modules.

tests/unit/app/models/test_config.py
raptorsun: new test case added for assignment by env var

raptorsun:
rebased on main

raptorsun:
@onmete @tisnik all tests passed, ready to review again :)

ols/app/models/config.py
```diff
@@ -120,6 +120,44 @@ def validate_yaml(self, disable_tls: bool = False) -> None:
...
+
+    proxy_url: Optional[AnyHttpUrl] = None
+    proxy_ca_cert_path: Optional[FilePath] = None
+
+    def __init__(self, data: Optional[dict] = None) -> None:
+        """Initialize configuration and perform basic validation."""
+        super().__init__()
+        if not data:
+            return
+        self.proxy_url = data.get("proxy_url")  # type: ignore
```
onmete, line 134: Is `# type: ignore` really needed?

ols/app/models/config.py
```diff
@@ -120,6 +120,44 @@ def validate_yaml(self, disable_tls: bool = False) -> None:
...
+        self.proxy_ca_cert_path = data.get("proxy_ca_cert_path", None)
```
onmete, line 135: `None` is redundant

ols/app/models/config.py
```diff
@@ -120,6 +120,44 @@ def validate_yaml(self, disable_tls: bool = False) -> None:
...
+            raise checks.InvalidConfigurationError(
+                "Proxy URL is invalid, only http:// and https:// URLs are supported"
+            )
+        if self.proxy_ca_cert_path is not None:
+            checks.file_check(self.proxy_ca_cert_path, "Proxy CA certificate")
+
+    @classmethod
+    def from_dict(cls, data: Optional[dict] = None) -> Self | None:
+        """Create a ProxyConfig instance from a dictionary."""
+        os_proxy_url = os.environ.get("https_proxy")
```
onmete, line 151: I see there is some exception to have the env `https_proxy` in lowercase. The lowercase have precedense over uppercase. Would it work if we define the env name in standard uppercase format?

ols/app/models/config.py
onmete: You can reimport in test or try this lazy way.
```python
from pydantic import Field
class ProxyConfig(BaseModel):
    """HTTPS Proxy configuration."""

    proxy_url: Optional[AnyHttpUrl] = Field(default_factory=lambda: os.getenv("https_proxy"))
    proxy_ca_cert_path: Optional[FilePath] = None
```
That would also enable the checks for URL and file on pydantic side - the `validate_yaml` can be removed then.

ols/src/llms/providers/provider.py
```diff
@@ -353,8 +369,8 @@ def _construct_httpx_client(
                 )
                 verify = custom_context
             if use_async:
-                return httpx.AsyncClient(verify=verify)
-            return httpx.Client(verify=verify)
+                return httpx.AsyncClient(verify=verify, proxies=proxy)
+            return httpx.Client(verify=verify, proxies=proxy)
```
onmete, line 373: I assume it configures proxy for https, not http, right?

ols/app/models/config.py
raptorsun: this implementation works in lower case only, will add a upper case.

ols/app/models/config.py
raptorsun: how to validate the model after removing `validate_yaml` ? 
after removing the validator, the malformatted URL does not raise exceptions and return an object as normal
```
    ProxyConfig(
        {
            "proxy_url": "invalid-url",
            "proxy_ca_cert_path": "tests/config/empty_cert.crt",
        }
    )
```

ols/src/llms/providers/provider.py
raptorsun: yes, it is for HTTPS proxy. 
any modification required on these 2 lines?

ols/app/models/config.py
raptorsun: The default_factory works, i'm removing the `from_dict` method

raptorsun: I remove the env var handling in from_dict except the return None object when URL is neither in env var nor in the data dictionary.

onmete: Pydantic is doing the validation according to the field types. See this snippet
```python
from pydantic import BaseModel, AnyHttpUrl

class ProxyConfig(BaseModel):
    """HTTPS Proxy configuration."""

    proxy_url: AnyHttpUrl


# pass
ProxyConfig(proxy_url="https://example.com")

# raise Input should be a valid URL, relative URL without a base 
ProxyConfig(proxy_url="invalid-url")
```

ols/app/models/config.py
raptorsun: yes. elsewise the type check will not pass 
```
pdm run mypy --explicit-package-bases --disallow-untyped-calls --disallow-untyped-defs --disallow-incomplete-defs ols/
INFO: Inside an active virtualenv /opt/app-root, reusing it.
Set env var PDM_IGNORE_ACTIVE_VENV to ignore it.
/opt/app-root/lib64/python3.11/site-packages/requests/__init__.py:86: RequestsDependencyWarning: Unable to find acceptable character detection dependency (chardet or charset_normalizer).
  warnings.warn(
ols/app/models/config.py:127: error: Incompatible types in assignment (expression has type "str | None", variable has type "AnyHttpUrl | None")  [assignment]
```

ols/app/models/config.py
raptorsun: I think I still need the validate_yaml. Testing with the constructor with empty arguments `pc = ProxyConfig()`, it initializes an object with a None proxy_url that is not a valid `AnyHttpUrl`. 
I removed the constructor and the Optional type hint for `proxy_url`. 
```
class ProxyConfig(BaseModel):
    """HTTPS Proxy configuration."""

    proxy_url: AnyHttpUrl = Field(
        default_factory=lambda: os.getenv("https_proxy") or os.getenv("HTTPS_PROXY")
    )  # type: ignore
    proxy_ca_cert_path: Optional[FilePath] = None
```

tisnik (approved):
approved

ols/app/models/config.py
onmete: One last chance, try:
```python
class ProxyConfig(BaseModel):
    proxy_url: Optional[AnyHttpUrl] = None
    proxy_ca_cert_path: Optional[FilePath] = None

    @model_validator(mode='before')
    @classmethod
    def set_default_proxy_url(cls, values: Any) -> Any:
        if 'proxy_url' not in values or values['proxy_url'] is None:
            env_proxy = os.getenv("https_proxy") or os.getenv("HTTPS_PROXY")
            if env_proxy:
                values['proxy_url'] = env_proxy
        return values
```

If you do `ProxyConfig()` it creates an instance with `None` as values (if no env is set). Which seems valid, you should then check `if proxy_config.proxy_url` when using the value.

Sorry for the delays, I'm just trying to see if it is possible to not add any more mess to the configs - not your fault, it is wrong/misused from the beginning. :)

onmete:
/lgtm

I think the init and validate_yaml are redundant though.

PR #2457, opened 2025-04-23
fix: set top-k chunks correctly

asamal4:
## Description

We need to set top-k chunks in both the places, otherwise default of 2 is considered.
Without this we will retrieve only 2 chunks, even if we have set to 5 in the 2nd part.
So the flow is like 
- use top k for individual retrieval (this will pick k chunks from each index)
- Use same/or different top k for final query fusion retrieval (this will give us final set of chunks from `n index * k chunks`)..

While using 1 index, first index only retrieves default 2 chunks and that's the reason we don't get 5 chunks in final retrieval

### DIFF
--- a/ols/src/rag_index/index_loader.py
+++ b/ols/src/rag_index/index_loader.py
@@ -147,7 +147,10 @@ class IndexLoader:
         ):
             return self._retriever
         retriever = QueryFusionRetriever(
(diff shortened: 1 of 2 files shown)


tisnik (approved):
it looks ok

raptorsun:
/lgtm
thanks, I didnt realise it comes from this argument :D

PR #2478, opened 2025-04-30
Tools improvements

onmete:
- More robust tool execution.
- Redaction of token.
- Small changes in tools (renaming).

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/ols/src/tools/oc_cli.py
+++ b/ols/src/tools/oc_cli.py
@@ -33,10 +33,7 @@ SECRET_NOT_ALLOWED_MSG = (
 def strip_args_for_oc_command(args: list[str]) -> list[str]:
     """Sanitize arguments for `oc` CLI commands if LLM adds it extra."""
     # sometimes within the list we may get two args combined, eg: [top pod]
-    splited_args = " ".join(args).split(" ")
-
-    # strip whitespaces from args
-    splited_args = [arg.strip() for arg in splited_args if arg.strip()]
+    splited_args = [arg for arg in " ".join(args).split(" ") if arg]
 
     # sometimes model gives args which are already added to the tool.
     remove_arg = ["oc", "get", "describe", "logs", "status", "adm", "top"]
--- a/ols/src/tools/oc_cli.py
+++ b/ols/src/tools/oc_cli.py
@@ -12,6 +12,7 @@
 
 import logging
 import subprocess
+import traceback
 from typing import Annotated
 
 from langchain.tools import tool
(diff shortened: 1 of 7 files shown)


ols/src/tools/oc_cli.py
```diff
@@ -61,38 +66,52 @@ def is_secret_in_args(args: list[str]) -> list[str]:
...
 
 
-def stdout_or_stderr(result: subprocess.CompletedProcess) -> str:
+def resolve_status_and_response(result: subprocess.CompletedProcess) -> tuple[str, str]:
     """Return stdout if it is not empty string, otherwise return stderr."""
     # some commands returns empty stdout and message like "namespace not found"
     # in stderr, but with return code 0
-    return result.stdout if result.stdout != "" else result.stderr
+    if result.returncode == 0:
+        return "success", result.stdout if result.stdout != "" else result.stderr
```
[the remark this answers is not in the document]
onmete: As in the comment right above this line.
Or when doing `oc get pods -n xyz`, you get `no resources in namespace xyz`, which returns 0 but text is in stderr.
image

ols/src/tools/oc_cli.py
```diff
@@ -31,16 +32,20 @@
...
 def strip_args_for_oc_command(args: list[str]) -> list[str]:
     """Sanitize arguments for `oc` CLI commands if LLM adds it extra."""
-    # Sometimes within the list we may get two args combined; ex: [top pod]
-    striped_args = " ".join(args).split(" ")
-    # Sometimes model gives args which are already added to the tool.
+    # sometimes within the list we may get two args combined, eg: [top pod]
+    splited_args = " ".join(args).split(" ")
+
+    # strip whitespaces from args
+    splited_args = [arg.strip() for arg in splited_args if arg.strip()]
```
asamal4, line 39: FYI -> default `strip` removes any whitespace character at the beginning/end. 
`string.whitespace -> ' \t\n\r\x0b\x0c'`. Just making sure that this is what you intend to do.

I have seen newline character but in the middle of the arg, so may be there is no valid arg with whitespace characters at the beginning/end.

ols/src/tools/oc_cli.py
```diff
@@ -61,38 +66,52 @@ def is_secret_in_args(args: list[str]) -> list[str]:
...
+        return "error", SECRET_NOT_ALLOWED_MSG
+    status, result = run_oc([*commands, *strip_args_for_oc_command(args)], token)
+    return status, result
 
 
 @tool
-def oc_get(oc_get_args: list[str], token: Annotated[str, InjectedToolArg]) -> str:
+def oc_get(
+    oc_get_args: list[str], token: Annotated[str, InjectedToolArg]
+) -> tuple[str, str]:
```
asamal4, line 114: Wondering how internally langchain send `status` to the model ? everything is string for the model, but does it get information about two returned values ? Because model sees function definition and expects two returned values.

Anyways so far it doesn't seem to have much impact.. we can keep this way, until we see a problem

asamal4 (commented):
LGTM. just one clarification.

ols/src/tools/oc_cli.py
onmete: The intention is more for the scenario of `" a".strip()`

asamal4: in that case you don't need additional strip(), earlier split(" ") already has removed additional whitespace..

asamal4, line 39: Based on [this](https://github.com/openshift/lightspeed-service/pull/2478/files#r2072737541) we can modify below logic
```suggestion
    # sometimes within the list we may get two args combined, eg: [top pod]
    splited_args = [arg for arg in " ".join(args).split(" ") if arg]
```

PR #2512, opened 2025-05-19
tool calling: fixes & improvements

asamal4:
## Description

- Enable langchain debug (now that we are handling tokens better)
- Create MCP client connection per user request rather than per iteration.
- Raise exception when we skip OC command execution due to some validation. 
- Raise exception when OC command fails, so that status in host can be error.
- handle <tool_call> text (for granite)

### DIFF
--- a/mcp_local/openshift.py
+++ b/mcp_local/openshift.py
@@ -34,36 +34,22 @@ def strip_args_for_oc_command(args: list[str]) -> list[str]:
     return split_args
 
 
-def is_blocked_char_in_args(args: list[str]) -> bool:
-    """Check if any of the arguments contain blocked characters."""
+def raise_for_unacceptable_args(args: list[str]) -> None:
+    """Check & raise exception if arguments are unacceptable."""
     arg_str = " ".join(args)
+
     if any(char in arg_str for char in BLOCKED_CHARS):
         logger.error("Blocked characters found in argument: %s", arg_str)
-        return True
-    return False
+        raise Exception(BLOCKED_CHARS_DETECTED_MSG)
 
-
-def is_secret_in_args(args: list[str]) -> bool:
-    """Check if any of the arguments contain 'secret'."""
-    arg_str = " ".join(args)
     if "secret" in arg_str:
         logger.error("'secret' keyword found in argument: %s", arg_str)
-        return True
-    return False
+        raise Exception(SECRET_NOT_ALLOWED_MSG)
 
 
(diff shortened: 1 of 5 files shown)


ols/src/query_helpers/docs_summarizer.py
```diff
@@ -311,9 +310,12 @@ async def iterate_with_tools(
         Yields:
             StreamedChunk objects representing parts of the response
         """
-        for i in range(1, max_rounds + 1):
-            async with asyncio.timeout(constants.TOOL_CALL_ROUND_TIMEOUT):
-                async with MultiServerMCPClient(self.mcp_servers) as mcp_client:
+        async with asyncio.timeout(constants.TOOL_CALL_ROUND_TIMEOUT):
```
onmete, line 313: Timeout is then applied for the full conversation, not just a single round. Given the current timeout value, we need to change it otherwise the risk of timeouting in the middle of the round.

mcp_local/openshift.py
```diff
@@ -89,12 +87,10 @@ def run_oc(args: list[str]) -> str:
...
 def safe_run_oc(commands: list[str], args: list[str]) -> str:
     """Run `oc` CLI with provided arguments and command."""
-    if is_blocked_char_in_args(args):
-        return BLOCKED_CHARS_DETECTED_MSG
-    if is_secret_in_args(args):
-        return SECRET_NOT_ALLOWED_MSG
-    result = run_oc([*commands, *strip_args_for_oc_command(args)])
-    return result
+    is_blocked_char_in_args(args)
+    is_secret_in_args(args)
```
onmete, line 91: Rename to `raise_for_blocked_char_in_args` and `raise_for_secret_in_args` then.

ols/src/query_helpers/docs_summarizer.py
asamal4: modified timeout to `constants.TOOL_CALL_ROUND_TIMEOUT * max_rounds`

mcp_local/openshift.py
asamal4: I was doing bit of refactoring also, modified to `raise_for_unacceptable_args` (covers both scenarios). PTAL

asamal4:
/assign @onmete 
test cases are failing for unrelated error. meanwhile you can have a look (added new commit for granite fix).

asamal4:
/hold
for other priority PR

asamal4:
/unhold
release window passed

PR #2536, opened 2025-05-29
OLS-1800: Add auth header for sse mcp config

onmete:
## Description

Implements user-token auth through header.

**How to test locally**
0. get the cluster, do `oc login` and store the kubeconfig `oc config view --raw > <kubeconfig_path>`
1. start the mcp server via `npx kubernetes-mcp-server@latest --sse-port 8090 --kubeconfig <kubeconfig_path>`
2. start ols service with this config
```yaml
ols_config:
       authentication_config:
    module: "noop-with-token"
    ...
mcp_servers:
  - name: k8s-mcp
    transport: sse
    sse:
      url: "http://localhost:8090/sse"
```
3. prompt through
```bash
curl -X POST http://localhost:8080/v1/query -H "Content-Type: application/json" -H "Authorization: Bearer sha256~blabla" -d '{"query": "Is there openshift-lightspeed namespace in my openshift 4 cluster?"}'
```

### DIFF
--- a/ols/src/tools/mcp_config_builder.py
+++ b/ols/src/tools/mcp_config_builder.py
@@ -1,0 +1,85 @@
+"""MCPConfigBuilder for building MultiServerMCPClient configuration."""
+
+import logging
+import os
+from typing import Any
+
+from ols.app.models.config import MCPServerConfig
+
+logger = logging.getLogger(__name__)
+
+
+# Additional header containing user token for k8s/ocp authentication
+# for SSE MCP servers.
+K8S_AUTH_HEADER = "kubernetes-authorization-bearer-token"
+
+
+class MCPConfigBuilder:
+    """Builds MCP config for MultiServerMCPClient."""
+
+    def __init__(
+        self, user_token: str, mcp_server_configs: list[MCPServerConfig]
+    ) -> None:
+        """Initialize the MCPConfigBuilder with user token and server config list."""
+        self.user_token = user_token
+        self.mcp_server_configs = mcp_server_configs
+
+    @staticmethod
+    def include_auth_header(user_token: str, config: dict[str, Any]) -> dict[str, Any]:
+        """Include user token in the config headers."""
+        if "headers" not in config:
+            config["headers"] = {}
+        if K8S_AUTH_HEADER in config["headers"]:
+            logger.warning(
+                "Kubernetes auth header is already set, overriding with actual user token."
+            )
+        config["headers"][K8S_AUTH_HEADER] = user_token
+        return config
+
+    def include_auth_to_stdio(self, server_envs: dict[str, str]) -> dict[str, str]:
+        """Resolve OpenShift stdio env config."""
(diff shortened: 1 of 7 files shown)


ols/src/tools/mcp_config_builder.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import Any
+
+from ols.app.models.config import MCPServerConfig
+
+logger = logging.getLogger(__name__)
+
+
+# Additional header containing user token for k8s/ocp authentication
+# for SSE MCP servers.
+K8S_AUTH_HEADER = "Kubernetes-Authorization-Bearer-Token"
```
onmete, line 14: @tremes @manusa Please confirm the header. I'm fine with this, but I don't know if there is special convention we need to follow here.

The example config we then pass to our MCP client is as
```python
'manusa':
   {
      'transport': 'sse',
      'url': 'http://localhost:8080/sse',
      'timeout': 5,
      'sse_read_timeout': 10,
      'headers': {
         'Kubernetes-Authorization-Bearer-Token': 'sha256~t....Y'   <-----
      }
}
```

I'll then polish the PR and add the tests.

manusa: I made up the header, no convention used.

The reasoning for the naming (from the MCP Server side of things) is that this is related to the Kubernetes domain of the server, it's related to the Authorization scope, and is used to replace the BearerToken field in the rest Configuration.

I'm fine changing it to something else.
Also probably make it lower-case to make it HTTP/2 compliant.

(Related PR https://github.com/fabric8io/kubernetes-client/pull/7111)

falox: @onmete @tremes Why not mimic the standard `Authorization` header and move the schema into the value? 

```
'kubernetes-authorization': 'Bearer sha256~t....Y'
```

This would shorten the name and allow for future enhancements.

cc: @manusa also for your https://github.com/manusa/kubernetes-mcp-server/pull/96

manusa: > @onmete @tremes Why not mimic the standard Authorization header and move the schema into the value?

This makes sense in the context that the header is used only to override standard HTTP auth headers.

However, other options and future enhancements could include providing [other configuration fields](https://github.com/manusa/kubernetes-mcp-server/blob/9b517585fbab45a19d250e323e8fc4049a4d7339/pkg/kubernetes/kubernetes.go#L132-L142) such as:
- host ---> `kubernetes-host`
- Username/Password ----> `kubernetes-authorization-username` / `kubernetes-authorization-password` (which could also be expressed by using `kuberentes-authorization: Basic <credentials>` :shrug: 
- insecure ---> `kubernetes-insecure`
- ...

So I'm OK to change it unless you see any possible issue when considering the previous scenario too.

asamal4:
I would have preferred to have `kubernetes-authorization` name also configurable (instead of hardcoding).
But I understand that this might be early stage optimization considering we don't what all other MCPs will be used.
looks good as of now..
/lgtm

PR #2554, opened 2025-06-13
granite tool call prompt update

asamal4:


ols/customize/ols/prompts.py
```diff
@@ -33,11 +33,13 @@
...
-When you request/produce function calls, add `<tool_call>` at the beginning, so that function calls can be identified.
-  Sample tool format: '<tool_call>[{{"arguments": {{"oc_adm_top_args": ["pods", "-A"]}}, "name": "oc_adm_top"}}]'
-Do not use function calls for below kind of queries (These kind of queries do not require real time cluster data):
+You have been also given set of tools.
+Your task is to decide if tool call is needed and produce a json list of tools required to generate response to the user utterance.
+When you request/produce tool call, add `<tool_call>` at the beginning, so that tool call can be identified.
+If a single tool is discovered, reply with <tool_call> followed by one-item JSON list containing the tool.
+Tool call must be a json list like below example.
+  Sample tool call format: '<tool_call>[{{"arguments": {{"oc_adm_top_args": ["pods", "-A"]}}, "name": "oc_adm_top"}}]'
+Do not use tool call for below kind of queries (These kind of queries do not require real time data):
```
xrajesh, line 42: Do not use tool call for below kind of queries -> Do not use tool call for the following kind of queries 
Is that better @asamal4 ?

PR #2617, opened 2025-09-24
add example for ls-eval

asamal4:
## Description

- Add default system.yaml as per OLS
- Add OLS eval data as per LS-Eval data format
- Basic instruction to run evaluation


asamal4:
@xrajesh PTAL.. no impact to any of the functionality.

PR #2669, opened 2025-11-20
Fix GradioUI system prompt

Akrog:
## Description

On the GradioUI the system prompt displayed is *always* using the default system prompt, which is problematic when running the service with a custom system prompt (`system_prompt_path`) and allowing overriding it (`enable_system_prompt_override: true`).

With this simple patch we default to the configured system prompt on the GradioUI and only use the default system prompt if we don't have it configured.

## Type of change

- [x] Bug fix

## Testing


onmete:
No response, merging the other PR and closing this one.

/close

PR #2874, opened 2026-04-01
Introduce TokenBudgetTracker for unified per-request token accounting

blublinsky:
## Description

**Motivation**

The previous approach relied on a single available_tokens integer threaded through method calls, with no breakdown of what consumed the budget. Tool-calling rounds charged tokens via a mutable ToolTokenUsage dataclass passed between methods. AI messages emitted between rounds were not tracked at all. This made it impossible to diagnose context window exhaustion or reason about remaining budget.

**Summary**

Replace the ad-hoc token budget management scattered across DocsSummarizer with a centralized TokenBudgetTracker that tracks every token allocation by category (prompt, history, RAG, skill, tool definitions, AI rounds, tool results). This gives full visibility into where the context window is spent and prevents unaccounted token usage — notably AI messages between tool-calling rounds that were previously invisible.

**What changed**

- ols/utils/token_handler.py — Added TokenBudgetTracker class with per-category charging (TokenCategory enum), computed properties for prompt_budget, history_budget, tools_round_budget, and remaining, plus a summary() logging method. Moved format_retrieved_chunk here (was only used by TokenHandler). Simplified truncate_rag_context to return only the chunk list. Relocated calculate_and_check_available_tokens to the end of TokenHandler (unchanged logic).
- ols/src/query_helpers/docs_summarizer.py — DocsSummarizer now creates a TokenBudgetTracker in __init__ and charges tokens explicitly at each stage: PROMPT after initial prompt sizing, RAG after truncation, SKILL after loading, HISTORY after preparation. Skill resolution moved before history, so the history budget reflects skill overhead. Removed ToolTokenUsage dataclass and the token_handler/max_tokens_for_tools parameters threaded through _build_final_prompt and _process_tool_calls_for_round. Tool rounds charge AI_ROUND and TOOL_RESULT categories and log a per-round summary. DocsSummarizer drops from 1115 to 1087 lines.
- scripts/evaluation/utils/rag.py — Updated truncate_rag_context call to match new single-return signature.
- Tests — Updated unit and integration tests to patch TokenBudgetTracker properties instead of ToolTokenUsage, removed obsolete assertions on available_tokens returns, added new tests for tracker properties and budget computation.

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
https://redhat.atlassian.net/browse/OLS-2875
- Closes #
https://redhat.atlassian.net/browse/OLS-2875

## Testing

### DIFF
--- a/ols/src/tools/tools.py
+++ b/ols/src/tools/tools.py
@@ -30,7 +30,7 @@ RATE_LIMIT_RETRY_BACKOFF_SECONDS = 1.0
 DO_NOT_RETRY_REMINDER = "Do not retry this exact tool call."
 
 _TRUNCATION_WARNING = (
-    "\n\n[OUTPUT TRUNCATED - The tool returned more data than can be "
+    "\n[OUTPUT TRUNCATED - The tool returned more data than can be "
     "processed. Please ask a more specific question to get complete results.]"
 )
 _TRUNCATION_WARNING_TOKENS = TokenHandler._get_token_count(
@@ -98,6 +98,18 @@ def _is_rate_limited_tool_error(error: Exception) -> bool:
 _CHARS_PER_TOKEN_ESTIMATE = 4
 
 
+def _truncate_text_to_char_limit(text: str, max_chars: int) -> str:
+    """Return up to ``max_chars`` characters, cutting at the last newline when possible."""
+    if max_chars <= 0:
+        return ""
+    if len(text) <= max_chars:
+        return text
+    prefix = text[:max_chars]
+    cut = prefix.rfind("\n")
+    chunk = prefix[:cut].rstrip("\r") if cut > 0 else prefix
+    return chunk.strip()
+
+
 def _extract_text_from_tool_output(
     output: Any, tools_token_budget: int
 ) -> tuple[str, bool]:
@@ -111,8 +123,10 @@ def _extract_text_from_tool_output(
     response size at the transport layer, so a cheap character-level limit
     (tools_token_budget * 4) is applied here before any tokenization to
     avoid the CPU cost of tokenizing arbitrarily large tool responses.
-    Strings are cut at the last newline boundary; lists are truncated by
-    dropping trailing blocks that would exceed the limit.
+    Strings are cut at the last newline boundary. For lists, blocks are
+    concatenated until the limit; the first block alone may be shortened
+    to the limit before appending the truncation notice; later blocks are
+    dropped if they would exceed the limit.
 
     Args:
         output: Tool output, either a string or list of content blocks.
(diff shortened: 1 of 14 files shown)


ols/utils/token_handler.py
```diff
@@ -231,3 +200,153 @@ def limit_conversation_history(
...
+            - self._usage[TokenCategory.SKILL]
+        )
+
+    def charge(self, category: TokenCategory, tokens: int) -> None:
+        """Add a token charge to the specified category.
+
+        Args:
+            category: Which budget category to charge.
+            tokens: Number of tokens to add.
+        """
```
xrajesh, line 286: Nice to check for negative count . @blublinsky

blublinsky: Callers always pass values from _get_token_count(), which can't be negative, so the check isn't needed.

falox:
LGTM. Truncation policy is fine for now. For troubleshooting mode, we should combine this with `tool_budget_ratio=0.5`, but that's for a separate PR @xrajesh @onmete

raptorsun:
The code looks good to me. 
However, I have some doubt about the spending of token budget. It seems the remaining budget of each round of execution is shrinking exponentially. Given an example of 60% `round_cap_fraction`, by the end of 4th round,  it has only 6% of original budge left:
100% -> 40% -> 16% -> 6%
Meanwhile there is already a fair share budget function `tools_round_execution_budget` only called in the `summary` function of token handler class. 
Will it perform better if we prefer the equal share budget for many rounds of tool calling and use the exponetial decay budget only when the remaining budget is above a hard limit?

ols/src/tools/tools.py
```diff
@@ -198,14 +229,15 @@ async def execute_tool_call(
...
+            raw = result[1].get("structured_content")
+            structured_content = raw if isinstance(raw, dict) else None
     else:
         tool_output, was_truncated = _extract_text_from_tool_output(
             result, tools_token_budget
         )
 
     status = "success"
-    logger.debug(
+    logger.warning(
```
raptorsun: should we keep it as debug log?

raptorsun: the UI should show the result is based on truncated messages.

blublinsky:
> The code looks good to me.
> However, I have some doubt about the spending of token budget. It seems the remaining budget of each round of execution is shrinking exponentially. Given an example of 60% `round_cap_fraction`, by the end of 4th round, it has only 6% of original budge left:
> 100% -> 40% -> 16% -> 6%
> Meanwhile there is already a fair share budget function `tools_round_execution_budget` only called in the `summary` function of token handler class.
> Will it perform better if we prefer the equal share budget for many rounds of tool calling and use the exponetial decay budget only when the remaining budget is above a hard limit?

tools_round_execution_budget is already the “fair share vs fraction” idea you describe: it returns min(int(remaining × round_cap_fraction), remaining // horizon) (see TokenBudgetTracker.tools_round_execution_budget in ols/utils/token_handler.py). The horizon is meant to be “rounds left including this one.” Right now that value is only used in summary() for the tools_exec_budget= line and last_tools_exec_budget — it is not what docs_summarizer passes into tool execution; the live path still uses tools_round_budget (pure fraction of remainder).

ols/src/tools/tools.py
blublinsky: fixed

PR #2886, opened 2026-04-09
OLS-2277: Offload large tool outputs to disk with search + read retrieval

onmete:
## Description

Replace the existing character-level truncation in the tool execution chain with an offload-to-disk mechanism. When a single tool's output exceeds `max_tokens_per_tool_output` (default 8,000 tokens), write the full content to a temporary file and inject a short placeholder into the conversation. Two retrieval tools are conditionally registered so the model can search and read offloaded content:

- **`search_offloaded_content`** — grep-style regex search with context lines (find where something is)
- **`read_offloaded_content`** — read a specific line range (examine a section found via search)

This mirrors the proven Cursor/Claude Code pattern: search to find, read to examine. All files are deleted when the HTTP request completes via a `try/finally` block in `generate_response`. Truncation is preserved as a fallback when disk writes fail or content exceeds the 50 MB hard cap.

**Story:** [OLS-2277](https://redhat.atlassian.net/browse/OLS-2277)
**Scope:** large
**Risk:** medium

### Key Design Decisions

- **Ref-ID allowlist** prevents path traversal from LLM input (ADR decision 5)
- **`O_CREAT|O_EXCL`** file creation prevents symlink attacks
- **Regex timeout** prevents ReDoS on search patterns
- **Truncation preserved** as fallback for disk failures or >50 MB content
- **`try/finally` cleanup** in `generate_response` covers both streaming and non-streaming paths
- **Startup cleanup** removes orphaned files from previous crashes
- **`readOnlyHint: True`** on retrieval tools bypasses approval flow

### Files Changed

| File | Action | Rationale |
|------|--------|-----------|
| `ols/src/tools/offloaded_content.py` | Create | `OffloadManager` and both retrieval tools |
| `ols/constants.py` | Modify | New constants for offloading defaults |
| `ols/app/models/config.py` | Modify | New config fields: `offload_storage_path`, `max_tokens_per_tool_output` |
| `ols/src/tools/tools.py` | Modify | Thread offload manager through execution chain |
| `ols/src/query_helpers/docs_summarizer.py` | Modify | OffloadManager lifecycle, retrieval tool registration |
| `ols/app/main.py` | Modify | Startup cleanup of offload storage directory |
| `tests/unit/tools/test_offloaded_content.py` | Create | Unit tests for offloading module |
| `tests/unit/tools/test_tools.py` | Modify | Updated for new offload_manager parameter |
| `tests/unit/query_helpers/test_docs_summarizer.py` | Modify | Tests for lifecycle and retrieval tool registration |
| `tests/unit/app/models/test_config.py` | Modify | Config model tests for new fields |

## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue [OLS-2277](https://redhat.atlassian.net/browse/OLS-2277)

## Testing

Unit tests cover all acceptance criteria:

- **`test_offloaded_content.py`** (new, ~460 lines): Tests for `OffloadManager.try_offload` (threshold, placeholder content, file creation), `search_offloaded_content` (regex matching, context lines, match cap, error handling, ReDoS timeout), `read_offloaded_content` (line ranges, boundary clamping, error handling), cleanup (file deletion, idempotency), and fallback paths (50 MB+ content, disk write failure).
- **`test_tools.py`**: Existing tests pass unchanged with `offload_manager=None`; new test verifies offloading integration in `execute_tool_call`.
- **`test_docs_summarizer.py`**: Tests for retrieval tool mid-loop registration and `OffloadManager` cleanup via `generate_response` finally block.
- **`test_config.py`**: Tests for new `max_tokens_per_tool_output` and `offload_storage_path` config fields with defaults and custom values.

```bash
make test-unit  # All unit tests
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/ols/src/tools/offloaded_content.py
+++ b/ols/src/tools/offloaded_content.py
@@ -1,0 +1,390 @@
+"""Offload large tool outputs to disk with search + read retrieval."""
+
+import logging
+import os
+import re
+import shutil
+import signal
+import tempfile
+from typing import Optional
+from uuid import uuid4
+
+from langchain_core.tools.structured import StructuredTool
+from pydantic import BaseModel, Field
+
+from ols import constants
+
+logger = logging.getLogger(__name__)
+
+
+def cleanup_offload_storage(storage_path: str) -> None:
+    """Remove orphaned offload files from a previous run.
+
+    Call this at application startup to clean up files that may have
+    been left behind after a crash (where try/finally didn't execute).
+
+    Args:
+        storage_path: Directory used for offloaded files.
+    """
+    if not os.path.isdir(storage_path):
+        return
+    try:
+        shutil.rmtree(storage_path)
+        os.makedirs(storage_path, exist_ok=True)
+        logger.info("Cleaned up offload storage directory: %s", storage_path)
+    except OSError:
+        logger.warning(
+            "Failed to clean up offload storage directory: %s",
+            storage_path,
+            exc_info=True,
+        )
+
+
+_CHARS_PER_TOKEN_ESTIMATE = 4
+
+_RETRIEVAL_TOOL_NAMES = frozenset(
+    {"search_offloaded_content", "read_offloaded_content"}
+)
+
+
+class OffloadManager:
+    """Manage offloading of large tool outputs to temporary files on disk.
+
+    Each instance is scoped to a single HTTP request and creates its own
+    isolated session directory via ``tempfile.mkdtemp``. This ensures
+    concurrent requests never interfere with each other's files. The
+    session directory is removed entirely on ``cleanup()``.
+
+    Security properties:
+    - Files are created with ``O_CREAT | O_EXCL`` to prevent symlink attacks.
+    - A ref_id allowlist prevents the LLM from accessing arbitrary paths.
+    """
+
+    def __init__(self, storage_path: str) -> None:
+        """Initialize the offload manager.
+
+        Args:
+            storage_path: Parent directory under which the per-session
+                temp directory will be created.
+        """
+        self._base_path = storage_path
+        self._session_dir: Optional[str] = None
+        self._allowlist: dict[str, str] = {}
+        self._retrieval_tools_built = False
+
+    @property
+    def has_offloaded_content(self) -> bool:
+        """Return True if any content has been offloaded."""
+        return len(self._allowlist) > 0
+
+    @property
+    def retrieval_tools_registered(self) -> bool:
+        """Return True if retrieval tools have been built and registered."""
+        return self._retrieval_tools_built
+
+    def mark_retrieval_tools_registered(self) -> None:
+        """Mark that retrieval tools have been added to the tool loop."""
+        self._retrieval_tools_built = True
+
+    def _ensure_session_dir(self) -> None:
+        """Create the per-session temp directory on first use."""
+        if self._session_dir is None:
+            os.makedirs(self._base_path, exist_ok=True)
+            self._session_dir = tempfile.mkdtemp(prefix="session-", dir=self._base_path)
+
+    def try_offload(self, text: str, tool_name: str, tools_token_budget: int) -> str:
+        """Offload text to disk if it exceeds the per-tool token budget.
+
+        The offload decision is driven by the per-tool share of the round's
+        token budget — the same budget that would truncate the output if
+        offloading is not available.
+
+        Args:
+            text: The full tool output text.
+            tool_name: Name of the tool that produced the output.
+            tools_token_budget: Per-tool token budget for this round.
+
+        Returns:
+            Either the original text (when under budget, fallback, or
+            skipped) or a placeholder string with retrieval instructions.
+        """
+        if tool_name in _RETRIEVAL_TOOL_NAMES:
+            return text
+
+        estimated_tokens = len(text) // _CHARS_PER_TOKEN_ESTIMATE
+        if estimated_tokens <= tools_token_budget:
+            return text
+
+        byte_size = len(text.encode("utf-8"))
+        if byte_size > constants.OFFLOAD_MAX_FILE_SIZE_BYTES:
+            logger.warning(
+                "Tool '%s' output exceeds 50 MB hard cap (%d bytes); "
+                "falling back to truncation",
+                tool_name,
+                byte_size,
+            )
+            return text
+
+        ref_id = str(uuid4())
+        try:
+            self._ensure_session_dir()
+            file_path = os.path.join(self._session_dir, f"{ref_id}.txt")  # type: ignore[arg-type]
+            fd = os.open(
+                file_path,
+                os.O_WRONLY | os.O_CREAT | os.O_EXCL,
+                0o600,
+            )
+            try:
+                os.write(fd, text.encode("utf-8"))
+            finally:
+                os.close(fd)
+        except OSError:
+            logger.warning(
+                "Failed to write offloaded content for tool '%s'; "
+                "falling back to truncation",
+                tool_name,
+                exc_info=True,
+            )
+            return text
+
+        self._allowlist[ref_id] = file_path
+        line_count = text.count("\n") + 1
+
+        return _build_placeholder(ref_id, tool_name, line_count, byte_size)
+
+    def cleanup(self) -> None:
+        """Delete the session directory and all offloaded files.
+
+        Safe to call multiple times. Uses ``shutil.rmtree`` on the
+        per-session directory for atomic cleanup that cannot leak files
+        from concurrent requests.
+        """
+        if self._session_dir is not None and os.path.isdir(self._session_dir):
+            try:
+                shutil.rmtree(self._session_dir)
+            except OSError:
+                logger.warning(
+                    "Failed to remove offload session directory: %s",
+                    self._session_dir,
+                    exc_info=True,
+                )
+        self._session_dir = None
+        self._allowlist.clear()
+
+    def build_retrieval_tools(self) -> list[StructuredTool]:
+        """Build the search and read retrieval tools.
+
+        Returns:
+            List of two StructuredTool instances.
+        """
+        manager = self
+
+        class SearchArgs(BaseModel):
+            """Arguments for search_offloaded_content."""
+
+            ref_id: str = Field(description="Reference ID from the placeholder message")
+            pattern: str = Field(description="Regex pattern to search for")
+            context_lines: int = Field(
+                default=3,
+                ge=0,
+                le=10,
+                description="Number of context lines before and after each match",
+            )
+
+        class ReadArgs(BaseModel):
+            """Arguments for read_offloaded_content."""
+
+            ref_id: str = Field(description="Reference ID from the placeholder message")
+            start_line: int = Field(description="1-based start line number")
+            end_line: int = Field(description="1-based end line number (inclusive)")
+
+        def _search_sync(**kwargs: object) -> str:
+            return _search_offloaded(manager, **kwargs)
+
+        async def _search_async(**kwargs: object) -> str:
+            return _search_offloaded(manager, **kwargs)
+
+        def _read_sync(**kwargs: object) -> str:
+            return _read_offloaded(manager, **kwargs)
+
+        async def _read_async(**kwargs: object) -> str:
+            return _read_offloaded(manager, **kwargs)
+
+        search_tool = StructuredTool(
+            name="search_offloaded_content",
+            description=(
+                "Search offloaded tool output using a regex pattern. "
+                "Returns matching lines with line numbers and context, "
+                "similar to grep -n -C N. Use this to find where "
+                "something is in a large tool output."
+            ),
+            func=_search_sync,
+            coroutine=_search_async,
+            args_schema=SearchArgs,
+            metadata={"annotations": {"readOnlyHint": True}},
+        )
+
+        read_tool = StructuredTool(
+            name="read_offloaded_content",
+            description=(
+                "Read a specific line range from offloaded tool output. "
+                "Returns lines with line numbers. Use this after "
+                "search_offloaded_content to examine a section in detail."
+            ),
+            func=_read_sync,
+            coroutine=_read_async,
+            args_schema=ReadArgs,
+            metadata={"annotations": {"readOnlyHint": True}},
+        )
+
+        return [search_tool, read_tool]
+
+
+def _build_placeholder(
+    ref_id: str, tool_name: str, line_count: int, byte_size: int
+) -> str:
+    """Build a placeholder string for offloaded content."""
+    return (
+        f"[Offloaded: {tool_name}]\n"
+        f"ref_id: {ref_id}\n"
+        f"lines: {line_count} | size: {byte_size} bytes\n"
+        f"\n"
+        f"The full output has been saved. To find information, use:\n"
+        f'  search_offloaded_content(ref_id="{ref_id}", pattern="<regex>")\n'
+        f"To read a specific section found by search, use:\n"
+        f'  read_offloaded_content(ref_id="{ref_id}", start_line=N, end_line=M)\n'
+    )
+
+
+def _load_lines_or_error(
+    manager: OffloadManager, ref_id: str
+) -> tuple[Optional[list[str]], Optional[str]]:
+    """Validate ref_id and load lines, returning (lines, None) or (None, error_message)."""
+    if ref_id not in manager._allowlist:
+        available = list(manager._allowlist.keys())
+        return None, (
+            f"Error: unknown reference '{ref_id}'. "
+            f"Available references: {available}"
+        )
+    file_path = manager._allowlist[ref_id]
+    try:
+        with open(file_path, encoding="utf-8") as f:
+            return f.readlines(), None
+    except OSError as e:
+        return None, f"Error: could not read offloaded content for '{ref_id}': {e}"
+
+
+def _find_matches_with_timeout(
+    lines: list[str], compiled: re.Pattern[str], pattern: str
+) -> tuple[Optional[list[int]], Optional[str]]:
+    """Run regex search with SIGALRM timeout. Return (indices, None) or (None, error)."""
+
+    def _timeout_handler(signum: int, frame: object) -> None:
+        raise TimeoutError()
+
+    match_indices: list[int] = []
+    old_handler = signal.signal(signal.SIGALRM, _timeout_handler)
+    signal.alarm(constants.OFFLOAD_REGEX_TIMEOUT_SECONDS)
+    try:
+        for i, line in enumerate(lines):
+            if compiled.search(line):
+                match_indices.append(i)
+    except TimeoutError:
+        return None, (
+            f"Error: search pattern '{pattern}' timed out. Try a simpler pattern."
+        )
+    finally:
+        signal.alarm(0)
+        signal.signal(signal.SIGALRM, old_handler)
+    return match_indices, None
+
+
+def _format_search_results(
+    lines: list[str],
+    match_indices: list[int],
+    ref_id: str,
+    pattern: str,
+    context_lines: int,
+) -> str:
+    """Format matched lines with context into grep-style output."""
+    total_matches = len(match_indices)
+    capped = match_indices[: constants.OFFLOAD_MAX_SEARCH_MATCHES]
+    shown = len(capped)
+
+    included: set[int] = set()
+    for idx in capped:
+        start = max(0, idx - context_lines)
+        end = min(len(lines), idx + context_lines + 1)
+        included.update(range(start, end))
+
+    match_set = set(match_indices)
+    result_parts: list[str] = [
+        f"Showing {shown} of {total_matches} total matches "
+        f"for '{pattern}' in {ref_id}",
+        "",
+    ]
+
+    sorted_included = sorted(included)
+    prev_line_idx = -2
+    for line_idx in sorted_included:
+        if line_idx != prev_line_idx + 1 and prev_line_idx >= 0:
+            result_parts.append("--")
+        line_text = lines[line_idx].rstrip("\n\r")
+        marker = ":" if line_idx in match_set else "-"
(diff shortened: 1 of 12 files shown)


.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+    action: modify
+    rationale: Tests for OffloadManager lifecycle and retrieval tool registration
+---
+
+## Approach
+
+Replace the existing character-level truncation in the tool execution
+chain with an offload-to-disk mechanism. When a single tool's output
+exceeds `max_tokens_per_tool_output` (default 8,000 tokens), write the
+full content to a temporary file and inject a short placeholder into
```
blublinsky: I really prefer not introducing these types of constants. They seem very artificial.  The decision to offload should be made based purely based on the available tools budget.
Introduction of additional constants makes code harder to maintain

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+the conversation. Two retrieval tools are conditionally registered so
+the model can search and read offloaded content:
+
+- **`search_offloaded_content`** — grep-style regex search with
+  context lines. Use case: find where something is.
+- **`read_offloaded_content`** — read a specific line range. Use
+  case: examine a section found via search.
+
```
blublinsky: This is useful for text data that is not always a case. I suggest in additiona to storing data try to decide data MIme type. Something like:

def sniff_mime_type(content: str) -> str:
    """Detect MIME type from content patterns"""
    stripped = content.strip()
    
    # JSON detection
    if (stripped.startswith('{') and stripped.endswith('}')) or \
       (stripped.startswith('[') and stripped.endswith(']')):
        try:
            json.loads(stripped[:1000])  # Test parse first 1KB
            return "application/json"
        except:
            pass
        # CSV detection (comma-separated with consistent columns)
    lines = stripped.split('\n')[:5]
    if len(lines) > 1 and ',' in lines[0]:
        comma_counts = [line.count(',') for line in lines]
        if len(set(comma_counts)) == 1:  # Consistent column count
            return "text/csv"
    
    # XML detection
    if stripped.startswith('<?xml') or stripped.startswith('<'):
        return "application/xml"
    
    # Log detection (timestamp patterns)
    if re.match(r'\d{4}-\d{2}-\d{2}', stripped):
        return "text/x-log"
    
    return "text/plain"

Once LLM has the MIME type, it can use tools based on the data type
Another important thing to return to LLM is the file size

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+read to examine. All files are deleted when the HTTP request completes
+via a try/finally block in `generate_response`. Truncation is preserved
+as a fallback when disk writes fail or content exceeds the 50 MB hard
+cap.
+
+This is a single coherent change: the offloading mechanism, the
+retrieval tools, the config surface, and the lifecycle management
+are all tightly coupled and meaningless in isolation.
+
+## Per-Criterion Plan
```
blublinsky: The usefull set of tools might include
For text
| Tool              | Purpose             | LLM Use Case                                     |
| ----------------- | ------------------- | ------------------------------------------------ |
| **`head`**        | First N lines       | Quick preview, check file format                 |
| **`tail`**        | Last N lines        | Check recent logs, endings                       |
| **`grep`**   | search within the file      | get partial data |

For structured data, based on Mime type:

| Tool                   | Purpose               | LLM Use Case                           |
| ---------------------- | --------------------- | -------------------------------------- |
| **`jq`**               | JSON querying         | Essential for JSON logs/API responses  |
| **`yq`**               | YAML processing       | Config files, Kubernetes manifests     |
| **`csvgrep`** (csvkit) | CSV filtering         | Table data without loading into memory |
| **`csvcut`**           | CSV column extraction | Select specific columns                |
| **`xmlstarlet`**       | XML processing        | Config files, legacy systems           |

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+  context lines. Use case: find where something is.
+- **`read_offloaded_content`** — read a specific line range. Use
+  case: examine a section found via search.
+
+This mirrors the proven Cursor/Claude Code pattern: search to find,
+read to examine. All files are deleted when the HTTP request completes
+via a try/finally block in `generate_response`. Truncation is preserved
+as a fallback when disk writes fail or content exceeds the 50 MB hard
+cap.
+
```
blublinsky: There is an interesting edge case in this proposal -recursion. If we save a large reply on disk and run a tool on it the result can be too large. Do we save partial in this case?

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+## Per-Criterion Plan
+
+### AC 1: Tool outputs exceeding threshold written to temp file
+
+- **Approach:** Introduce an `OffloadManager` class in the new module
+  that owns a `try_offload(text, tool_name)` method. This method
+  estimates token count using the existing 4 chars/token heuristic,
+  and if the output exceeds the threshold, writes the full text to a
+  uniquely-named file under the configured storage directory. The
+  method returns either the original text (under threshold or fallback)
```
blublinsky: checking size does not seem to belong here. It should be done before invoking this method. This decision can come from 2 different places - read or verify that lools replies fit into window

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+  the full text. Test that outputs under the threshold pass through
+  unchanged.
+
+### AC 2: Placeholder message with ref_id, size, and instructions
+
+- **Approach:** The placeholder is a short string (~50-100 tokens)
+  built by `OffloadManager.try_offload`. It includes: the ref_id,
+  the original tool name, total line count, byte size, and a clear
+  instruction telling the model to use `search_offloaded_content`
+  to find relevant information or `read_offloaded_content` to read
```
blublinsky: The response only needs this:
"type": "resource",
                    "resource": {
                        "uri": offload_info["uri"],
                        "mimeType": "text/plain",
                        "size": size,
It is a response for a specific tool, there is no reason to return tool info here. URI is a file location in the form  file://temp_path  LLM/tools will know how to use it

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+
+- **Test:** Unit test that calls search then read on the same
+  ref_id, verifying each returns correct results. Test multiple
+  search calls with different patterns.
+
+### AC 5: Search results capped with total match count
+
+- **Approach:** A constant (e.g. 50) limits the number of match
+  blocks returned per search call. The search function counts all
+  matches first, then returns only the first N with a header line:
```
blublinsky: This is the recursion that I commented on before. The number of match blocks is crude - the block size can vary a lot
We need a better approach here

.reins/plans/OLS-2277.md
```diff
@@ -0,0 +1,402 @@
...
+
+- **Test:** Offload a file with many matches, verify only the cap
+  number are returned, and the header correctly reports the total.
+  Test read with a range exceeding the cap, verify it is clamped.
+
+### AC 6: Offloaded files deleted on request completion
+
+- **Approach:** `OffloadManager.cleanup` iterates the allowlist and
+  deletes each file, logging any I/O errors but never raising.
+  `DocsSummarizer.generate_response` wraps its body in try/finally,
```
blublinsky: should the manager keep the list of created files or it blindly deletes peredefined folder?

ols/src/tools/offloaded_content.py
```diff
@@ -0,0 +1,373 @@
...
+        f"[Offloaded: {tool_name}]\n"
+        f"ref_id: {ref_id}\n"
+        f"lines: {line_count} | size: {byte_size} bytes\n"
+        f"\n"
+        f"The full output has been saved. To find information, use:\n"
+        f'  search_offloaded_content(ref_id="{ref_id}", pattern="<regex>")\n'
+        f"To read a specific section found by search, use:\n"
+        f'  read_offloaded_content(ref_id="{ref_id}", start_line=N, end_line=M)\n'
+    )
+
```
blublinsky, line 257: This seems way too complex. What about something like:


import re
import subprocess
from langchain.tools import tool

@tool
def grep_local_file(uri: str, pattern: str, context_lines: int = 2) -> str:
    """
    Search for pattern in a locally offloaded file using grep.
    Use this when the MCP result indicates a local file resource.
    
    Args:
        uri: The file:// URI from the MCP resource reference
        pattern: Regex pattern to search for
        context_lines: Number of context lines before/after match
    """
    path = offload_manager.get_path(uri)
    if not path or not path.exists():
        return f"Error: File not found for URI {uri}"
    
    try:
        # Use ripgrep if available, fallback to grep
        cmd = [
            "rg", 
            "-i",  # case insensitive
            "-C", str(context_lines),  # context
            "--color=never",
            pattern,
            str(path)
        ]
        result = subprocess.run(cmd, capture_output=True, text=True, timeout=30)
        
        if result.returncode == 0:
            matches = result.stdout
            # Truncate if too large
            if len(matches) > 8000:
                matches = matches[:8000] + "\n... [truncated]"
            return f"Found matches:\n{matches}"
        elif result.returncode == 1:
            return "No matches found"
        else:
            return f"Error: {result.stderr}"
            
    except FileNotFoundError:
        # Fallback to Python implementation if rg not available
        return _python_grep(path, pattern, context_lines)

def _python_grep(path: Path, pattern: str, context: int) -> str:
    """Pure Python grep fallback"""
    matches = []
    regex = re.compile(pattern, re.IGNORECASE)
    
    with open(path, 'r') as f:
        lines = f.readlines()
    
    for i, line in enumerate(lines):
        if regex.search(line):
            start = max(0, i - context)
            end = min(len(lines), i + context + 1)
            match_block = "".join(lines[start:end])
            matches.append(f"Line {i+1}:\n{match_block}\n---")
    
    if not matches:
        return "No matches found"
    
    result = "\n".join(matches[:20])  # Limit matches
    if len(matches) > 20:
        result += f"\n... and {len(matches) - 20} more matches"
    return result

ols/src/tools/offloaded_content.py
```diff
@@ -0,0 +1,373 @@
...
+    """Manage offloading of large tool outputs to temporary files on disk.
+
+    Each instance is scoped to a single HTTP request. Files are cleaned
+    up via ``cleanup()`` in a try/finally block.
+    """
+
+    def __init__(
+        self,
+        storage_path: str,
+        max_tokens_per_tool_output: int,
```
blublinsky: This seems to belong to the invoker, not here

ols/src/tools/offloaded_content.py
```diff
@@ -0,0 +1,373 @@
...
+
+_CHARS_PER_TOKEN_ESTIMATE = 4
+
+_RETRIEVAL_TOOL_NAMES = frozenset(
+    {"search_offloaded_content", "read_offloaded_content"}
+)
+
+
+class OffloadManager:
+    """Manage offloading of large tool outputs to temporary files on disk.
```
blublinsky, line 51: How about keeping things simple:

import os
import hashlib
import tempfile
from pathlib import Path
from typing import Optional, Dict, Any, List
from langchain.tools import BaseTool
from pydantic import Field
import aiofiles
import asyncio
import shutil

class DiskOffloadManager:
    """Manages offloaded MCP results with automatic cleanup"""
    
    def __init__(self, cache_dir: Optional[str] = None, persist: bool = False):
        """
        Args:
            cache_dir: Directory for offloaded files (auto-generated if None)
            persist: If True, don't delete files on cleanup (for debugging)
        """
        self.persist = persist
        self._registry: Dict[str, Path] = {}  # uri -> path mapping
        self._files: List[Path] = []  # Track all created files for cleanup
        
        if cache_dir:
            self.cache_dir = Path(cache_dir)
        else:
            # Unique temp dir per session
            self.cache_dir = Path(tempfile.mkdtemp(prefix="mcp_offload_"))
            
        self.cache_dir.mkdir(parents=True, exist_ok=True)
    
    async def offload(self, data: str, mime_type: str = "text/plain") -> Dict[str, Any]:
        """Offload large data to disk, return resource reference"""
        content_hash = hashlib.sha256(data.encode()).hexdigest()[:16]
        filename = f"{content_hash}.txt"
        filepath = self.cache_dir / filename
        
        # Write to disk
        async with aiofiles.open(filepath, 'w') as f:
            await f.write(data)
        
        # Track for cleanup
        self._files.append(filepath)
        
        # Register for lookup
        uri = f"file://localhost{filepath}"
        self._registry[uri] = filepath
        
        return {
            "uri": uri,
            "mimeType": mime_type,
            "size": len(data),
            "path": str(filepath)
        }
    
    def get_path(self, uri: str) -> Optional[Path]:
        """Resolve URI to filesystem path"""
        # Handle both full URIs and just paths
        if uri.startswith("file://"):
            return self._registry.get(uri)
        # Fallback: check if it's a direct path reference
        path = Path(uri)
        return path if path.exists() else None
    
    async def cleanup(self) -> None:
        """Remove all offloaded files and optionally the cache directory"""
        if self.persist:
            return
            
        # Remove individual files
        for filepath in self._files:
            try:
                if filepath.exists():
                    filepath.unlink()
            except OSError as e:
                print(f"Warning: Could not delete {filepath}: {e}")
        
        # Try to remove directory if empty and we created it
        try:
            if self.cache_dir.exists() and not any(self.cache_dir.iterdir()):
                self.cache_dir.rmdir()
        except OSError:
            pass  # Directory not empty or permission issue
            
        self._registry.clear()
        self._files.clear()

.reins/plans/OLS-2277.md
onmete: Good point. Removed `max_tokens_per_tool_output` as a separate constant and config field. The offload decision now uses the per-tool share of the round's token budget — the same budget that would truncate the output if offloading is unavailable. This makes the decision context-aware: large outputs are offloaded when budget is tight, and passed through inline when budget is ample.

See commit a2f96632.

.reins/plans/OLS-2277.md
onmete: The model already knows the tool name from the tool description, which provides a natural hint about the data format. If MCP metadata provides content type in the future, we can surface it then. For now we treat everything as plain text — adding heuristic MIME sniffing would be unreliable (is `{"key": "value"}` JSON or a Python dict?) and the grep/read pattern works the same regardless of format.

.reins/plans/OLS-2277.md
onmete: Interesting direction for the future. For this story we're keeping it simple with grep + read — two tools that work on any text format. Format-specific tools (jq, yq, csvgrep) would require external binaries in the container image and significantly expand scope. Could be a follow-up if we find the model struggles with structured data through regex search.

.reins/plans/OLS-2277.md
onmete: Good question. This is handled: `try_offload` skips tools whose names match the retrieval tools (`search_offloaded_content`, `read_offloaded_content`), so their outputs are never offloaded. The retrieval results are also naturally bounded — search caps at 50 matches, read caps at 500 lines — so they stay within normal token budget enforcement.

.reins/plans/OLS-2277.md
onmete: Agreed — refactored. Split `_extract_text_from_tool_output` into `_convert_tool_output_to_text` (pure format conversion) and `_truncate_tool_text` (size enforcement). The `skip_truncation` flag and the redundant dual-call pattern are gone. `execute_tool_call` now converts once, then either offloads or truncates.

See commit a2f96632.

.reins/plans/OLS-2277.md
onmete: LLMs don't natively understand `file://` URIs — they need explicit instructions about which tools to call. The text placeholder with tool call examples is how Cursor and Claude Code achieve reliable retrieval behavior. Also, LangChain's ToolMessage format is just text content, so we'd encode the resource reference as text anyway. The explicit instructions are the UX interface to the model.

.reins/plans/OLS-2277.md
onmete: Fair point about variable block sizes. The match cap is a first-order bound; the existing `enforce_tool_token_budget` in the tool execution chain is the actual safety net — it truncates any tool output that exceeds the remaining budget regardless of offloading. So even if search produces a large result, it gets budget-enforced before entering the conversation.

.reins/plans/OLS-2277.md
onmete: Both. Each `OffloadManager` now creates its own per-session temp directory via `tempfile.mkdtemp`. Files are tracked in an allowlist (for ref_id validation — security), and cleanup uses `shutil.rmtree` on the session directory. This ensures concurrent requests never interfere with each other's files. The startup `cleanup_offload_storage` wipes the entire parent directory to clean up orphaned session dirs from previous crashes.

See commit a2f96632.

ols/src/tools/offloaded_content.py
onmete: The runtime image is `rhel9/python-312-minimal` — a stripped-down image with no external tools installed (`rg`, `grep`, etc. are not available). Using `subprocess.run` would also introduce a dependency on external binaries and testing complexity. Python `re` with a SIGALRM timeout is portable, has no external dependencies, and is appropriate for searching a single file we control.

ols/src/tools/offloaded_content.py
onmete: Agreed — refactored. Split into `_convert_tool_output_to_text` (pure conversion) and `_truncate_tool_text` (size enforcement). The size decision is now cleanly separated from format conversion.

See commit a2f96632.

ols/src/tools/offloaded_content.py
onmete: Good suggestion on per-session isolation — adopted that idea. Each `OffloadManager` now creates its own temp directory via `tempfile.mkdtemp(prefix=\"session-\", dir=storage_path)`. Cleanup is `shutil.rmtree(session_dir)`, which is simpler and guarantees no cross-request interference.

We kept the security properties from the original design that the proposed rewrite would have lost:
- **\`O_CREAT | O_EXCL\`** for file creation prevents symlink-following attacks
- **ref_id allowlist** prevents the LLM from accessing arbitrary file paths (the model only sees opaque UUIDs, never file paths)
- No new dependencies (\`aiofiles\` is not needed — asyncio is single-threaded so sync file I/O has no race conditions)

See commit a2f96632.

.reins/plans/OLS-2277.md
blublinsky: They do not need to do this. This can be done by the tool

.reins/plans/OLS-2277.md
blublinsky: There is only one created by doc_summarizer for every request?

ols/src/tools/offloaded_content.py
blublinsky: I would rather use external tools then bloat Python code. Just me

.reins/plans/OLS-2277.md
onmete: LangChain's `ToolMessage` is plain text — there's no resource/URI type that the tool framework auto-resolves. The tool loop in `iterate_with_tools` receives string content from each tool call and feeds it back as a `ToolMessage` into the next LLM turn. There's no built-in mechanism for "the tool handles the URI transparently."

The placeholder text with search/read instructions is what gives the model the contract for retrieval. The model sees a ref_id, knows the available tools, and calls them — same pattern Cursor and Claude Code use.

.reins/plans/OLS-2277.md
onmete: Yes, exactly one per `generate_response` call. It's created right before the tool loop and cleaned up in the `finally` block, so the lifecycle matches the HTTP request. Each manager creates its own session directory via `tempfile.mkdtemp`, so concurrent requests get fully isolated directories.

ols/src/tools/offloaded_content.py
```diff
@@ -0,0 +1,390 @@
...
+    def mark_retrieval_tools_registered(self) -> None:
+        """Mark that retrieval tools have been added to the tool loop."""
+        self._retrieval_tools_built = True
+
+    def _ensure_session_dir(self) -> None:
+        """Create the per-session temp directory on first use."""
+        if self._session_dir is None:
+            os.makedirs(self._base_path, exist_ok=True)
+            self._session_dir = tempfile.mkdtemp(prefix="session-", dir=self._base_path)
+
```
blublinsky, line 94: This is classical example of LLM-based overengineering. This could be done creating a unique id in init

ols/src/tools/offloaded_content.py
```diff
@@ -0,0 +1,390 @@
...
+                os.write(fd, text.encode("utf-8"))
+            finally:
+                os.close(fd)
+        except OSError:
+            logger.warning(
+                "Failed to write offloaded content for tool '%s'; "
+                "falling back to truncation",
+                tool_name,
+                exc_info=True,
+            )
```
blublinsky, line 147: this goes back to _extract_text_from_tool_output to cut it?
Much simpler would be to do it right here

ols/src/tools/offloaded_content.py
```diff
@@ -0,0 +1,390 @@
...
+            if compiled.search(line):
+                match_indices.append(i)
+    except TimeoutError:
+        return None, (
+            f"Error: search pattern '{pattern}' timed out. Try a simpler pattern."
+        )
+    finally:
+        signal.alarm(0)
+        signal.signal(signal.SIGALRM, old_handler)
+    return match_indices, None
```
blublinsky, line 299: did you look at using
asyncio.to_thread(...) + asyncio.wait_for(..., timeout=...) around a plain synchronous “scan all lines” function.

Pros: portable, idiomatic in an async app, no signals.
Cons: uses a thread for that search (usually fine for rare, bounded work).
much simpler

PR #2906, opened 2026-04-27
OLS-2460 fix uvicorn TLS profile handling

onmete:
## Description

Fix a startup crash in the OLS service when `tlsSecurityProfile` is set in the generated `olsconfig.yaml`.

### Problem

The operator PR [openshift/lightspeed-operator#1430](https://github.com/openshift/lightspeed-operator/pull/1430) wires the TLS security profile from the CR into the generated `olsconfig.yaml`. This is the first time the `tlsSecurityProfile` field is actually populated at the `ols_config` level, which activated a code path in the service that had never been exercised.

The uvicorn runner received an `ssl.TLSVersion` enum value (e.g. `TLSv1_2 = 771`) from the TLS profile resolution code and passed it as `ssl_version=771` to `uvicorn.run()`. Uvicorn feeds that directly into `ssl.SSLContext(771)`, which only accepts `ssl.PROTOCOL_*` constants (e.g. `ssl.PROTOCOL_TLS_SERVER = 17`), not raw TLS version numbers.

Result: `ValueError: invalid or unsupported protocol version 771` → the service CrashLoops and the operator e2e CI (`bundle-e2e-4-21`) fails on every run.

The operator sends correct data — the bug was in how the service consumed it.

### Fix

Separate two concerns that were conflated in `get_ssl_version()`:

1. **Protocol constant** for `ssl.SSLContext()` construction — always `ssl.PROTOCOL_TLS_SERVER`
2. **Minimum TLS version** to enforce — set via `context.minimum_version` after the context is created

The runner now uses `uvicorn.Config` + `uvicorn.Server` instead of `uvicorn.run()` so it can call `config.load()`, set `config.ssl.minimum_version` from the resolved profile, and then `server.run()`. Uvicorn guards against double-load (`assert not self.loaded` / `if not config.loaded`), so pre-loading is safe.

The provider-level TLS code (`ols/src/llms/providers/provider.py`) already handles this correctly — builds `ssl.create_default_context()` and sets `context.minimum_version`. This PR aligns the uvicorn runner with the same pattern.

Unblocks https://github.com/openshift/lightspeed-operator/pull/1430

## Type of change

- Bug fix

- Related Issue #[OLS-2460](https://redhat.atlassian.net/browse/OLS-2460)
- Closes #[OLS-2460](https://redhat.atlassian.net/browse/OLS-2460)

## Testing

- `uv run pytest tests/unit/utils/test_ssl.py tests/unit/runners/test_uvicorn_runner.py`
- Result: 17 passed

Made with Cursor

### DIFF
--- a/ols/runners/uvicorn.py
+++ b/ols/runners/uvicorn.py
@@ -4,7 +4,7 @@ import logging
 
 import uvicorn
 
-from ols.utils import ssl
+from ols.utils import ssl as ssl_utils
 from ols.utils.config import AppConfig
 
 logger: logging.Logger = logging.getLogger(__name__)
@@ -25,7 +25,7 @@ def start_uvicorn(config: AppConfig) -> None:
     )
     log_level = config.ols_config.logging_config.uvicorn_log_level
 
-    # The tls fields can be None, which means we will pass those values as None to uvicorn.run
+    # The tls fields can be None, which means we will pass those values through to Uvicorn.
     ssl_keyfile = config.ols_config.tls_config.tls_key_path
     ssl_certfile = config.ols_config.tls_config.tls_certificate_path
     ssl_keyfile_password = config.ols_config.tls_config.tls_key_password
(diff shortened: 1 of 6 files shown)


onmete:
stuck prow tests
/retest
/approve

onmete:
@JoaoFula mind rechecking please?

PR #2926, opened 2026-05-12
OLS-3310: Add OKP Solr hybrid RAG via tool-based retrieval

blublinsky:
## Description

**Summary**

Adds optional Solr hybrid retrieval against the OKP portal-rag `hybrid-search`
endpoint, exposed as a `search_openshift_documentation` LangChain tool.
When `ols_config.solr_hybrid` is present in the config, the model can invoke
the tool to search published Red Hat product documentation. Solr failures
degrade to empty passages instead of failing the request.

Embedding models (`sentence-transformers/all-mpnet-base-v2` for BYOK/tools/skills
RAG, `ibm-granite/granite-embedding-30m-english` for OKP Solr) are bundled in
the service image via Cachi2 and loaded offline — no runtime downloads.

BYOK FAISS RAG is preserved and coexists with the Solr tool path.

**Key changes**

- `SolrHybridSettings` config model (`solr_http_base`, `max_results`,
  `hybrid_pool_docs`, `hybrid_vector_boost`, `hybrid_score_threshold`,
  `chunk_filter_query`, `hybrid_solr_timeout_s`)
- `SolrHybridSearch` async client with reused `httpx.AsyncClient` for
  connection pooling
- `search_openshift_documentation` async `StructuredTool` for agent tool loop
- Prompt supplement for ASK-mode when the Solr docs tool is active
- Embedding models pre-populated in HuggingFace cache at build time
- `Containerfile` + `artifacts.lock.yaml` updated for both embedding models

## Type of change

- [x] New feature

- Closes [OLS-3310](https://redhat.atlassian.net/browse/OLS-3310)
- Related [OLS-3311](https://redhat.atlassian.net/browse/OLS-3311)

## Testing

Unit tests (`make test-unit`) and integration tests (`make test-integration`)
pass. `make verify` clean (only pre-existing `woke` finding on `origin/main`).

**Local validation:**
1. Set `OLS_CONFIG_FILE` with `solr_hybrid.solr_http_base` pointing at an
   OKP portal-rag instance
2. Ask an OpenShift question — the model invokes
   `search_openshift_documentation` and cites `docs_url` from results
3. Without `solr_hybrid` section — BYOK FAISS RAG works as before

## Summary by CodeRabbit

* **New Features**
  * Added optional Solr-hybrid documentation search (`solr_hybrid`) with `solr_http_base`, retrieval/reranking controls, and a fixed Granite embedding model for vector consistency.
  * Enabled dedicated prompt/tool guidance for the OpenShift docs Solr tool when active.
* **Bug Fixes**
  * Improved configuration validation (URL scheme enforcement, BYOK-only local index requirement when Solr hybrid is enabled).
  * Updated response generation and readiness behavior to correctly handle missing reference content/RAG.
* **Documentation**
  * Documented `solr_hybrid` YAML structure and example/migration guidance.
* **Tests / Chores**
  * Expanded unit tests for Solr-hybrid parsing/validation/tool behavior; updated container/lockfile for the bundled embedding model.


ols/src/prompts/prompts.py
```diff
@@ -85,6 +85,20 @@
 * Terseness must not omit critical info.
 """
 
+SOLR_DOCS_TOOL_AGENT_SUPPLEMENT = """
+Solr docs tool:
+* ``search_openshift_documentation`` searches published Red Hat product docs (not the cluster). Prefer one short query before answering from memory on product behavior, APIs, operators, upgrades, or whether docs exist.
```
raptorsun: "(not the cluster)" seems not clear: not the documents about this specific cluster? tool does not search inside the cluster for related objects?

docs/ai/config.md
```diff
@@ -103,6 +103,27 @@ Config classes have two validation phases:
 
 If you add a new config class that needs file existence checks or cross-section validation, add a `validate_yaml()` method and call it from the appropriate parent config's `validate_yaml()`.
 
+### Optional Solr hybrid RAG (`ols_config.solr_hybrid`)
+
+Omit the key to leave Solr hybrid RAG off (no client, no tool). When present, values map to `SolrHybridSettings`; `enabled: false` disables the feature without removing the block. `solr_http_base` must be a valid `http` or `https` URL with a host when `enabled` is true (checked in `validate_yaml()`).
+
+When ``solr_hybrid.enabled`` is true, every entry under ``reference_content.indexes``
```
raptorsun: solr_hybrid in config does not have the property enabled.
we can change the phrase to "when solr_hybrid" is defined in the config" corresponding to current code in PR

ols/src/rag_index/solr_support.py
```diff
@@ -0,0 +1,587 @@
...
+            return dedupe_retrieved_chunks_by_parent(hybrid_chunks)[: cfg.max_results]
+
+        if not hybrid_docs:
+            logger.warning("No results (hybrid-search) for: %s", query)
+        elif cfg.hybrid_score_threshold > 0:
+            logger.warning(
+                "No docs above hybrid_score_threshold=%s for: %s",
+                cfg.hybrid_score_threshold,
+                query,
+            )
```
raptorsun, line 380: should it be added here the fallback to "/select" endpoint?

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -161,9 +194,22 @@ async def _prepare_prompt_context(
...
+
         if rag_retriever:
-            retrieved_nodes = rag_retriever.retrieve(query)
-            logger.info("Retrieved %d documents from indexes", len(retrieved_nodes))
+            retrieved_nodes.extend(rag_retriever.retrieve(query))
+
+        if retrieved_nodes:
+            logger.info(
+                "Retrieved %d document nodes for RAG context", len(retrieved_nodes)
+            )
```
raptorsun, line 197: would it be useful to sort the documents from multiple sources using the same standard to avoid those at the beggining of the list are most weighted to the LLM's reasoning?

ols/src/prompts/prompts.py
blublinsky: fixed

ols/src/rag_index/solr_support.py
blublinsky: hybrid_score_threshold defaults to 0.0 (no filtering) and the operator does not expose this parameter — it always uses the default. When someone explicitly sets a threshold, they're choosing "no context over bad context." Falling back to /select would return the same or weaker documents that the hybrid path already rejected, undermining that intent.

docs/ai/config.md
blublinsky: fixed

ols/src/query_helpers/docs_summarizer.py
blublinsky: Valid concern, but in practice:

FAISS and Solr are never used together for OpenShift docs — once we switch to Solr, FAISS-based product documentation goes away.
They can coexist for OpenShift docs (Solr) + BYOK. In that case there are two options: BYOK as direct RAG + Solr as a tool, or both BYOK and Solr as tools (PR for this is coming).
So the mixed-source scoring problem doesn't arise in the direct RAG path — it's either one source or the other, never both unsorted.

raptorsun:
the new code looks good :) 
there is some tests about tool calling failing, worth our attention, like `test_offloaded_tool_output_triggers_retrieval`.

ols/src/rag_index/solr_support.py
```diff
@@ -0,0 +1,720 @@
...
+                    )
+                else:
+                    ordered = [doc]
+                chunk = self._assemble_chunk(doc, ordered)
+                logger.debug(
+                    "Chunk %s: expanded %d→%d siblings (family=%d, budget=%d)",
+                    doc.get("chunk_index", "?"),
+                    1,
+                    len(ordered),
+                    len(family) if per_chunk_budget > 0 else 0,
```
vimalk78: **Bug (must-fix):** `NameError` when `max_expansion_neighbors=0` and `token_budget>0`.

The debug log guards `len(family)` with `per_chunk_budget > 0`, but `family` is only assigned inside the `if per_chunk_budget > 0 and cfg.max_expansion_neighbors > 0` block. When a user sets `max_expansion_neighbors: 0` (valid — documented as "disables expansion") and the tool loop provides a token budget, the first iteration hits `NameError: name 'family' is not defined`. On later iterations it silently uses stale data from the previous doc.

Fix: either initialize `family = []` before the loop, or change the guard to match the assignment condition:
```python
len(family) if (per_chunk_budget > 0 and cfg.max_expansion_neighbors > 0) else 0,
```

ols/src/rag_index/solr_support.py
```diff
@@ -0,0 +1,720 @@
...
+        hybrid_url = f"{base}/solr/{_SOLR_COLLECTION}/hybrid-search"
+
+        def _encode() -> list[float]:
+            vec = self._encode_fn(cleaned)
+            return [float(x) for x in (vec.tolist() if hasattr(vec, "tolist") else vec)]
+
+        query_embedding = await asyncio.to_thread(_encode)
+        vector_str = "[" + ",".join(str(v) for v in query_embedding) + "]"
+        form = self._build_hybrid_form(cleaned=cleaned, vector_str=vector_str)
+        async with httpx.AsyncClient(timeout=cfg.hybrid_solr_timeout_s) as client:
```
vimalk78: **Should-fix:** `httpx.AsyncClient` is created fresh per `search()` call — no connection pooling across invocations.

The PR description says "reused `httpx.AsyncClient` for connection pooling" but only intra-call reuse (hybrid POST + family GETs within one search) happens. Each tool invocation gets a new TCP connection to Solr. For production traffic this causes unnecessary connection churn.

Consider storing a persistent `httpx.AsyncClient` on `SolrHybridSearch` (with proper lifecycle via an `aclose()` hook or startup/shutdown events).

pyproject.toml
```diff
@@ -140,7 +140,7 @@ dependencies = [
     "qdrant-client>=1.13.3",  # For ToolsRAG vector storage (pure Python, no Rust/Cargo)
     "rank-bm25>=0.2.2",  # For ToolsRAG sparse retrieval
     "python-frontmatter>=1.1.0",  # For parsing skill YAML frontmatter
-    "opentelemetry-api>=1.20.0",  # For audit tracing
+    "opentelemetry-api>=1.30.0",
```
vimalk78: **Nice-to-have:** Unrelated dependency change — `opentelemetry-api` floor bumped from `>=1.20.0` to `>=1.30.0`. Keeping changes scoped to the feature helps reviewers and bisection. Consider splitting this into its own commit or PR.

vimalk78 (changes requested):
## Review — OLS-3310 (jira mode, round 1)

**Score: 84/100** · 4/4 Jira ACs pass · 3 code quality findings (1 must-fix, 1 should-fix, 1 nice-to-have)

### Adherence

All four acceptance criteria from OLS-3310 are met:
1. **OCP docs retrieval only via tool** — Solr hybrid retrieval is exposed exclusively through `search_openshift_documentation` StructuredTool.
2. **Direct-RAG code path removed** — no `solr_direct_rag` config or code path exists.
3. **BYOK FAISS unchanged** — `ReferenceContent.indexes`, `IndexLoader`, and the BYOK retrieval path are preserved.
4. **Unit tests updated** — 491 new lines in `test_solr_support.py`, plus updates to config, summarizer, prompt, index loader, health, and environment tests.

### Code quality issues below as inline comments.

ols/src/rag_index/solr_support.py
blublinsky: Fixed in 99ba1e1c — the debug guard now matches the assignment condition (`per_chunk_budget > 0 and cfg.max_expansion_neighbors > 0`), and a unit test covers `max_expansion_neighbors=0` with a non-zero token budget.

ols/src/rag_index/solr_support.py
blublinsky: Fixed in 99ba1e1c — `SolrHybridSearch` now holds a persistent `httpx.AsyncClient` via `_get_http_client()`, with `aclose()` / `close_http_client_sync()` called on config reload. Unit test confirms the client is reused across `search()` calls. Verified locally with 5 concurrent `/v1/query` requests against RHOKP.

pyproject.toml
blublinsky: Fixed in 99ba1e1c — reverted `opentelemetry-api` floor back to `>=1.20.0` and regenerated `uv.lock`.

blublinsky:
@vimalk78 All review items addressed in 99ba1e1c:

1. **Must-fix** (`family` NameError) — debug guard aligned with assignment condition; unit test added.
2. **Should-fix** (connection pooling) — persistent `httpx.AsyncClient` on `SolrHybridSearch` with cleanup on config reload; reuse covered by unit test; concurrent sync queries verified locally.
3. **Nice-to-have** (OTel bump) — `opentelemetry-api` reverted to `>=1.20.0`, `uv.lock` updated.

`make verify` (black, ruff, pylint, mypy) plus unit (1137) and integration (124) tests all pass locally. PTAL when you have a chance.

xrajesh:
The Solr tool returns docs_url and title in every chunk, but  the API response's referenced_documents field is not setting yet ? We need that for the UI to render the clickable links. @blublinsky

blublinsky:
> The Solr tool returns docs_url and title in every chunk, but the API response's referenced_documents field is not setting yet ? We need that for the UI to render the clickable links. @blublinsky

done

PR #2955, opened 2026-06-16
feat: in-process LLM credential hot-reload

lalan7:
## Summary

When `credentialsSecretRef` is updated by a CronJob or external rotation, the lightspeed-operator triggers a rolling restart of `lightspeed-app-server`. For customers using short-lived LLM tokens (1-hour validity), this causes **hourly pod restarts and temporary capacity loss** — including reloading RAG vector indexes and embedding models.

This PR makes `ProviderConfig` re-read the credential file on every LLM request via a new `get_credentials()` method, so rotated secrets are picked up without a pod restart. This follows the Prometheus `credentials_file` pattern: *re-read on every request, no caching, no edge cases with K8s symlink swaps*.

## Changes

- **`ols/utils/checks.py`** — New `read_secret_from_path()` helper for repeated file reads
- **`ols/app/models/config.py`** — Store `_credentials_path` in `ProviderConfig.__init__()`, add `get_credentials()` method that re-reads the file when a path is configured
- **All LLM providers** (`openai`, `azure_openai`, `watsonx`, `rhoai_vllm`, `rhelai_vllm`, `google_vertex`) — Use `get_credentials()` instead of `.credentials`
- **Unit tests** — Tests for the helper, `ProviderConfig.get_credentials()` (cached, re-read, fallback, directory), and OpenAI provider credential rotation
- **`docs/credential-hot-reload.md`** — Design rationale and test plan

## Why this approach

| Concern | Answer |
|---|---|
| K8s symlink safety | `kubelet` uses atomic symlink swaps (`..data`); `os.stat()` mtime can be unreliable after swaps. Re-reading avoids the edge case. |
| Precedent | Prometheus `credentials_file`, Envoy, controller-runtime `CertWatcher`, kube-proxy all re-read files. |
| Dependencies | None — no watchdog, fsnotify, or background threads. |
| Performance | One `open()+read()` of a <100 byte file per LLM request is negligible vs LLM call latency (seconds). |
| Thread safety | Each request reads independently; no shared mutable state. |

## Scope

Hot-reload applies to **top-level `credentials_path`**, which is the path used by operator-managed deployments (`credentialsSecretRef`). Provider-specific config blocks (`openai_config.credentials_path`, `azure_openai_config.credentials_path`, etc.) continue to cache credentials at startup — this matches the existing operator behavior, which does not generate these blocks.

Azure AD credentials (`tenant_id`, `client_id`, `client_secret`) are also read once at startup. These can be addressed in a follow-up if needed.

## Test plan

- [x] `read_secret_from_path()` — file, directory, nonexistent, rotation
- [x] `ProviderConfig.get_credentials()` — cached fallback, disk re-read, read failure fallback, directory path
- [x] OpenAI provider — verifies rotated credential is picked up across two `load_llm()` calls
- [x] Full suite: 1067 tests pass, `ruff check` clean, `mypy` clean

## Ref

- [RFE-9380](https://issues.redhat.com/browse/RFE-9380)

## Summary by CodeRabbit

* **New Features**
  * Added optional credential hot-reload (`credential_hot_reload`, default off).
  * When enabled, providers re-read token/secret values from the configured path (including directory paths) to pick up rotated credentials without restarts.
* **Bug Fixes**
  * If credentials can’t be re-read temporarily, the last successfully loaded values are used.
  * Improved error handling for missing credentials in supported providers.
* **Documentation**
  * Added a credential hot-reload validation guide with rollout/checklist steps.
* **Tests**
  * Added unit tests covering token rotation, caching behavior, and path/directory secret reading.

### DIFF
--- a/docs/credential-hot-reload.md
+++ b/docs/credential-hot-reload.md
@@ -1,0 +1,175 @@
+# RFE-9380: In-Process LLM Credential Hot-Reload
+
+## Problem
+
+When `credentialsSecretRef` is updated, the Lightspeed operator triggers a rolling
+restart of `lightspeed-app-server`. For customers using short-lived LLM tokens
+(e.g. 1-hour validity rotated by CronJob), this causes hourly pod restarts and
+temporary capacity loss.
+
+- **RFE:** [RFE-9380](https://redhat.atlassian.net/browse/RFE-9380)
+- **Customer environment:** OCP 4.21, pre-production
+- **Current workaround:** replicas >= 3 with manual PDB (minAvailable: 2)
+
+## Solution
+
+The service re-reads credential files from disk on every LLM request (Prometheus
+`credentials_file` pattern), so rotated secrets are picked up without a pod
+restart.
+
+### Why this matters
+
+The app-server pod contains the FAISS vector indexes (~100MB) and the
+SentenceTransformer embedding model (~400MB), all loaded into memory at startup.
+A rolling restart means:
+
+| Step | Time |
+|------|------|
+| FAISS index reload | ~5-15s |
+| Embedding model reload | ~3-5s |
+| Readiness probe pass | ~2-5s |
+| **Total cold start** | **~10-25s per pod** |
+
+With hourly rotation and 3 replicas, that is up to 24 restart cycles/day with
+degraded capacity during each one.
+
+### Cost comparison
+
+| Factor | Hot-Reload | 2+ Replicas workaround |
+|--------|-----------|------------------------|
+| CPU per request | 1 file read (~0.05ms) | Zero extra |
+| Memory | Zero extra | 2-3x pod memory |
+| Compute cost | Negligible | 2-3x vCPU reserved |
+| Availability during rotation | 100% | ~95-99% (brief capacity drop) |
+| Ops complexity | None | PDB + replica tuning + alert tuning |
+
+## Design
+
+Since `load_llm()` already creates a new `LLMProvider` per request, we add a
+`get_credentials()` method to `ProviderConfig` that re-reads the credential file
+each time it is called.
+
+```text
+Request → load_llm() → LLMProvider.__init__()
+                          → provider_config.get_credentials()
+                              → open(credentials_path) + read()
+                              → return fresh value
+```
+
+### Why re-read on every request (vs caching or fsnotify)
+
+- **Kubernetes symlink safety** — kubelet uses atomic symlink swaps (`..data`);
+  `os.stat()` mtime can be unreliable after swaps. Re-reading avoids edge cases.
+- **Proven pattern** — Prometheus `credentials_file` / `password_file` re-reads
+  on every scrape request.
+- **No new dependencies** — no watchdog, fsnotify, or background threads.
+- **Negligible overhead** — one `open()+read()` of a <100 byte file per LLM
+  request is trivial vs multi-second LLM call latency.
+- **Thread-safe by default** — each request reads independently.
+
+### Ecosystem alignment
+
+| Project | Pattern |
+|---------|---------|
+| Prometheus | `credentials_file` re-read on every scrape |
+| OpenShift library-go | `fileobserver` hash-based polling (Go operators) |
+| controller-runtime | `CertWatcher` fsnotify + periodic re-read |
+| kube-proxy | watches parent directory (PR #139204) |
+| Stakater Reloader | triggers rolling restart (what we are moving away from) |
+
+## Files Changed
+
+### Core
+
+- `ols/utils/checks.py` — added `read_secret_from_path()` helper
+- `ols/app/models/config.py` — added `_credentials_path` private attr and
+  `get_credentials()` method to `ProviderConfig`
+
+### Providers (7 files)
+
+All providers changed from `self.provider_config.credentials` to
+`self.provider_config.get_credentials()`:
+
+- `ols/src/llms/providers/openai.py`
+- `ols/src/llms/providers/azure_openai.py`
+- `ols/src/llms/providers/watsonx.py`
+- `ols/src/llms/providers/rhoai_vllm.py`
+- `ols/src/llms/providers/rhelai_vllm.py`
+- `ols/src/llms/providers/google_vertex.py` (both Gemini and Anthropic)
+- `ols/src/llms/providers/bedrock.py`
+
+### Tests (9 new tests)
+
+- `tests/unit/utils/test_checks.py` — 4 tests for `read_secret_from_path()`
+- `tests/unit/app/models/test_config.py` — 4 tests for `get_credentials()`
+- `tests/unit/llms/providers/test_openai.py` — 1 end-to-end rotation test
+
+## Testing on a Cluster
+
+### Local with Podman
+
+```bash
+# Build image with changes
+OLS_API_IMAGE=localhost/ols:hot-reload make images
+
+# Prepare config
+mkdir -p /tmp/ols-test/config
+cp examples/olsconfig.yaml /tmp/ols-test/config/olsconfig.yaml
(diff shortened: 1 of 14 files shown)


joshuawilson:
## Adversarial Review

Nice PR — the motivation is solid and the Prometheus `credentials_file` pattern is the right call. A few gaps surfaced during review:

### 1. Bedrock provider not updated (`ols/src/llms/providers/bedrock.py:37`)

Bedrock still uses `self.credentials = self.provider_config.credentials` (direct attribute access). All other 6 providers were updated. Credential rotation will silently fail for Bedrock deployments.

### 2. Provider-specific `api_key` silently overrides hot-reloaded credentials (`ols/src/llms/providers/openai.py:33`)

In `default_params`, line 28 calls `get_credentials()`, but lines 30–34 check `openai_config.api_key` and override with the static value read once at init (via `read_api_key()` at `config.py:577–587`). Same pattern in Azure OpenAI, WatsonX, RHOAI vLLM, and RHELAI vLLM. Deployments using provider-specific config blocks will never pick up rotated credentials. The PR description acknowledges the scope boundary, but there's no runtime warning when this override happens — a deployment could silently have hot-reload disabled.

### 3. Silent fallback to stale credentials masks persistent read failures (`ols/app/models/config.py:612`)

When `_credentials_path` is set but the file can't be read, `get_credentials()` logs a warning and falls back to `self.credentials`. Correct for transient K8s symlink swaps (milliseconds), but no mechanism to distinguish transient from permanent failure. If the file is permanently deleted, the service uses stale/expired credentials until LLM API calls return 401 — with only a warning log as signal.

Consider: logging at a higher level or incrementing an error counter after N consecutive failures, so ops teams can detect persistent credential-read failures.

### 4. Dual access pattern is a maintenance trap (`ols/app/models/config.py:600`)

`.credentials` (stale, cached at init) remains a public field alongside `.get_credentials()` (fresh). New providers or contributors will naturally use `.credentials`. Consider making `.credentials` a `@property` that delegates to `get_credentials()`, or at minimum adding a docstring/comment on the field warning about staleness.

### 5. `read_secret_from_path()` duplicates `read_secret()` (`ols/utils/checks.py:71`)

Both implement directory-resolution (`os.path.isdir` + `os.path.join`) and file-reading (`open` + `read` + `rstrip`). They diverge only in error handling (`print()` vs `logger.warning()`). A shared inner helper would reduce duplication.

### 6. PR title missing `OLS-XXXX` prefix

Per repo conventions, PR titles should start with the Jira reference (`OLS-XXXX description`). Current title uses conventional-commit format instead.

---

Items 1–2 are the most actionable — they're correctness gaps where hot-reload silently doesn't work. Items 3–5 are design/cleanup suggestions. Item 6 is a convention nit.

blublinsky (commented):
Overall this is a clean, well-scoped change. The Prometheus-style per-request re-read is the right pattern — no new dependencies, negligible overhead, and avoids the fsnotify/mtime edge cases with K8s symlink swaps.

Two items:

**1. Stale fallback after successful rotation** (`ols/app/models/config.py:612-618`)

When `get_credentials()` reads a rotated credential successfully, it returns the fresh value but doesn't update `self.credentials`. If a later read fails (transient I/O error), the fallback returns the original startup key — which may have been revoked hours ago — instead of the last-known-good rotated key.

Scenario with 1-hour token rotation on a long-running pod:
1. Startup: `self.credentials = "key-v1"`
2. Hour 1: reads `"key-v2"` from disk, returns it — but `self.credentials` still `"key-v1"`
3. Hour 2: transient read failure → falls back to `self.credentials` → returns revoked `"key-v1"`

Suggested fix — one line:

```python
            if fresh is not None:
                self.credentials = fresh
                return fresh
```

The test `test_provider_config_get_credentials_falls_back_on_read_failure` should also cover this sequence: startup → successful rotation → file disappears → assert fallback returns the rotated key, not the startup key.

**2. Bedrock provider not updated**

The Bedrock provider (`ols/src/llms/providers/bedrock.py`) was merged to `main` via #2959 after this branch was created. It uses `self.provider_config.credentials` directly in `default_params` (line 37), same pattern this PR replaces in all other providers. After rebase it will be the only provider without hot-reload — please update it to `self.provider_config.get_credentials()`.

lalan7:
Hi @joshuawilson — thanks for the thorough adversarial review. Could you clarify the Bedrock point? Bedrock uses IAM-based auth via `boto3` session credentials, not API key files — so `get_credentials()` from the mounted secret path doesn't apply there. The hot-reload is scoped to providers that read API keys from the mounted `credentialsSecretRef` path (OpenAI, Azure OpenAI, Watsonx, etc.).

The CodeRabbit suggestions have been noted — happy to address them in a follow-up if preferred, or squash them in here. All CI checks are green.

Ready for `lgtm` when you're comfortable. cc @tisnik

blublinsky:
> Could you clarify the Bedrock point? Bedrock uses IAM-based auth via `boto3` session credentials, not API key files — so `get_credentials()` from the mounted secret path doesn't apply there.

Bedrock can use 2 different credentials:
1. Bedrock API token
2. IAM creds, potentially with an additional role, that have Bedrock access

Both approaches are referencing a secret with credentials of a certain type. If the secret changes, OLS will restart.
So for your PR you will have to do the same thing that you are doing for other providers - read the secret and update credentials.
You are correct that short lived token is regenerated by Boto3, but IAM creds can change as well

lalan7:
@blublinsky Hi team, this PR implements [RFE-9380](https://redhat.atlassian.net/browse/RFE-9380) (in-process LLM credential hot-reload). All CI checks are passing. Would appreciate a review when you get a chance. Thanks!

blublinsky:
**Should-fix: torn-read risk in `get_aws_credentials()`**

`get_aws_credentials()` writes `self.aws_access_key_id`, `self.aws_secret_access_key`, and `self.role_arn` non-atomically, then returns by reading those same `self.*` attributes. A concurrent request arriving between writes could observe a mismatched pair (e.g. `new_access_key + old_secret_key`), which would fail AWS auth.

Fix: return from local variables instead of `self.*` — matching the pattern `get_credentials()` already uses correctly:

```python
return access_key, secret_key, self.role_arn  # use locals, not self.*
```

Practical risk is low (narrow GIL window, only during rare credential rotation), but it's a genuine correctness bug worth fixing.

blublinsky:
Overall the implementation is clean — we would like to merge it as soon as the comment above is addressed.

lalan7:
@blublinsky  Hi the torn-read fix you flagged on Sep 1 is addressed in commit cd45145. get_aws_credentials() now snapshots self.* into locals before any writes and returns from those locals, matching the pattern get_credentials() uses. All CI is green. Thanks!

blublinsky:
Request changes: Bedrock API-key credentials can be misclassified as IAM credentials when hot reload is enabled.

`ProviderConfig.get_aws_credentials()` currently calls `read_secret_from_path()` for `aws_access_key_id` and `aws_secret_access_key` whenever `_credentials_path` is set. `read_secret_from_path()` treats a non-directory path as the credential file itself, so a Bedrock API-key configuration such as `credentials_path=/.../apitoken` causes both IAM values to be populated with the API-token contents.

That makes `Bedrock.load()` compute `use_iam=True` and select the boto3/SigV4 path instead of the API-key path. The API token is then incorrectly used as both the AWS access key and secret key.

Please restrict IAM hot-reload to directory paths:

```python
if (
    self._credential_hot_reload
    and self._credentials_path is not None
    and os.path.isdir(self._credentials_path)
):
    ...
```

Please also add a regression test covering a Bedrock API-key file with hot reload enabled, asserting that `get_aws_credentials()` returns no IAM credentials and that `Bedrock.load()` uses the API-key authentication path. The provider supports both IAM-directory and API-key-file authentication, so this distinction must be preserved.

PR #2964, opened 2026-06-24
OLS-3221 Add PostgreSQL auto-recovery after DB restart

sriroopar:
Implement background health-check loop, smarter error classification in the @connection decorator, operation timeouts, and enhanced liveness/readiness probes so OLS can recover automatically when the backing PostgreSQL database is restarted.

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

- **New Features**
  - Enhanced the `/liveness` endpoint to detect repeated PostgreSQL connectivity failures.
  - Returns HTTP `503` with a clear “database unreachable” reason when the failure threshold is reached.
  - Added configurable PostgreSQL timeouts, health-check intervals, and liveness failure thresholds.
  - Liveness responses now optionally include an explanation for unhealthy status.

- **Bug Fixes**
  - Improved database connection recovery, transaction safety, and error handling.
  - Added background health monitoring to improve cache readiness reporting.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -944,6 +944,16 @@
                                 }
                             }
                         }
+                    },
+                    "503": {
+                        "description": "Service is not alive",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/LivenessResponse"
+                                }
+                            }
+                        }
                     }
                 }
             }
@@ -1711,6 +1721,17 @@
                     "alive": {
                         "type": "boolean",
                         "title": "Alive"
+                    },
+                    "reason": {
+                        "anyOf": [
+                            {
+                                "type": "string"
+                            },
+                            {
+                                "type": "null"
+                            }
+                        ],
+                        "title": "Reason"
                     }
                 },
                 "type": "object",
@@ -1718,7 +1739,7 @@
                     "alive"
                 ],
                 "title": "LivenessResponse",
-                "description": "Model representing a response to a liveness request.\n\nAttributes:\n    alive: If app is alive.\n\nExample:\n    ```python\n    liveness_response = LivenessResponse(alive=True)\n    ```",
+                "description": "Model representing a response to a liveness request.\n\nAttributes:\n    alive: If app is alive.\n    reason: Optional reason when not alive.\n\nExample:\n    ```python\n    liveness_response = LivenessResponse(alive=True)\n    ```",
                 "examples": [
                     {
                         "alive": true
--- a/ols/app/endpoints/health.py
+++ b/ols/app/endpoints/health.py
@@ -10,7 +10,7 @@ import logging
 import time
 from typing import Any
 
-from fastapi import APIRouter, HTTPException, status
+from fastapi import APIRouter, HTTPException, Response, status
 from langchain_core.messages.ai import AIMessage
 
 from ols import config
@@ -19,6 +19,7 @@ from ols.app.models.models import (
     NotAvailableResponse,
     ReadinessResponse,
 )
+from ols.src.cache.postgres_cache import PostgresCache
 from ols.src.llms.llm_loader import load_llm
 
 router = APIRouter(tags=["health"])
@@ -133,10 +134,31 @@ get_liveness_responses: dict[int | str, dict[str, Any]] = {
         "description": "Service is alive",
         "model": LivenessResponse,
     },
+    503: {
+        "description": "Service is not alive",
+        "model": LivenessResponse,
+    },
 }
 
 
-@router.get("/liveness", responses=get_liveness_responses)
-def liveness_probe_get_method() -> LivenessResponse:
+def _get_postgres_cache() -> PostgresCache | None:
+    """Return the conversation cache if it is a PostgresCache, None otherwise."""
+    try:
+        cache = config.conversation_cache
+        return cache if isinstance(cache, PostgresCache) else None
+    except Exception:
+        return None
+
+
+@router.get(
+    "/liveness", responses=get_liveness_responses, response_model_exclude_none=True
+)
+def liveness_probe_get_method(response: Response) -> LivenessResponse:
     """Live status of service."""
+    cache = _get_postgres_cache()
+    if cache is not None:
+        threshold = config.ols_config.liveness_db_failure_threshold
+        if cache.consecutive_failures >= threshold:
+            response.status_code = status.HTTP_503_SERVICE_UNAVAILABLE
+            return LivenessResponse(alive=False, reason="database unreachable")
     return LivenessResponse(alive=True)
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -53,6 +53,21 @@ def validate_tool_round_cap_fraction_config(v: float) -> float:
     return v
 
 
+def validate_liveness_db_failure_threshold(raw_value: Any) -> int:
+    """Validate ``liveness_db_failure_threshold`` for ``OLSConfig``."""
+    try:
+        threshold = int(raw_value)
+    except (TypeError, ValueError) as e:
+        raise checks.InvalidConfigurationError(
+            f"liveness_db_failure_threshold must be a positive integer, got {raw_value!r}"
+        ) from e
+    if threshold < 1:
+        raise checks.InvalidConfigurationError(
+            f"liveness_db_failure_threshold must be at least 1, got {threshold}"
+        )
+    return threshold
+
+
 class ModelParameters(BaseModel):
     """Model parameters."""
 
@@ -807,6 +822,12 @@ class PostgresConfig(BaseModel):
     gss_encmode: str = constants.POSTGRES_CACHE_GSSENCMODE
     ca_cert_path: Optional[FilePath] = None
     max_entries: PositiveInt = constants.POSTGRES_CACHE_MAX_ENTRIES
+    statement_timeout: PositiveInt = constants.POSTGRES_STATEMENT_TIMEOUT
+    lock_timeout: PositiveInt = constants.POSTGRES_LOCK_TIMEOUT
+    health_check_interval: PositiveInt = constants.CACHE_HEALTH_CHECK_INTERVAL
+    health_check_connect_timeout: PositiveInt = (
+        constants.CACHE_HEALTH_CHECK_CONNECT_TIMEOUT
+    )
     tls_security_profile: Optional["TLSSecurityProfile"] = None
 
     def __init__(self, **data: Any) -> None:
@@ -1261,7 +1282,9 @@ class OLSConfig(BaseModel):
 
     offload_storage_path: str = constants.DEFAULT_OFFLOAD_STORAGE_PATH
 
-    def __init__(  # noqa: C901
+    liveness_db_failure_threshold: int = constants.LIVENESS_DB_FAILURE_THRESHOLD
+
+    def __init__(
         self, data: Optional[dict] = None, ignore_missing_certs: bool = False
     ) -> None:
         """Initialize configuration and perform basic validation."""
@@ -1309,6 +1332,10 @@ class OLSConfig(BaseModel):
         self.quota_handlers = QuotaHandlersConfig(data.get("quota_handlers", None))
         self._propagate_tls_profile()
         self.proxy_config = ProxyConfig(data.get("proxy_config"))
+        self._init_optional_sections(data)
+
+    def _init_optional_sections(self, data: dict) -> None:
+        """Initialize optional config sections to keep __init__ complexity low."""
         if data.get("tool_filtering", None) is not None:
             self.tool_filtering = ToolFilteringConfig(**data.get("tool_filtering"))
         if data.get("tools_approval", None) is not None:
@@ -1333,6 +1360,12 @@ class OLSConfig(BaseModel):
 
         self.offload_storage_path = data.get(
             "offload_storage_path", constants.DEFAULT_OFFLOAD_STORAGE_PATH
+        )
+
+        self.liveness_db_failure_threshold = validate_liveness_db_failure_threshold(
+            data.get(
+                "liveness_db_failure_threshold", constants.LIVENESS_DB_FAILURE_THRESHOLD
+            )
         )
 
     def _propagate_tls_profile(self) -> None:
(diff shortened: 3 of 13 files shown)


sriroopar:
## Adversarial Code Review - PostgreSQL Auto-Recovery ([OLS-3221](https://redhat.atlassian.net/browse/OLS-3221))

### Critical Issues

#### 1. **Race Condition: Failure Counter Reset vs Increment** (`postgres_cache.py:184-188, 226-230`)
There's a race condition between `_mark_unhealthy()` (called from request threads via `@connection` decorator) and the health check loop resetting `_consecutive_failures`:

```python
# Thread A (request): _mark_unhealthy()
self._consecutive_failures += 1  # increments to 3

# Thread B (health loop): success path
self._consecutive_failures = 0   # resets to 0

# Thread A continues but the failure was "lost"
```

**Impact**: Liveness check could oscillate between healthy/unhealthy unpredictably.

**Suggestion**: Consider using an atomic counter or ensuring the health loop only resets failures after a configurable number of successful probes.

---

#### 2. **Potential Data Loss: `@connection` Retries Write Operations** (`postgres.py:31-45`)
The `@connection` decorator retries after `OperationalError`/`InterfaceError`, but `insert_or_append` could have partially committed before the connection error:

```python
# In insert_or_append: connection commits successfully
self.connection.commit()  # <-- succeeds
# Then network drops here, raising OperationalError
# Decorator catches and retries entire operation
# Result: duplicate entry appended
```

**Impact**: Message duplication in conversation history after network glitches.

**Suggestion**: Add idempotency tracking or disable retry for write operations. The advisory lock helps within a transaction but doesn't prevent retry-induced duplicates.

---

#### 3. **Liveness Endpoint Accesses Private Attributes** (`health.py:139-140`)
```python
with cache._health_lock:
    failures = cache._consecutive_failures
```

Accessing `_health_lock` and `_consecutive_failures` directly violates encapsulation and creates tight coupling.

**Suggestion**: Add a public method like `get_consecutive_failures()` or `is_healthy_for_liveness(threshold: int)` to `PostgresCache`.

---

### Medium Issues

#### 4. **Health Check Thread Has No Graceful Shutdown** (`postgres_cache.py:217-244`)
```python
def _health_check_loop(self) -> None:
    while True:  # No exit condition
        time.sleep(self._health_check_interval)
```

The daemon thread runs forever with no way to stop it cleanly. While `daemon=True` means it won't block process exit, this prevents clean shutdown in tests and can cause issues with connection cleanup.

**Suggestion**: Add a `_shutdown_event = threading.Event()` and check it in the loop.

---

#### 5. **Health Check Doesn't Run Immediately** (`postgres_cache.py:219-220`)
```python
def _health_check_loop(self) -> None:
    while True:
        time.sleep(self._health_check_interval)  # Sleeps FIRST
        # ... then checks
```

Initial health status is `True` but the first actual check happens after `health_check_interval` seconds (default 30s). If the DB is unreachable at startup, `ready()` returns `True` for 30 seconds.

**Suggestion**: Perform initial health check before sleeping, or set initial `_health_status = False` until first successful check.

---

#### 6. **Lock Timeout of 0 Causes Immediate Failure** (`postgres_cache.py:260-264`)
```python
def _acquire_lock(self) -> None:
    if not self._tx_lock.acquire(timeout=self._lock_timeout):
        raise CacheError("lock acquisition timeout")
```

If `lock_timeout` is configured as 0, every cache operation will fail immediately. The `NonNegativeInt` validation allows 0.

**Suggestion**: Either use `PositiveInt` for `lock_timeout` in config, or handle 0 specially (e.g., use `blocking=True` for infinite wait).

---

#### 7. **Health Connection State Inconsistency** (`postgres_cache.py:190-210`)
```python
def _connect_health(self) -> None:
    if self._health_connection is not None:
        try:
            self._health_connection.close()
        except Exception:
            pass  # Old connection closed or failed
    # ... if connect() fails here, _health_connection is still the old (closed) object
    self._health_connection = psycopg2.connect(**connect_kwargs)
```

If `psycopg2.connect()` fails, `_health_connection` still references the old closed connection.

**Suggestion**: Set `self._health_connection = None` before attempting new connection.

---

### Minor Issues

#### 8. **Hardcoded `connect_timeout: 10`** (`postgres_cache.py:207`)
The health check connection timeout is hardcoded to 10 seconds. This should be configurable via `PostgresConfig`.

---

#### 9. **No Test for Concurrent Access Under Lock Contention**
The test suite doesn't verify behavior when multiple threads compete for `_tx_lock` with timeout enabled. Consider adding a stress test.

---

### Summary

The auto-recovery implementation is a significant improvement over the previous state, but the thread safety and retry semantics need attention before production deployment. The most critical issues are:

1. Race condition in failure counting
2. Potential duplicate writes on retry
3. Breaking encapsulation in liveness endpoint

sriroopar:
## Adversarial Review - Updated Code (Follow-up)

### Issues Addressed ✅

Great progress! The following issues from my previous review have been addressed:

1. **Encapsulation** - Now using `cache.consecutive_failures` property instead of private attributes
2. **Graceful shutdown** - Added `shutdown()` method and `_shutdown_event`
3. **Configurable connect_timeout** - Added `health_check_connect_timeout` to `PostgresConfig`
4. **Health check runs immediately** - Loop now checks first, then waits at end
5. **Initial connect in try block** - Moved inside try in `@connection` decorator

---

### Remaining Issues

#### 1. **Health Connection State Inconsistency** (`postgres_cache.py:191-211`)

Still not addressed:

```python
def _connect_health(self) -> None:
    if self._health_connection is not None:
        try:
            self._health_connection.close()
        except Exception:
            pass
    # If psycopg2.connect() fails here, _health_connection still references
    # the old (closed) connection object
    self._health_connection = psycopg2.connect(**connect_kwargs)
```

**Fix**: Set `self._health_connection = None` after closing and before attempting new connection:

```python
if self._health_connection is not None:
    try:
        self._health_connection.close()
    except Exception as close_err:
        logger.debug("Failed to close old health connection: %s", close_err)
    self._health_connection = None  # <-- Add this
```

---

#### 2. **`shutdown()` Doesn't Clean Up Resources** (`postgres_cache.py:218-220`)

```python
def shutdown(self) -> None:
    """Signal the health-check thread to stop."""
    self._shutdown_event.set()
```

The method only signals shutdown but doesn't:
- Wait for the thread to finish (`join()`)
- Close `_health_connection`
- Close the main `connection`

**Suggestion**:
```python
def shutdown(self, timeout: float = 5.0) -> None:
    """Signal the health-check thread to stop and clean up."""
    self._shutdown_event.set()
    self._health_thread.join(timeout=timeout)
    if self._health_connection:
        try:
            self._health_connection.close()
        except Exception:
            pass
```

---

#### 3. **Lock Timeout of 0 Still Causes Immediate Failure** (`postgres_cache.py:265-269`)

`lock_timeout` uses `NonNegativeInt` which allows 0. If configured as 0, every cache operation fails immediately.

**Suggestion**: Either:
- Use `PositiveInt` for `lock_timeout` in config, OR
- Handle 0 specially to mean "infinite wait":

```python
def _acquire_lock(self) -> None:
    timeout = self._lock_timeout if self._lock_timeout > 0 else None
    if not self._tx_lock.acquire(timeout=timeout):
        raise CacheError("lock acquisition timeout")
```

---

#### 4. **Startup Race: First Health Check May Increment Failures** (`postgres_cache.py:222-249`)

On first loop iteration, `_health_connection` is `None`, triggering `_connect_health()`. If DB is unreachable at startup, this increments `_consecutive_failures` immediately, potentially triggering liveness failure before any real operations have been attempted.

This may be acceptable behavior, but worth documenting.

---

### Minor Observations

#### 5. **Potential Duplicate Writes on Retry** (Unchanged)

This is documented as a known trade-off. The advisory lock helps within a transaction, but the `@connection` retry could still cause duplicates if the commit succeeds but connection drops immediately after. Consider adding a note in the docstring.

---

### Summary

The code is much improved! The main remaining issues are:

1. **Health connection state** - should be set to `None` before reconnect attempt
2. **`shutdown()` cleanup** - should join thread and close connections
3. **Lock timeout edge case** - 0 value handling

None of these are blockers, but fixing #1 and #2 would improve robustness.

xrajesh:
@coderabbit review

sriroopar:
## Adversarial Review — [OLS-3221](https://redhat.atlassian.net/browse/OLS-3221) PostgreSQL Auto-Recovery

### 🟡 Thread-Safety: Double-Counting Inflates `_consecutive_failures`

Both the `@connection` decorator (via `_mark_unhealthy()`) and the health-check loop independently increment `_consecutive_failures`. During a DB outage:

```
Thread A (request):  _mark_unhealthy() → failures = 1
Thread B (health):   health check fails → failures = 2
Thread A (request):  retry fails, _mark_unhealthy() → failures = 3  (premature 503!)
```

With `liveness_db_failure_threshold=3` (default), two real failures + one co-fire = premature `503 Service Unavailable`.

**Suggestion**: Either dedupe by tracking last-failure timestamp, or have request-path failures delegate to the health loop rather than directly incrementing.

---

### 🟡 `_mark_healthy()` Called Before Retry Succeeds

In `_handle_connection_error()` (postgres.py:33-45):

```python
_do_reconnect(connectable, func_name)  # reconnect succeeds
if hasattr(connectable, "_mark_healthy"):
    connectable._mark_healthy()  # ← marked healthy NOW
try:
    return f(connectable, *args, **kwargs)  # ← but retry could still fail
```

If reconnect succeeds but the retry fails with another `OperationalError`, liveness incorrectly reports healthy until the next health tick (up to 30s).

**Suggestion**: Call `_mark_healthy()` only after the retry invocation succeeds.

---

### 🟡 `ready()` Lags Up to 30s on Recovery

```python
def ready(self) -> bool:
    with self._health_lock:
        return self._health_status
```

This returns the background loop's *last-known* status. With `health_check_interval=30` (default), `ready()` can return `False` for up to 30 seconds after the DB recovers. During this window, Kubernetes readiness probes may restart the pod unnecessarily.

**Mitigation options**:
- Lower default `health_check_interval` (e.g. 10s)
- Perform an on-demand probe inside `ready()` if status is unhealthy (with rate-limiting)
- Document that recovery lag is expected and safe

---

### 🔵 Nit — `_suppress_health_loop` Fixture Applied Twice

In `tests/unit/cache/test_postgres_cache.py`:

```python
pytestmark = pytest.mark.usefixtures("_suppress_health_loop")  # line 14

@pytest.fixture(autouse=True)  # lines 17-20
def _suppress_health_loop() -> Generator[None, None, None]:
    ...
```

Both `autouse=True` and `pytestmark` apply the fixture to all tests — one is redundant. Remove the `pytestmark` line since `autouse=True` already covers it.

---

### 🔵 Nit — E2E Test Changes Appear Out of Scope

`tests/e2e/test_query_endpoint.py` and `test_streaming_query_endpoint.py` add assertions checking for refusal-language patterns. These don't appear related to Postgres auto-recovery. If they fix a regression, that should be a separate PR/commit for traceability.

---

### ✅ Core Design Confirmed Sound

- **No deadlock in `_handle_connection_error`**: The lock acquisition path is correct — reconnect holds `_tx_lock` only if present.
- **Health thread shutdown is correct**: Uses `_shutdown_event.set()` + `join(timeout)` with connection cleanup in `finally`.
- **`_acquire_lock` with bounded timeout**: Prevents indefinite hangs — raises `CacheError` on timeout.
- **Ambiguous commit handling**: Correctly wraps commit failures in `CacheError` rather than silently retrying writes.

---

### Verdict

**Needs changes** —  The thread-safety and recovery-lag issues are medium-severity and should be addressed or documented as known limitations.

sriroopar:
## Follow-up: Fixes for Medium Issues

I've addressed the three medium issues from the adversarial review:

### 1. ✅ Fixed: Double-counting of `_consecutive_failures`

**Problem**: Both the `@connection` decorator (via `_mark_unhealthy()`) and the health-check loop independently incremented `_consecutive_failures`, causing premature 503s.

**Fix**: `_mark_unhealthy()` now only sets `_health_status = False` without incrementing the counter. Only the health-check loop is the sole authority for `_consecutive_failures`:

```python
def _mark_unhealthy(self) -> None:
    """Mark health status as unhealthy (from cache operations).

    Only sets status to False without incrementing the failure counter.
    The health-check loop is the sole authority for consecutive_failures
    to avoid double-counting between request threads and the health loop.
    """
    with self._health_lock:
        self._health_status = False
```

### 2. ✅ Fixed: `_mark_healthy()` called before retry succeeds

**Problem**: In `_handle_connection_error()`, `_mark_healthy()` was called right after reconnect, before the retry operation. If the retry failed, liveness incorrectly reported healthy.

**Fix**: Moved `_mark_healthy()` call to after the retry succeeds in the `@connection` decorator:

```python
try:
    result = f(connectable, *args, **kwargs)
    # Only mark healthy AFTER the retry succeeds
    if hasattr(connectable, "_mark_healthy"):
        connectable._mark_healthy()
    return result
except (psycopg2.OperationalError, psycopg2.InterfaceError) as retry_err:
    raise CacheError(...)
```

### 3. ✅ Fixed: `ready()` lags up to 30s on recovery

**Problem**: `ready()` only returned the health loop's cached status, which could be stale for up to 30 seconds after DB recovery.

**Fix**: Added on-demand probe in `ready()` when status is unhealthy, with rate-limiting (5 seconds) to avoid hammering the DB:

```python
def ready(self) -> bool:
    with self._health_lock:
        if self._health_status:
            return True
        # Rate-limited on-demand probe when unhealthy
        now = time.monotonic()
        if now - self._last_ready_probe < 5.0:
            return False
        self._last_ready_probe = now

    # Do on-demand probe outside the lock
    try:
        if self._health_connection is None or self._health_connection.closed:
            self._connect_health()
        with self._health_connection.cursor() as cursor:
            cursor.execute("SELECT 1")
        # Probe succeeded — mark healthy
        with self._health_lock:
            self._health_status = True
            self._consecutive_failures = 0
        return True
    except Exception:
        return False
```

All unit tests pass (49 postgres tests + 15 health endpoint tests).

sriroopar:
changes merged on another pr

PR #2975, opened 2026-06-30
test: trigger CI to verify main Konflux build

onmete:
Empty commit to verify whether the Konflux `lightspeed-service-on-pull-request` build passes on current `main`. Will close after CI reports.

Made with [Cursor](https://cursor.com)


onmete:
Confirmed main is broken — same uv venv --seed failure. Fix is in PR #2973.

PR #2980, opened 2026-07-01
OLS-3327 - Bedrock tests

JoaoFula:
## Description

add bedrock test files and adaptations to existing scripts

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for Bedrock Anthropic and DeepSeek models, including tool-calling scenarios.
  * Expanded coverage for direct IAM and role-based authentication, including smoke tests.
  * Added configuration fixtures for filtering, authentication, logging, and data collection.
  * Updated invalid-question checks to accept additional valid responses.

* **Documentation**
  * Updated Bedrock testing guidance, model coverage, configuration examples, and authentication options.

* **Chores**
  * Improved automated setup of Bedrock IAM credentials for end-to-end testing.

### DIFF
--- a/tests/e2e/test_query_endpoint.py
+++ b/tests/e2e/test_query_endpoint.py
@@ -45,7 +45,7 @@ def test_invalid_question():
         # at least acknowledges that query is non-ocp.
         # Below assert is minimal due to model randomness.
         assert re.search(
-            r"(sorry|questions|assist)",
+            r"(sorry|questions|assist|help)",
             json_response["response"],
             re.IGNORECASE,
         )
@@ -75,7 +75,7 @@ def test_invalid_question_without_conversation_id():
         # Randomness in response is expected.
         # assert json_response["response"] == INVALID_QUERY_RESP
         assert re.search(
-            r"(sorry|questions|assist)",
+            r"(sorry|questions|assist|help)",
             json_response["response"],
             re.IGNORECASE,
         )
--- a/tests/e2e/test_query_endpoint.py
+++ b/tests/e2e/test_query_endpoint.py
@@ -45,7 +45,7 @@ def test_invalid_question():
         # at least acknowledges that query is non-ocp.
         # Below assert is minimal due to model randomness.
         assert re.search(
-            r"(sorry|questions|assist|help)",
+            r"(sorry|questions|assist|can('t|not) help)",
             json_response["response"],
             re.IGNORECASE,
         )
@@ -75,7 +75,7 @@ def test_invalid_question_without_conversation_id():
         # Randomness in response is expected.
         # assert json_response["response"] == INVALID_QUERY_RESP
         assert re.search(
-            r"(sorry|questions|assist|help)",
+            r"(sorry|questions|assist|can('t|not) help)",
             json_response["response"],
             re.IGNORECASE,
         )
--- a/tests/e2e/test_streaming_query_endpoint.py
+++ b/tests/e2e/test_streaming_query_endpoint.py
@@ -73,7 +73,7 @@ def test_invalid_question():
         response_utils.check_content_type(response, constants.MEDIA_TYPE_TEXT)
 
         assert re.search(
-            r"(sorry|questions|assist)",
+            r"(sorry|questions|assist|help)",
             response.text,
             re.IGNORECASE,
         )
--- a/tests/e2e/test_streaming_query_endpoint.py
+++ b/tests/e2e/test_streaming_query_endpoint.py
@@ -73,7 +73,7 @@ def test_invalid_question():
         response_utils.check_content_type(response, constants.MEDIA_TYPE_TEXT)
 
         assert re.search(
-            r"(sorry|questions|assist|help)",
+            r"(sorry|questions|assist|can('t|not) help)",
             response.text,
             re.IGNORECASE,
         )
--- a/tests/scripts/test-e2e-cluster-periodics.sh
+++ b/tests/scripts/test-e2e-cluster-periodics.sh
@@ -57,6 +57,21 @@ function run_suites() {
     run_suite "rhaiis_vllm" "not azure_entra_id and not certificates and not (tool_calling and not smoketest and not rag) and not byok1 and not byok2 and not quota_limits and not data_export" "rhaiis_vllm" "$RHAIIS_PROVIDER_KEY_PATH" "meta-llama/Llama-3.1-70B-Instruct" "$OLS_IMAGE" "default"
     (( rc = rc || $? ))
 
+    # Bedrock suites — PROVIDER_KEY_PATH carries a discriminator ("iam" or
+    # "iam_role") instead of a credential file path; see
+    # ensure_bedrock_iam_secret() in ols_installer.py.
+    run_suite "bedrock_anthropic" "not azure_entra_id and not certificates and not (tool_calling and not smoketest and not rag) and not byok1 and not byok2 and not quota_limits and not data_export" "bedrock_anthropic" "iam" "anthropic.claude-sonnet-4-6" "$OLS_IMAGE" "default"
+    (( rc = rc || $? ))
+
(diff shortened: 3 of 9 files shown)


JoaoFula:
test_invalid_question failed because deepseek rejected answering "how to make a burger" in a different way from any other provider.

JoaoFula:
failure was from known flakiness being investigated in parallel. retesting

blublinsky (commented):
`ensure_bedrock_iam_secret` can silently fail: unlike other providers where `create_secrets` always creates `llmcreds` (even with `creds == "empty"`), for Bedrock this function IS the only path that creates `llmcreds` — so a silent skip (when env vars are missing) means no credentials exist at all after `create_secrets` returns.

Consider raising instead of returning when the required env vars are absent — the test will fail anyway, and an explicit error with the missing variable names is more helpful than a pod crash minutes later when the operator cannot mount the secret.

tests/e2e/test_query_endpoint.py
```diff
@@ -45,7 +45,7 @@ def test_invalid_question():
         # at least acknowledges that query is non-ocp.
         # Below assert is minimal due to model randomness.
         assert re.search(
-            r"(sorry|questions|assist)",
+            r"(sorry|questions|assist|help)",
```
blublinsky: Adding bare `help` here applies to every provider suite, not just DeepSeek. Prefer a `PROVIDER`-gated pattern (or a tighter phrase like `can't help`) rather than weakening this assert globally.

Same change is also at line 78 in this file and in `test_streaming_query_endpoint.py`.

tests/e2e/test_query_endpoint.py
```diff
@@ -75,7 +75,7 @@ def test_invalid_question_without_conversation_id():
         # Randomness in response is expected.
         # assert json_response["response"] == INVALID_QUERY_RESP
         assert re.search(
-            r"(sorry|questions|assist)",
+            r"(sorry|questions|assist|help)",
```
blublinsky: Same as above — global `help` weakens the invalid-question check for non-DeepSeek providers.

blublinsky (commented):
### Invalid-question regex: avoid global `help`

Agree that DeepSeek may need different refusal wording, but adding bare `help` to the shared pattern weakens the check for **all** providers. A non-refusal like "I can help you make a burger" would now pass.

Prefer scoping the DeepSeek quirk instead of diluting every suite:
1. Gate on `PROVIDER` (e.g. allow `help` only for `bedrock_deepseek`), or
2. Use a tighter DeepSeek-oriented phrase if refusals are consistent (e.g. `can't help` / `cannot help`).

tests/e2e/test_streaming_query_endpoint.py
```diff
@@ -73,7 +73,7 @@ def test_invalid_question():
         response_utils.check_content_type(response, constants.MEDIA_TYPE_TEXT)
 
         assert re.search(
-            r"(sorry|questions|assist)",
+            r"(sorry|questions|assist|help)",
```
blublinsky: Same as the query-endpoint asserts — scope the DeepSeek wording quirk instead of adding bare `help` for all providers.

tests/scripts/test-e2e-cluster.sh
```diff
@@ -64,8 +64,25 @@ function run_suites() {
   run_suite "rhelai_vllm" "smoketest" "rhelai_vllm" "$OPENAI_PROVIDER_KEY_PATH" "gpt-3.5-turbo" "$OLS_IMAGE" "default"
   (( rc = rc || $? ))
 
+  # Bedrock suites — PROVIDER_KEY_PATH carries a discriminator ("iam" or
+  # "iam_role") instead of a credential file path; see
+  # ensure_bedrock_iam_secret() in ols_installer.py.
```
blublinsky, line 77: **Nit (non-blocking):** Good call-site note. Consider also updating the script header (line 5) so it no longer claims `PROVIDER_KEY_PATH` is always a credentials file, and list the required `BEDROCK_*` env vars.

tests/scripts/test-e2e-cluster-periodics.sh
```diff
@@ -57,6 +57,21 @@ function run_suites() {
     run_suite "rhaiis_vllm" "not azure_entra_id and not certificates and not (tool_calling and not smoketest and not rag) and not byok1 and not byok2 and not quota_limits and not data_export" "rhaiis_vllm" "$RHAIIS_PROVIDER_KEY_PATH" "meta-llama/Llama-3.1-70B-Instruct" "$OLS_IMAGE" "default"
     (( rc = rc || $? ))
 
+    # Bedrock suites — PROVIDER_KEY_PATH carries a discriminator ("iam" or
+    # "iam_role") instead of a credential file path; see
+    # ensure_bedrock_iam_secret() in ols_installer.py.
```
blublinsky, line 70: **Nit (non-blocking):** Same as `test-e2e-cluster.sh` — call-site comment is clear; the script header still describes a file path only.

blublinsky (commented):
### Nit (non-blocking): document Bedrock `PROVIDER_KEY_PATH` overload in script headers

The new Bedrock call sites correctly note that `PROVIDER_KEY_PATH` carries `"iam"` / `"iam_role"` instead of a file path. The script **headers** (line 5 in both `test-e2e-cluster.sh` and `test-e2e-cluster-periodics.sh`) still say it is only a credentials file path, and they do not list the required `BEDROCK_AWS_*` / `BEDROCK_ROLE_*` env vars.

Docs polish only — not a blocker.

blublinsky:
## Review note: OpenAI-prefix Bedrock coverage gap

**Priority:** must-fix (vs [OLS-3327](https://redhat.atlassian.net/browse/OLS-3327) / `.ai/spec/how/e2e-bedrock.md`)

### What the AC/spec ask for

Bedrock `load()` has two distinct model-prefix branches that e2e was meant to cover:

1. **`anthropic.*`** → `ChatBedrockConverse`
2. **`openai.*`** → `ChatOpenAI` with `base_url={mantle}/openai/v1` and `use_responses_api=True`

The planned second family was `bedrock_openai` (e.g. `openai.gpt-5.4`), plus iam_role smoke and tool_calling for that family.

### What this PR does

| Suite | Model | Path exercised |
|-------|--------|----------------|
| `bedrock_anthropic` | `anthropic.claude-sonnet-4-6` | Anthropic / Converse ✅ |
| `bedrock_deepseek` | `deepseek.v3.2` | ChatOpenAI on `{url}/v1`, `use_responses_api=False` |

`deepseek.v3.2` does **not** start with `openai.`, so it never enters the OpenAI-prefix branch in `ols/src/llms/providers/bedrock.py` (`/openai/v1` + Responses API). It hits the default ChatOpenAI path instead — related, but not the branch named in the AC.

### Why it matters

- Anthropic coverage is fine (sonnet vs opus is a model choice on the same path).
- DeepSeek still validates Bedrock + IAM + a ChatOpenAI-style client.
- It does **not** e2e-prove Mantle’s OpenAI-compatible Responses endpoint wiring. A regression there would not be caught by these suites.

This is a **coverage / AC mismatch**, not a broken test run (`e2e-ols-cluster` already passed with DeepSeek).

### How to resolve

Either:

1. **Add** `bedrock_openai` (+ iam_role smoke + tool_calling) with an `openai.*` model as in the spec, **or**
2. **Keep DeepSeek** and **update** [OLS-3327](https://redhat.atlassian.net/browse/OLS-3327) AC + `.ai/spec/how/e2e-bedrock.md` to document DeepSeek / generic `/v1` ChatOpenAI as the second family, and mark `openai.*` Responses coverage as out of scope or deferred.

Until one of those is done, merge either ships incomplete AC or leaves the written plan out of date.

---
*AI-assisted review finding — please validate before treating as a blocking request.*

PR #2994, opened 2026-07-10
ols-3523: no longer download embedding models but keep them in the repository

raptorsun:
To avoid an EC exception of downloading external files.

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change
- [x] Konflux configuration change

- Related Issue # [OLS-3523](https://redhat.atlassian.net/browse/OLS-3523)
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added support for packaging embedding models into compressed chunks during builds.
  * Embedding models are now bundled directly for offline runtime use.

* **Bug Fixes**
  * Added validation to ensure reconstructed embedding model files are valid before completion.
  * Removed obsolete generic prefetch configuration.

* **Chores**
  * Removed outdated model artifact lock entries and download-based runtime fallback logic.


Makefile
```diff
@@ -152,10 +152,19 @@ verify-packages-completeness:	requirements.txt ## Verify that requirements.txt f
 
 get-embeddings: get-embeddings-byok get-embeddings-okp ## Download all embedding model binaries
 
-get-embeddings-byok: ## Download BYOK/FAISS embedding model binary
+get-embeddings-byok: ## Download and split BYOK/FAISS embedding model binary
 	curl --fail --location --retry 3 --retry-delay 2 --connect-timeout 15 \
 		-o embeddings_model/all-mpnet-base-v2/model.safetensors \
 		"https://huggingface.co/sentence-transformers/all-mpnet-base-v2/resolve/main/model.safetensors"
+	cd embeddings_model/all-mpnet-base-v2 && \
+		tar czf - model.safetensors | split -b 95M - model.safetensors.tar.gz.
```
xrajesh, line 161: we can remove the original file model.safetensors , after split.

Containerfile
```diff
@@ -67,6 +67,16 @@ RUN echo "Verifying dependencies installation..." && \
     TIKTOKEN_CACHE_DIR=/app-root/.tiktoken_cache python -c "import tiktoken; tiktoken.get_encoding('cl100k_base')" && \
     echo "All dependencies installed and verified successfully!"
 
+# Step 5: Reassemble embedding model safetensors from compressed chunks and validate.
+COPY --chmod=775 embeddings_model ./embeddings_model
+RUN for model_dir in all-mpnet-base-v2 granite-embedding-30m-english; do \
+      cat "embeddings_model/${model_dir}"/model.safetensors.tar.gz.* | \
+        tar xzf - --no-same-owner -C "embeddings_model/${model_dir}" && \
```
xrajesh: We can have a error log if the tar fails here - than catching all.

xrajesh (commented):
Over all looks good - I think you tried the git lfs option too .

Makefile
raptorsun: we can keep the tensor file for local testing.

raptorsun:
> Over all looks good - I think you tried the git lfs option too .

yes, i find complains about missing git LFS support in konflux, so I changed to this split - assembly approach.

Containerfile
raptorsun: corrected. now it emits error message when `tar` command fails

Makefile
blublinsky, line 161: `split` does not remove pre-existing output files. If a model is re-split after shrinking (fewer chunks than a prior run), stale `model.safetensors.tar.gz.*` files remain. Since the Containerfile reassembles via `cat … model.safetensors.tar.gz.*`, those leftovers would be concatenated and corrupt the extracted model.

Suggest adding cleanup before each split:

```makefile
cd embeddings_model/all-mpnet-base-v2 && \
    rm -f model.safetensors.tar.gz.* && \
    tar czf - model.safetensors | split -b 95M - model.safetensors.tar.gz.
```

Same for the `get-embeddings-okp` target below.

blublinsky (commented):
Minor review comment on the Makefile split targets.

PR #3000, opened 2026-07-13
OLS-3548 Omit temperature for models that deprecate it

thoraxe:
## Summary

- Adds `temperature_supported` boolean flag to `ModelParameters` (defaults to `True` for backward compatibility)
- The base `LLMProvider` strips `temperature` from LLM params when the model's config sets `temperature_supported: false`
- Fixes HTTP 400 errors from models like `claude-sonnet-5` that have deprecated the `temperature` parameter

## How to use

In `olsconfig.yaml`, set `temperature_supported: false` on any model that rejects `temperature`:

```yaml
models:
  - name: anthropic.claude-sonnet-5
    parameters:
      temperature_supported: false
```

Models without this setting (or with `temperature_supported: true`) continue to receive `temperature` as before.

## Test plan

- [x] New unit tests verify temperature is stripped when `temperature_supported=False`
- [x] New unit tests verify temperature is still present when `temperature_supported=True` (default)
- [x] New unit tests verify caller-supplied temperature is also stripped when not supported
- [x] All 98 existing LLM provider unit tests pass
- [x] All 210 config model tests pass
- [x] mypy type check passes on modified files

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to indicate whether a model supports temperature controls.
  * Automatically omit `temperature` for models that do not support it; supported models retain existing behavior.

* **Documentation**
  * Updated the example configuration to show how to disable temperature support for specific models.

* **Tests**
  * Added coverage for the new option and temperature handling across configuration scenarios.

### DIFF
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -13,6 +13,7 @@ llm_providers:
         #   reasoning_effort: low      # low | medium | high (default: low)
         #   reasoning_summary: concise  # auto | concise | detailed (default: concise)
         #   verbosity: low             # low | medium | high (default: low)
+        #   temperature_supported: true  # set to false for models that reject temperature (default: true)
   - name: my_azure_openai
     type: azure_openai
     url: "https://myendpoint.openai.azure.com/"
--- a/examples/olsconfig.yaml
+++ b/examples/olsconfig.yaml
@@ -13,7 +13,6 @@ llm_providers:
         #   reasoning_effort: low      # low | medium | high (default: low)
         #   reasoning_summary: concise  # auto | concise | detailed (default: concise)
         #   verbosity: low             # low | medium | high (default: low)
-        #   temperature_supported: true  # set to false for models that reject temperature (default: true)
   - name: my_azure_openai
     type: azure_openai
     url: "https://myendpoint.openai.azure.com/"
@@ -60,6 +59,10 @@ llm_providers:
       # use fully qualified Bedrock model IDs (e.g. anthropic.claude-opus-4-7,
       # openai.gpt-5.4, deepseek.v3.1)
       - name: anthropic.claude-opus-4-7
+      # For models that reject the temperature parameter, set temperature_supported: false:
+      # - name: anthropic.claude-sonnet-5
+      #   parameters:
+      #     temperature_supported: false
   # - name: my_bedrock_iam
   #   type: bedrock
   #   url: "https://bedrock-mantle.us-east-1.api.aws"
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -62,6 +62,7 @@ class ModelParameters(BaseModel):
     reasoning_effort: ReasoningLevel = ReasoningLevel.LOW
     reasoning_summary: ReasoningSummary = ReasoningSummary.CONCISE
     verbosity: ReasoningLevel = ReasoningLevel.LOW
+    temperature_supported: bool = True
 
     @field_validator("tool_budget_ratio")
     @classmethod
--- a/ols/src/llms/providers/provider.py
+++ b/ols/src/llms/providers/provider.py
@@ -374,6 +374,13 @@ class LLMProvider(AbstractLLMProvider):
             )
             updated_params = {**updated_params, **config.dev_config.llm_params}
 
+        # Temperature stripping is intentionally applied after all param merges
+        # (including dev_config overrides) because it enforces a physical model
+        # constraint: if the model's API does not accept a temperature argument,
+        # passing one causes an API error regardless of how it was configured.
+        # This is not a precedence decision — dev_config still wins over defaults
+        # and call-site params — it is a hard capability check that must happen
+        # last, after the final merged param set is known.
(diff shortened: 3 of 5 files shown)


thoraxe:
Addressed the CodeRabbit nitpick — removed the dead `getattr`/`None` guard in `_model_supports_temperature`. `ModelConfig.parameters` is declared with a default `ModelParameters()`, so it can never be `None`. Simplified to access `model_config.parameters.temperature_supported` directly.

All 22 bedrock provider tests pass.

xrajesh:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

---

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

xrajesh:
### Code review

Found 2 issues:

1. **Parameter override precedence violated** - Temperature stripping is applied AFTER developer config overrides in `_override_params`, contradicting the documented precedence order that states "config params overrides everything" (line 371-375). If a user sets `dev_config.llm_params.temperature`, it will still be stripped if the model doesn't support temperature, violating the documented contract.

https://github.com/openshift/lightspeed-service/blob/fee421e101365c9e4a601fe319a742ea9f3fafcf/ols/src/llms/providers/provider.py#L370-L379

2. **Commits should be squashed** - PR contains 2 commits that are not independent (second commit removes dead code from first). CLAUDE.md requires one logical commit per PR unless changes are explicitly independent.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

xrajesh:
**Additional note (lower priority):** The new `temperature_supported` field in `ModelParameters` should have test coverage in `tests/unit/app/models/test_config.py` to verify the default value and explicit setting behavior, following the pattern established for other `ModelParameters` fields like `reasoning_effort`, `reasoning_summary`, and `verbosity`. The provider-level tests are good but model-level validation is missing.

xrajesh:
But @thoraxe - we will need the operator changes to go along with this change right - the user will have to configure this in the CR.

thoraxe:
No operator changes are needed — `temperature_supported` is just another field in `ModelParameters` alongside `reasoning_effort`, `reasoning_summary`, and `verbosity`. The operator already passes model parameters through to the service config, so no CR schema change is required. The default is `true`, so existing deployments are unaffected.

This is purely a bug fix: models that reject temperature were getting it sent anyway. A user _could_ configure it, but that's already possible through the existing parameters passthrough.

thoraxe:
**Re: xrajesh's concern about temperature stripping overriding dev_config precedence**

The behavior is intentional and not a bug — here's the reasoning:

`dev_config` *does* win over defaults and call-site params (the merge order is preserved). But temperature stripping that happens afterward is not a precedence decision — it is a **physical capability enforcement**. If the model's API does not accept a temperature argument, sending one will cause an API error regardless of which layer configured it. The strip has to happen last, after the final merged param set is known, because only at that point do we know all sources have had their say.

The analogy: `dev_config` can override anything about *how the model is called*, but it cannot override *what the model's API physically accepts*. That is a constraint imposed by the model vendor, not by our config system.

To make this intent clear to future readers, I added an inline comment in `_override_params` (commit `80b9f27f`) explaining exactly this. The CodeRabbit `monkeypatch` fix for `test_temperature_stripped_when_set_via_dev_config` is included in the same commit — the manual reset at the end of that test has been removed, so test isolation is now guaranteed even on assertion failure.

All 23 bedrock tests pass.

thoraxe:
Rebased onto current main — the branch was cleanly replayable with no conflicts, which resolves the `needs-rebase` state. All 182 tests in `test_bedrock.py` and `test_config.py` pass locally after the rebase.

xrajesh:
`The operator already passes model parameters through to the service config, so no CR schema change is required. ` -- @thoraxe from your comment above. PTAL https://github.com/openshift/lightspeed-operator/blob/362bce7446e37aeb9100b89cfef0edea59e99f4b/api/v1alpha1/olsconfig_types.go#L529 -- It cannot pass arbitrary list of parameters. so , for a model where I don't want to pass temperature , I have no way to set `temperature_supported` from the CR.

xrajesh:
approving with comment above.
/lgtm
/approve

thoraxe:
You're right, and I was wrong above. I checked `ModelParametersSpec` in the operator (`olsconfig_types.go`) and none of `temperature_supported`, `reasoning_effort`, `reasoning_summary`, or `verbosity` are actually exposed through the CR — only `maxTokensForResponse` and `toolBudgetRatio` are. So there's no way to set `temperature_supported: false` via the operator CR today; that only works if you're hand-editing `olsconfig.yaml` directly.

That's a pre-existing gap in the operator (it's never surfaced the full `ModelParameters` set), not something this PR introduces — the default (`true`) preserves current behavior, so I don't think it blocks this fix. But it's a real gap and worth a follow-up on the operator side. Thanks for catching it.

PR #3001, opened 2026-07-13
ols-3523: konflux: update python package index to avoid EC exceptions of wild binary wheels

raptorsun:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [x] Configuration Update
- [x] CI configuration change
- [x] Konflux configuration change

- Related Issue # https://redhat.atlassian.net/browse/OLS-3523
- Closes #

## Testing

## Summary by CodeRabbit

* **New Features**
  * Updated the Konflux hermetic build workflow to generate hash-pinned dependency artifacts under `.konflux/`, using CPU-profile driven resolution and RHOAI-first package selection to feed Tekton prefetch.
* **Documentation**
  * Added/updated guidance for regenerating hermetic requirements and following the hermetic build process.
* **Chores**
  * Refreshed build/tooling pins, adjusted linting/verifier exclusions, bumped builder/runtime base image, and updated Tekton task bundle/pipeline pins used during prefetch/build stages.


AGENTS.md
```diff
@@ -16,12 +16,7 @@ OpenShift LightSpeed (OLS) is an AI-powered assistant service for OpenShift buil
...
-  1. **Do NOT run `make konflux-requirements`** — full regeneration causes version drift that breaks hermeto prefetch
-  2. Instead, surgically append hash entries for new packages: `uv pip compile <pkg-file> --refresh --generate-hashes --python-version 3.12 --no-deps --no-annotate`
-  3. Put packages in `requirements.hashes.source.txt` (PyPI source) unless they match RHOAI index versions exactly
-  4. Check `requirements-build.txt` for build deps (e.g. `uv-build`, `hatchling`) — bump if a new package needs a newer version
-  5. Check transitive deps — if a new package pulls in deps not already in either hash file, add those too
+- **Konflux hermetic build** - The Konflux build installs from `.konflux/requirements.hashes.source.txt`, `.konflux/requirements.hashes.wheel.txt`, and `.konflux/requirements.hashes.wheel.pypi.txt`, NOT from `uv.lock`. Regenerate with `make konflux-requirements` (runs `scripts/konflux_resolve.py --profile cpu`). Requires the custom uv fork (`syedriko/uv` branch `syedriko-prefer-index`) on PATH or pointed to via `UV_BINARY` env var. Config lives in `.konflux/profiles.toml`.
```
vimalk78: this is a hard external dependency devs will need to run `make konflux-requirements` 
can we do without it

.konflux/profiles.toml
```diff
@@ -0,0 +1,12 @@
+[common]
```
vimalk78, line 1: add a README.md for the new folder?

vimalk78 (commented):
update commit message and PR description to describe the changes

AGENTS.md
raptorsun: when python dependency upgrades, we run make konflux-requirements to update the wheel list for konflux.
when there is a new RPM package installed by dnf or microdnf in the container, we run make konflux-rpm-lock

.konflux/profiles.toml
raptorsun: readme added to .konflux folder

AGENTS.md
raptorsun: I added a document on upgrading dependencies here docs/update-requirements.md

pyproject.toml
```diff
@@ -2,6 +2,7 @@
 
 # description of all rules are available on https://docs.astral.sh/ruff/rules/
 lint.select = ["D", "E", "F", "W", "C", "S", "I", "G", "TCH", "SLOT", "RUF", "C90", "N", "YTT", "ASYNC", "A", "C4", "T10", "FURB", "PERF", "AIR", "NPY", "FLY", "PLW2901"]
```
blublinsky, line 4: The script passes ruff, black, pylint, and mypy cleanly today — zero violations. These exclusions can be removed since they add no value now and will only prevent detection of future regressions if someone else modifies the script.

blublinsky (commented):
Review of the resolve script and build pipeline changes.

scripts/konflux_resolve.py
```diff
@@ -0,0 +1,1209 @@
...
+    v = parse_version(version)
+    for raw_part in constraint.split(","):
+        part = raw_part.strip()
+        if not part:
+            continue
+        op, ver_str = _parse_specifier(part)
+        if not _check_single_specifier(version, v, op, ver_str):
+            return False
+    return True
+
```
blublinsky, line 69: ~170 lines of hand-rolled regex for PEP 440 version parsing and PEP 508 marker evaluation. The `packaging` library is already in the dependency tree and provides `Version`, `SpecifierSet`, and `Marker` that handle all edge cases (epochs, dev+pre combos, parenthesized marker groups) that this regex cannot:

```python
from packaging.version import Version
from packaging.specifiers import SpecifierSet
from packaging.markers import Marker

# Replaces parse_version + version_satisfies + all the regex:
Version("1.2.3.post1") in SpecifierSet(">=1.2,<2.0")

# Replaces _eval_marker + _eval_single_marker:
Marker('sys_platform == "linux"').evaluate({"sys_platform": "linux"})
```

Would eliminate ~170 lines and be more correct and maintainable.

scripts/konflux_resolve.py
```diff
@@ -0,0 +1,1209 @@
...
+_UV_COMPILED_RE = re.compile(r"^([a-zA-Z0-9][a-zA-Z0-9._-]*)([=<>!~].*)?$")
+
+
+UV_BINARY = os.environ.get("UV_BINARY", "uv")
+
+
+def uv_resolve(
+    python_version: str,
+    rhoai_index_url: str,
```
blublinsky: Dependency on a custom `uv` fork (`syedriko/uv`, branch `syedriko-prefer-index`) for `--index-strategy prefer-index` is a significant maintenance risk:

- Requires Rust toolchain to build from source
- Must be kept in sync with upstream uv releases
- Single-person bus factor

This can be avoided using stock `uv`. The script already scrapes the full RHOAI index via `RhoaiIndex.load()` — it knows every available package and version. Instead of relying on a custom flag:

1. Auto-generate a comprehensive overrides file from the RHOAI index (pin every RHOAI package to its latest version)
2. Run stock `uv pip compile --override <auto-generated-overrides>` — no custom flag needed
3. Use `--emit-index-annotation` to verify classification

The existing `requirements.overrides.txt` is already doing this manually for ~24 packages. Automating it for all RHOAI packages would eliminate the fork dependency entirely.

scripts/konflux_resolve.py
```diff
@@ -0,0 +1,1209 @@
...
+        "https://pypi.org/simple",
+    )
+
+    # Step 6: Build dependencies via pybuild-deps
+    sdist_names = list(buckets["pypi_sdist"].keys())
+    build_output = os.path.join(KONFLUX_DIR, f"requirements-build{suffix}.txt")
+    if sdist_names:
+        tmp_sdist_file = os.path.join(KONFLUX_DIR, f"_tmp_sdist_list{suffix}.txt")
+        try:
+            with open(tmp_sdist_file, "w") as f:
```
blublinsky: `griffe`, `griffecli`, `griffelib` appear twice in the generated Tekton binary packages list — they're in both `bootstrap_packages` and the resolved RHOAI wheel bucket. Use `sorted(set(...))`:

```python
wheel_package_names = sorted(set(
    list(buckets["rhoai_wheel"].keys())
    + list(buckets["pypi_wheel"].keys())
    + [normalize_name(p) for p in bootstrap_packages]
))
```

scripts/konflux_resolve.py
```diff
@@ -0,0 +1,1209 @@
+#!/usr/bin/env python3
```
blublinsky, line 1: 1209 lines with BFS dependency resolution, version constraint matching, network I/O, wheel compatibility checks, and file generation — with zero tests. At minimum the pure functions (`parse_version`, `version_satisfies`, `normalize_name`, `is_wheel_compatible`) would benefit from unit tests, especially since this is the build pipeline's single point of failure.

pyproject.toml
raptorsun: Done — removed all linter exclusions (ruff, black, pylint). The script now passes all linters cleanly: black (unchanged), ruff (all checks passed), pylint (9.97/10).

scripts/konflux_resolve.py
raptorsun: Done — wrapped in `sorted(set(...))` to deduplicate.

scripts/konflux_resolve.py
raptorsun: Good point. The script was ported from lightspeed-core which intentionally avoids external dependencies (stdlib-only). Switching to `packaging` would be cleaner — will track as a follow-up.

scripts/konflux_resolve.py
raptorsun: Agreed on the maintenance risk. The auto-generate-overrides approach is a solid alternative. For this PR we're aligning with the existing lightspeed-core tooling (same fork, same script pattern). Eliminating the fork dependency is worth a dedicated follow-up.

scripts/konflux_resolve.py
raptorsun: Fair point. The pure functions (`parse_version`, `version_satisfies`, `normalize_name`, `is_wheel_compatible`) are good candidates for unit tests. Will add in a follow-up PR.

scripts/konflux_resolve.py
raptorsun: Addressed — replaced ~130 lines of hand-rolled PEP 440 version parsing and PEP 508 marker evaluation with `packaging.version.Version`, `packaging.specifiers.SpecifierSet`, and `packaging.markers.Marker`. Script went from 1209 to 1077 lines, same output.

scripts/konflux_resolve.py
raptorsun: Addressed — the script now uses a two-pass resolution with stock `uv`:

1. **Pass 1**: `uv pip compile` with manual overrides → initial dependency tree
2. **Auto-generate overrides**: load RHOAI index, pin every RHOAI-available package to its latest version (skips packages already in manual overrides)
3. **Pass 2**: re-resolve with manual + auto overrides → final resolution
4. **Classify**: check each resolved package against RHOAI index (not `# from` annotations)

The custom uv fork (`--index-strategy prefer-index`) is no longer required. `UV_BINARY` env var removed. Works with stock `uv`.

Note: without the RHOAI index in the uv command, torch resolves from PyPI which pulls nvidia-* CUDA deps (17 extra wheel-only packages). These are harmless — the RHOAI CPU torch is installed at build time — but the output is slightly larger than with the fork.

raptorsun: new commit pushed including the change to eliminate dependency on customized `uv` fork

scripts/konflux_resolve.py
raptorsun: refactored using `packaging` libarary :)

blublinsky (changes requested):
`patch_tekton_packages` silently succeeds when the regex doesn't match.

If the Tekton YAML format changes (quoting, line splitting, etc.), `re.sub` returns the original content unchanged, the file is rewritten as-is, and the caller logs `"Patched %s"` — reporting success when nothing happened. The binary packages list silently falls out of sync, and `prefetch-dependencies` fails in Konflux with no indication from the script that the patching was a no-op.

Suggestion — check whether a substitution occurred:

```python
new_content = re.sub(r'"packages":\s*"[^"]*"', replacement, content)
if new_content == content:
    raise RuntimeError(f"No 'packages' pattern found in {yaml_path}")
```

`scripts/konflux_resolve.py:718-728`

blublinsky:
One more minor item: `_run_uv_compile` captures stdout/stderr but doesn't surface them on failure. When `uv pip compile` fails, the user sees Python's generic `CalledProcessError` traceback without the actual error message from uv (e.g. "no solution found for package X").

Suggestion:

```python
try:
    result = subprocess.run(cmd, capture_output=True, text=True, check=True)
except subprocess.CalledProcessError as exc:
    logger.error("uv pip compile failed:\n%s", exc.stderr)
    raise
```

`scripts/konflux_resolve.py:799-801`

raptorsun:
thank you for reviewing twice :) @blublinsky  both points addressed in new commit.

raptorsun:
/override "ci/prow/ols-evaluation"
/override "ci/prow/e2e-ols-cluster"
unrelated to this change

raptorsun (commented):
Both addressed — thanks for the catches.

1. `patch_tekton_packages` now raises `RuntimeError` if the regex doesn't match (no silent no-ops).
2. `_run_uv_compile` now logs `exc.stderr` before re-raising, so the actual uv error is visible.

raptorsun:
Fixed — `_run_uv_compile` now surfaces the uv stderr on failure.

PR #3016, opened 2026-07-22
OLS-3540 Add MCP semantic convention attributes on tool spans

vimalk78:
## Summary

- Add OTel GenAI semantic convention attributes (`gen_ai.operation.name`, `gen_ai.tool.name`, `gen_ai.tool.call.id`, `gen_ai.tool.type`) to all tool execution spans
- Add MCP-specific attributes (`mcp.method.name`, `network.transport`) to MCP-sourced tool spans, with conditional `mcp.session.id` and `mcp.protocol.version` when available
- Capture MCP server metadata (server name, transport type) during tool gathering in `mcp_utils.py` and propagate via `tool.metadata`
- Map MCP transport types to OTel `network.transport` OSI-layer values (`streamable_http` → `tcp`, `sse` → `tcp`, `stdio` → `pipe`)
- Capture tool output content in `tool.result` span event, gated by `capture_content` flag (per spec rule 3: full fidelity when audit is enabled)

## Test plan

- [x] Unit tests: new tests covering MCP tool span attributes, non-MCP tool span attributes, MCP metadata capture, and tool result content capture
- [x] `make test-unit` — 1169 tests pass
- [x] `make format` / `make verify` — clean
- [x] In-cluster: verified spans in Jaeger show correct `gen_ai.*` and `mcp.*` attributes on tool execution spans, and `tool.result` event with output content

image

### DIFF
--- a/ols/src/tools/tools.py
+++ b/ols/src/tools/tools.py
@@ -530,19 +530,29 @@ async def _execute_single_tool_call_stream(
     """
     tool_id, tool_args, tool = tool_call
     tool_name = getattr(tool, "name", "unknown")
-    mcp_server = (getattr(tool, "metadata", None) or {}).get("mcp_server", "")
+    metadata = getattr(tool, "metadata", None) or {}
+    mcp_server = metadata.get("mcp_server", "")
+
+    span_attrs: dict[str, str] = {
+        "gen_ai.operation.name": "execute_tool",
+        "gen_ai.tool.name": tool_name,
+        "gen_ai.tool.call.id": tool_id,
+        "gen_ai.tool.type": "function",
+    }
+    if mcp_server:
+        span_attrs["mcp.method.name"] = "tools/call"
+        mcp_transport = metadata.get("mcp_transport", "")
+        if mcp_transport:
+            span_attrs["network.transport"] = mcp_transport
+        session_id = metadata.get("mcp_session_id", "")
+        if session_id:
+            span_attrs["mcp.session.id"] = session_id
+        protocol_version = metadata.get("mcp_protocol_version", "")
+        if protocol_version:
+            span_attrs["mcp.protocol.version"] = protocol_version
 
     tool_span = (
-        audit_ctx.span(
-            f"execute_tool {tool_name}",
-            **{
-                "gen_ai.operation.name": "execute_tool",
-                "gen_ai.tool.name": tool_name,
-                "gen_ai.tool.call.id": tool_id,
-                "gen_ai.tool.type": "function",
-            },
-            mcp_server=mcp_server,
-        )
+        audit_ctx.span(f"execute_tool {tool_name}", **span_attrs, mcp_server=mcp_server)
         if audit_ctx
         else nullcontext()
     )
@@ -582,6 +592,7 @@ async def _execute_single_tool_call_stream(
                 output_length=len(tool_output),
                 success=status == "success",
                 duration_ms=duration_ms,
+                output_content=tool_output if audit_ctx.capture_content else None,
             )
 
         yield _tool_result_event(
--- a/ols/utils/audit_logger.py
+++ b/ols/utils/audit_logger.py
@@ -146,13 +146,18 @@ class AuditLogger:
         output_length: int,
         success: bool,
         duration_ms: Optional[int],
+        output_content: Optional[str] = None,
     ) -> None:
-        """Set tool result attributes on execute_tool span."""
+        """Set tool result attributes and emit tool.result event on execute_tool span."""
         self._set_span_attrs(
             output_length=output_length,
             success=success,
             duration_ms=duration_ms,
         )
+        attrs: dict[str, Any] = {"success": success}
+        if output_content is not None:
+            attrs["output"] = output_content
+        self._add_span_event("tool.result", **attrs)
 
     def tool_approval_requested(
         self,
--- a/ols/utils/mcp_utils.py
+++ b/ols/utils/mcp_utils.py
@@ -32,6 +32,18 @@ ClientHeaders: TypeAlias = dict[str, dict[str, str]]
 MCPServersDict: TypeAlias = dict[str, MCPServerTransport]
 
 
+_MCP_TO_NETWORK_TRANSPORT: dict[str, str] = {
+    "streamable_http": "tcp",
+    "sse": "tcp",
+    "stdio": "pipe",
+}
+
+
+def _mcp_transport_to_network(mcp_transport: str) -> str:
+    """Map MCP transport type to OTel network.transport value (OSI layer)."""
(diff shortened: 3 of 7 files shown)


vimalk78 (commented):
**Re: CodeRabbit comments**

1. `test_mcp_metadata_capture` rename — will fix, minor naming improvement.
2. `test_otel.py` changes — this file is from PR #3007 (cherry-picked as a dependency), not part of this PR's changes. Skipping.

vimalk78:
## Adversarial self-review

Ran an adversarial review before merging. Findings and disposition:

### Fixed

1. **`network.transport` could be empty string** (`tools.py:544`) — When MCP server config lacked a `transport` key, `_mcp_transport_to_network("")` returned `""`, producing a meaningless `network.transport=""` span attribute. Now guarded with `if mcp_transport:`, matching the pattern used for `mcp.session.id` / `mcp.protocol.version`.

2. **`_mcp_transport_to_network` had no direct unit test** (`mcp_utils.py:41`) — Only tested indirectly via `test_gather_mcp_tools_metadata_capture`. Added parametrized test covering `streamable_http→tcp`, `sse→tcp`, `stdio→pipe`, unknown passthrough, and empty string.

### Intentionally skipped

3. **Spec says `network.transport` = `stdio` / `sse`, code maps to OSI values (`tcp`, `pipe`)** — Code is correct per OTel semantic conventions. Spec update is a separate change.

4. **Unbounded tool output in span event** — Spec rule 3 says full fidelity when audit is enabled. No truncation by design.

5. **`mcp_session_id` / `mcp_protocol_version` never populated** — Intentional future-proofing for when MCP client exposes session metadata. Dead paths are harmless and gated behind `if` checks.

6. **Test style: MagicMock vs real OTel fixture** — Mock approach is valid for verifying argument forwarding to `audit_ctx.span()`. Existing `TestExecuteToolGenAISpans` class covers real OTel span creation.

7. **`capture_content=False` untested through tools.py** — The `audit_logger.test_tool_result_no_content` test covers the `output_content=None` path at the logger layer.

blublinsky (commented):
## Review notes

Two should-fix items around keeping `.ai/spec/what/audit-logging.md` aligned with this change. See inline comments.

ols/utils/mcp_utils.py
```diff
@@ -32,6 +32,18 @@ class MCPServerTransport(TypedDict, total=False):
...
+_MCP_TO_NETWORK_TRANSPORT: dict[str, str] = {
+    "streamable_http": "tcp",
+    "sse": "tcp",
+    "stdio": "pipe",
+}
+
+
+def _mcp_transport_to_network(mcp_transport: str) -> str:
+    """Map MCP transport type to OTel network.transport value (OSI layer)."""
+    return _MCP_TO_NETWORK_TRANSPORT.get(mcp_transport, mcp_transport)
```
blublinsky, line 44: **should-fix:** Spec/code mismatch on `network.transport` values.

This mapping correctly emits OTel OSI-layer values (`tcp` / `pipe`) per the [OTel MCP semantic conventions](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/mcp.md).

However, `.ai/spec/what/audit-logging.md` rule 11 still documents Recommended values as `stdio` or `sse`. Please update that rule in this PR (or land a linked follow-up before merge) so the governing audit spec matches what we ship.

Suggested wording: Recommended values `tcp` / `pipe` (mapped from MCP transports `streamable_http`/`sse` → `tcp`, `stdio` → `pipe`).

ols/utils/audit_logger.py
```diff
@@ -146,13 +146,18 @@ def tool_result(
...
+        """Set tool result attributes and emit tool.result event on execute_tool span."""
         self._set_span_attrs(
             output_length=output_length,
             success=success,
             duration_ms=duration_ms,
         )
+        attrs: dict[str, Any] = {"success": success}
+        if output_content is not None:
+            attrs["output"] = output_content
+        self._add_span_event("tool.result", **attrs)
```
blublinsky, line 160: **should-fix:** Document `tool.result` (and content gating) in the audit spec.

This adds a `tool.result` span event with optional `output`, gated by `capture_content` at the call site in `tools.py`. That behavior is not covered by `.ai/spec/what/audit-logging.md` today:

- Span Events (rules 13–14a) only specify `gen_ai.choice` / LLM content
- Rule 14a only mentions completion/thinking under `capture_content`

Also, the PR description cites "spec rule 3: full fidelity when audit is enabled," which does not match current rule 3 (`user_id`).

Please extend the audit spec to require/allow `tool.result`, clarify that tool `output` is Opt-In via `audit.capture_content`, and drop or correct the rule-3 citation.

ols/src/tools/tools.py
```diff
@@ -530,19 +530,29 @@ async def _execute_single_tool_call_stream(
...
+        span_attrs["mcp.method.name"] = "tools/call"
+        mcp_transport = metadata.get("mcp_transport", "")
+        if mcp_transport:
+            span_attrs["network.transport"] = mcp_transport
+        session_id = metadata.get("mcp_session_id", "")
+        if session_id:
+            span_attrs["mcp.session.id"] = session_id
+        protocol_version = metadata.get("mcp_protocol_version", "")
+        if protocol_version:
+            span_attrs["mcp.protocol.version"] = protocol_version
```
blublinsky, line 552: **nice-to-have:** Nested MCP attribute `if`s push `_execute_single_tool_call_stream` from radon B(8) → C(13).

The block is still readable for this PR size, but extracting something like `_mcp_span_attrs(metadata) -> dict[str, str]` would keep complexity down and isolate the optional session/protocol/transport handling from the main execution path.

Not a blocker.

ols/utils/mcp_utils.py
vimalk78: Fixed — updated `.ai/spec/what/audit-logging.md` rule 11 to document OTel OSI-layer values: `tcp` (mapped from `streamable_http`/`sse`) and `pipe` (mapped from `stdio`).

ols/utils/audit_logger.py
vimalk78: Fixed — added rule 14b documenting `tool.result` span event on `execute_tool` spans (carries `success` boolean, plus `output` when `capture_content` is enabled). Extended rule 14a to cover tool output under the content capture policy.

Re: rule 3 citation — good catch, that was in the PR description not the spec. Rule 3 is `user_id`. The governing requirement is rule 14a (content capture policy) plus the new 14b.

ols/src/tools/tools.py
vimalk78: Acknowledged — keeping as-is for this PR. The nested `if`s add ~3 lines each and follow the same guard pattern used throughout `_execute_single_tool_call_stream`. Happy to extract `_mcp_span_attrs()` if complexity grows with future MCP attributes.

PR #3019, opened 2026-07-23
OLS-3541 Add gen_ai.* Prometheus histogram metrics

vimalk78:
## Summary

- Add 3 new Prometheus histograms aligned with OTel GenAI semantic conventions:
  - `gen_ai_client_token_usage` — token consumption by model/provider/type (input/output)
  - `gen_ai_client_operation_duration_seconds` — LLM call latency by model/provider/operation
  - `gen_ai_execute_tool_duration_seconds` — tool execution latency by tool name
- Existing `ols_*` metrics are untouched — new metrics run in parallel
- Token histogram observed in `TokenMetricUpdater.__exit__()`
- LLM duration histogram observed in `_invoke_llm()` with try/finally to capture on timeout/GeneratorExit
- Tool duration histogram observed in `_execute_single_tool_call_stream()`

## Test plan

- [x] Unit tests: 3 new tests covering token histogram, LLM duration histogram, and tool duration histogram
- [x] `make test-unit` — 1150 tests pass
- [x] `make format` / `make verify` — clean

## Adversarial review

One round of adversarial review completed:

**Finding 1 (CRITICAL — fixed):** LLM duration histogram was observed after the `async for` loop in `_invoke_llm()`. On timeout or `GeneratorExit`, the code after `async for` is silently skipped, so the histogram would never be observed. Fixed with try/finally pattern wrapping the entire async generator loop.

**Finding 2 (IMPORTANT — fixed):** Tool duration used `int(elapsed * 1000)` for the histogram, losing sub-millisecond precision. Fixed to use raw float `elapsed` for the histogram `.observe()` call, keeping the int conversion only for the audit log `duration_ms` field.

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for GenAI token usage, LLM operation duration, and tool execution duration.
  * Token metrics distinguish input and output usage by model, provider, and operation.
  * LLM and tool timings are recorded consistently, including failed operations, improving monitoring visibility.
  * Added tailored metric buckets to better represent token usage and long-running operations.

* **Tests**
  * Added coverage validating token, LLM duration, and tool execution metrics.


blublinsky:
**Should-fix: `gen_ai_client_operation_duration_seconds` missing `error_type` label**

The OTel GenAI semantic conventions define `error.type` as **conditionally required** on `gen_ai.client.operation.duration` — it must be present when the operation ends in error.

Currently the duration is recorded in `finally` with the same label set for both success and failure. This means a 300s timeout observation (from `asyncio.timeout(300)` in `_collect_round_llm_chunks`) gets mixed into the same histogram as normal 2–5s responses, inflating p99/p999 and making dashboards misleading. There's also no way to filter `rate(…_count)` by success vs error for availability queries.

Suggested fix — add `error_type` label:

```python
# In metrics.py — add label to histogram definition:
gen_ai_client_operation_duration_seconds = Histogram(
    "gen_ai_client_operation_duration_seconds",
    ...,
    ["gen_ai_request_model", "gen_ai_provider_name", "gen_ai_operation_name", "error_type"],
)

# In llm_execution_agent.py — set label based on outcome:
error_type = ""
try:
    async for chunk in chain.astream(...):
        yield chunk
except Exception as exc:
    error_type = type(exc).__name__
    raise
finally:
    gen_ai_client_operation_duration_seconds.labels(
        ..., error_type=error_type
    ).observe(elapsed)
```

blublinsky:
**Nice-to-have: Remove `if input_tokens:` / `if output_tokens:` guards on histogram observations**

In `token_counter.py`, the `gen_ai_client_token_usage` histogram observations are guarded by `if input_tokens:` / `if output_tokens:`. Unlike `Counter.inc(0)` (a genuine no-op), `Histogram.observe(0)` is meaningful — it increments `_count` by 1 and places the observation in the `le="1.0"` bucket.

Skipping zero observations means `rate(gen_ai_client_token_usage_count[5m])` would under-report LLM call throughput in edge cases (streaming timeout, early LLM error).

Practical impact is minor, but the fix is trivial — just remove the guards:

```python
gen_ai_client_token_usage.labels(
    gen_ai_token_type="input", **labels
).observe(input_tokens)
gen_ai_client_token_usage.labels(
    gen_ai_token_type="output", **labels
).observe(output_tokens)
```

blublinsky:
**Nice-to-have: `"LLM invocation completed"` debug log now fires on error path too**

The PR moved `logger.debug("LLM invocation completed: ...")` into the `finally` block alongside the histogram observation. Before the PR, this log only ran on success (it sat after the `try/except`, and `except` re-raises). Now on failure you get both `logger.error("LLM invocation failed")` and then `logger.debug("LLM invocation completed")`, which is misleading.

Suggested fix — use `try/except/else/finally` so only the histogram goes into `finally`:

```python
try:
    async for chunk in chain.astream(...):
        yield chunk
except Exception:
    logger.error("LLM invocation failed: ...")
    raise
else:
    logger.debug("LLM invocation completed: ...")
finally:
    elapsed = time.monotonic() - llm_start_time
    gen_ai_client_operation_duration_seconds.labels(...).observe(elapsed)
```

vimalk78:
Addressed review comments from @blublinsky:

**1. `error_type` label (Should-fix):** The spec at `.ai/spec/what/observability.md` defines `gen_ai.client.operation.duration` with labels `gen_ai.request.model`, `gen_ai.provider.name`, `gen_ai.operation.name` — no `error_type`. Adding it would be a spec change and needs its own ticket. Out of scope for this PR.

**2. Remove `if input_tokens:` / `if output_tokens:` guards (Nice-to-have):** Fixed. `Histogram.observe(0)` is meaningful — it increments `_count` and populates the `le="1.0"` bucket. Guards removed.

**3. `"LLM invocation completed"` in `finally` (Nice-to-have):** Fixed. Moved the success log to `else` block so it only fires on success. Histogram observation stays in `finally` to capture both success and error durations.

---

Re: CodeRabbit comments on `error_type` label and per-LLM-call token observations — both are beyond the current spec. The `error_type` label is a spec change (needs separate ticket), and per-call token histograms would require restructuring how `TokenMetricUpdater` works. Both are valid improvements but out of scope here.

vimalk78:
## Adversarial review

One round of adversarial review completed:

**Finding 1 (CONFIRMED — fixed):** Duration histogram buckets for both `gen_ai_client_operation_duration_seconds` and `gen_ai_execute_tool_duration_seconds` used Prometheus defaults that max out at 10s. Streaming LLM calls routinely take 30-120+ seconds, so most observations would land in `+Inf` with zero granularity — making `histogram_quantile()` useless for SLO dashboards. Fixed with custom buckets: LLM `(1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120, 180)`, tool `(0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120)`.

**Finding 2 (PLAUSIBLE — fixed):** `TokenMetricUpdater.__exit__` unconditionally observed the token histogram, recording `observe(0)` on LLM failure. Unlike Counters where `.inc(0)` is a no-op, histogram observations are real data points that would pollute p50/p99 statistics. Fixed by guarding with `if exc_type is None:`.

**Finding 3 (CONFIRMED — fixed):** Token histogram had 5 unreachable upper bucket boundaries (262K–67M tokens) far exceeding any model's context window (~200K max). Each creates one time series per label combination. Fixed by capping power-of-4 buckets at 65536.

ols/app/metrics/metrics.py
```diff
@@ -49,6 +49,32 @@
...
+gen_ai_client_token_usage = Histogram(
+    "gen_ai_client_token_usage",
+    "Per-request token usage distribution (OTel GenAI semantic conventions)",
+    [
+        "gen_ai_operation_name",
+        "gen_ai_token_type",
+        "gen_ai_request_model",
+        "gen_ai_provider_name",
+    ],
+    buckets=(1, 4, 16, 64, 256, 1024, 4096, 16384, 65536),
```
blublinsky, line 61: **should-fix:** Token-usage buckets stop at `65536`, but `.ai/spec/what/observability.md` rule 1 (and OTel GenAI v1.41) require continuing through `67108864`:

`[1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864]`

I saw the adversarial-review rationale (cardinality / context-window), but large prompts on modern long-context models will all collapse into `+Inf` and lose distribution resolution.

Please either restore the full OTel/spec list, or update `.ai/spec/what/observability.md` to document this intentional truncation.

ols/app/metrics/token_counter.py
```diff
@@ -162,3 +164,17 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
         llm_reasoning_token_total.labels(provider=self.provider, model=self.model).inc(
             self.token_counter.token_counter.reasoning_tokens
         )
+        if exc_type is None:
```
blublinsky: **should-fix:** Success-only gating under-reports real token usage vs sibling metrics.

`ols_llm_*` counters above always update, and LLM duration observes in `finally`, but `gen_ai_client_token_usage` only runs when `exc_type is None`.

On hard LLM failure or client disconnect (`GeneratorExit`) after partial multi-round work, accumulated tokens from completed rounds are omitted from the distribution. Avoiding `observe(0)` pollution is reasonable, but prefer observing whenever tokens were actually counted (e.g. `llm_calls > 0` or tokens > 0), not only on clean exit.

ols/app/metrics/metrics.py
```diff
@@ -49,6 +49,32 @@
...
+)
+
+gen_ai_client_operation_duration_seconds = Histogram(
+    "gen_ai_client_operation_duration_seconds",
+    "LLM inference call duration (OTel GenAI semantic conventions)",
+    ["gen_ai_request_model", "gen_ai_provider_name", "gen_ai_operation_name"],
+    buckets=(1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120, 180),
```
blublinsky, line 68: **nice-to-have:** Duration buckets intentionally diverge from OTel GenAI advisory boundaries (`0.01…81.92`) in favor of longer streaming-friendly ranges. That tradeoff is understandable, but the metric descriptions still say “OTel GenAI semantic conventions.”

Please document the intentional choice in `.ai/spec/what/observability.md` (or note the divergence in the metric help text) so consumers don’t assume OTel-default bucket alignment.

Same note applies to `gen_ai_execute_tool_duration_seconds` below.

blublinsky (commented):
## Review notes

Two should-fix items and one nice-to-have around histogram buckets / observation gating. See inline comments.

ols/app/metrics/metrics.py
vimalk78: the adverserial review i did for myself wanted to change the bucket to stop at 65536

vimalk78: Updated `.ai/spec/what/observability.md` to document the truncated bucket list `[1, 4, 16, 64, 256, 1024, 4096, 16384, 65536]` with rationale: buckets above 65536 exceed any current model's per-request token count and create unused time series per label combination.

ols/app/metrics/token_counter.py
vimalk78: Good point — fixed. Changed guard from `if exc_type is None` to `if input_tokens or output_tokens`. Now observes the histogram whenever tokens were actually counted (including partial failure after completed rounds), but still skips truly-zero observations. Added test `test_token_metric_updater_observes_histogram_on_failure_with_tokens` to verify.

ols/app/metrics/metrics.py
vimalk78: Updated `.ai/spec/what/observability.md` to document the custom bucket boundaries inline with each metric, including rationale: LLM duration extends to 180s for streaming calls (OTel advisory maxes at ~82s); tool duration starts at 0.1s for sub-second local tools and extends to 120s for long MCP calls.

.ai/spec/what/observability.md
```diff
@@ -18,9 +18,9 @@ The service exposes Prometheus metrics, records conversation transcripts and use
...
+   | `gen_ai.client.operation.duration` | Histogram | `gen_ai.request.model`, `gen_ai.provider.name`, `gen_ai.operation.name` | LLM inference call duration. Bucket boundaries: [1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120, 180] (custom range for streaming LLM calls that routinely take 30–120s; OTel advisory boundaries max at ~82s which loses granularity for long-running inferences). Unit: `s`. |
+   | `gen_ai.execute_tool.duration` | Histogram | `gen_ai.tool.name` | Tool execution duration. Bucket boundaries: [0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120] (custom range covering sub-second local tools through long-running MCP calls). Unit: `s`. |
```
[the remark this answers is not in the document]
vimalk78: The spec uses OTel semantic convention names (dotted) which is standard for specs. Prometheus `prometheus_client` automatically converts dots to underscores in metric/label names. This is documented behavior and doesn't need a separate mapping — any Prometheus consumer already knows this convention.

tests/unit/app/metrics/test_token_counter.py
```diff
@@ -163,3 +163,74 @@ def test_token_metric_updater_reports_reasoning_tokens():
...
+    updater.__exit__(RuntimeError, RuntimeError("LLM failed"), None)
+
+    after = gen_ai_client_token_usage.labels(
+        gen_ai_operation_name="chat",
+        gen_ai_token_type="input",  # noqa: S106
+        gen_ai_request_model="fail_mod",
+        gen_ai_provider_name="fail_prov",
+    )._sum.get()
+
+    assert after - before == 200.0
```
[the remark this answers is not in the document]
vimalk78: The failure-path test is intentional. Per reviewer feedback from @blublinsky, `if exc_type is None` was changed to `if input_tokens or output_tokens` — the histogram now observes whenever tokens were actually counted, even on failure (e.g. partial failure after completed LLM rounds in a multi-turn tool loop). Only truly-zero observations are skipped. This avoids under-reporting vs the sibling `ols_llm_token_*` counters which always update.

blublinsky (commented):
## Non-blocking notes

**Nit 1 — token vs duration granularity (non-blocking)**

`gen_ai_client_token_usage` is observed once in `TokenMetricUpdater.__exit__` around the whole multi-round tool loop (summed tokens), while `gen_ai_client_operation_duration_seconds` is observed per `_invoke_llm` call. A multi-round tool loop therefore yields N duration samples but 1 token sample.

This is fine if “Per-request” in the observability spec means agent/user request (which matches the current wrap). It only matters if dashboards assume OTel-style 1:1 pairing of token usage with duration under `operation_name=chat`.

Suggestion: either leave as-is and briefly clarify in `.ai/spec/what/observability.md` that token usage is per agent-request aggregate (not per LLM round), or later observe per LLM call if you want true per-operation distributions. Not a merge blocker.

**Nit 2 — tool duration mixes success and failure (non-blocking)**

`gen_ai_execute_tool_duration_seconds` is observed for both successful and failed tool runs (`status=error` from `_execute_with_retries`) on the same series (label is only `gen_ai_tool_name`). Same class of gap as LLM `error.type`, which was deferred to match the current spec.

Optional follow-up: add an error/success discriminator if you want clean p99s for successful tool latency. Not blocking.

vimalk78:
Re: review https://github.com/openshift/lightspeed-service/pull/3019#pullrequestreview-4896704469

**Nit 1 — token vs duration granularity:** Good callout. Clarified in `.ai/spec/what/observability.md` that `gen_ai.client.token.usage` is "per-request (agent-request aggregate, not per-LLM-round)." A multi-round tool loop yields 1 token sample (summed) but N duration samples — this matches our "per-request" semantics. Per-LLM-round token histograms would require restructuring `TokenMetricUpdater` and is a potential follow-up.

**Nit 2 — tool duration mixes success/failure:** Agreed, same class of gap as the LLM `error.type` label which was deferred to match current spec scope. Will address both together in a follow-up if we add error discriminators.

PR #3029, opened 2026-07-27
OLS-3718: fix quota limits test

JoaoFula:
## Description

1. Defer Solr version resolution into background thread
  - runner.py: Moved config.solr_hybrid_search access from the main startup path
  (where it blocked uvicorn for up to 120s) into load_index(), which runs in an
  existing background thread. Uvicorn now starts immediately, and /readiness stays
  503 until both Solr resolution and RAG index loading complete.

  2. Require all pod containers ready in e2e wait helper
  - tests/e2e/utils/cluster.py: Replaced the fragile minimum-count container check
  (1 for normal, 2 for tool-calling) with all(s == "true" for s in statuses).
  This ensures RHOKP/Solr sidecars are fully ready before e2e tests proceed, not
  just the API container

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup so Solr hybrid search configuration is resolved during background index loading, reducing main-thread blocking and making hybrid resolution non-fatal.
  * Updated service readiness behavior to align with index loading/serving progress.
* **Tests**
  * Refined e2e pod readiness checks to succeed only when all reported container readiness states are `"true"`, removing mode-based thresholding.

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -27,9 +27,8 @@ from ols.version import __version__
 def load_index():
     """Load the index."""
     # Resolve Solr hybrid search (OCP version resolution) in the background
-    # so it does not block uvicorn startup.  The readiness endpoint gates on
-    # rag_index which is loaded right after, so the service stays 503 until
-    # both complete.
+    # so it does not block uvicorn startup.  Solr failure is non-fatal: the
+    # pod goes ready once rag_index loads and serves without hybrid search.
     config.solr_hybrid_search  # pylint: disable=W0104, E0606
     config.rag_index  # pylint: disable=W0104, E0606
 
--- a/runner.py
+++ b/runner.py
@@ -26,8 +26,11 @@ from ols.version import __version__
 
 def load_index():
     """Load the index."""
-    # accessing the config's rag_index property will trigger the loading
-    # of the index
+    # Resolve Solr hybrid search (OCP version resolution) in the background
+    # so it does not block uvicorn startup.  The readiness endpoint gates on
+    # rag_index which is loaded right after, so the service stays 503 until
+    # both complete.
+    config.solr_hybrid_search  # pylint: disable=W0104, E0606
     config.rag_index  # pylint: disable=W0104, E0606
 
 
@@ -99,9 +102,6 @@ if __name__ == "__main__":
             "Pyroscope url is not specified. To enable profiling please set `pyroscope_url` "
             "in the `dev_config` section of the configuration file."
         )
-    # Eagerly initialize Solr hybrid search so OCP version resolution
-    # (requires OCP_CLUSTER_VERSION and a reachable Solr) fails fast.
-    config.solr_hybrid_search  # pylint: disable=W0104
 
     # create and start the rag_index_thread - allows loading index in
     # parallel with starting the Uvicorn server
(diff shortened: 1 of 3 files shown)


xrajesh (commented):
Two critical concerns with moving `solr_hybrid_search` from the main thread into the background `load_index()` thread.

runner.py
```diff
@@ -26,8 +26,11 @@
 
 def load_index():
     """Load the index."""
-    # accessing the config's rag_index property will trigger the loading
-    # of the index
+    # Resolve Solr hybrid search (OCP version resolution) in the background
+    # so it does not block uvicorn startup.  The readiness endpoint gates on
+    # rag_index which is loaded right after, so the service stays 503 until
+    # both complete.
+    config.solr_hybrid_search  # pylint: disable=W0104, E0606
```
xrajesh, line 32: **Critical: readiness does not actually gate on `solr_hybrid_search`**

The comment above says "the service stays 503 until both complete", but the readiness endpoint (`ols/app/endpoints/health.py:63-68`) only checks `config.rag_index is not None` — it never checks `solr_hybrid_search`.

If Solr resolution fails (exception caught → returns `None`), `rag_index` still loads fine, the pod goes ready, and Solr-dependent queries **silently fail at runtime**.

To match the stated intent, either:
- Add a `solr_hybrid_search` check to the readiness endpoint, or
- Acknowledge in the comment that Solr failure is non-fatal and the pod will serve without hybrid search.

runner.py
```diff
@@ -99,9 +102,6 @@ def load_index():
             "Pyroscope url is not specified. To enable profiling please set `pyroscope_url` "
             "in the `dev_config` section of the configuration file."
         )
-    # Eagerly initialize Solr hybrid search so OCP version resolution
-    # (requires OCP_CLUSTER_VERSION and a reachable Solr) fails fast.
-    config.solr_hybrid_search  # pylint: disable=W0104
```
xrajesh, line 104: **Critical: moving this to a background thread changes the failure mode for uncaught exceptions**

The `solr_hybrid_search` property (`ols/utils/config.py:268-279`) wraps the embedding-model resolution in a try/except, but the `SolrHybridSearch(settings, encode_fn)` constructor on line 279 is **outside** that try block.

If the constructor throws:
- **Old behavior (main thread):** process crashes → Kubernetes restarts pod → error is immediately visible in logs.
- **New behavior (background thread):** thread dies silently → `config.rag_index` never executes → pod stuck at 503 **permanently** with no clear error. Kubernetes eventually kills it via liveness timeout, but the root cause is buried.

Consider either:
- Wrapping the entire `solr_hybrid_search` property body in the try/except, or
- Adding a separate try/except around `config.solr_hybrid_search` here in `load_index()` so `rag_index` still loads even if Solr fails.

runner.py
JoaoFula: Fixed — the comment was misleading. The readiness endpoint only gates on `rag_index`, not `solr_hybrid_search`. Updated the comment to state that Solr failure is non-fatal and the pod serves without hybrid search.

runner.py
JoaoFula: Fixed — moved the `SolrHybridSearch(settings, encode_fn)` constructor inside the existing try/except block so any constructor exception is caught, logged, and returns `None` instead of killing the background thread. This ensures `rag_index` always loads regardless of Solr failures.

PR #3033, opened 2026-07-29
OLS-3736 fix IPv6 addresses in no-proxy causing invalid port error

thoraxe:
## Summary

- When `::1` (or any bare IPv6 address) is present in the cluster no-proxy list, the httpx mount-dict key was constructed as `all://*::1`, which httpx misparses — treating `:1` as a port — raising `"Invalid port: ':1'"` and crashing the LLM connection health check.
- Extracts `_no_proxy_mount_key()` in `provider.py` to wrap bare IPv6 addresses (detected by `host.count(":") >= 2`) in brackets per RFC 3986 § 3.2.2, producing `all://[::1]`. `host:port` entries (exactly one colon) and plain hostnames continue to use `all://*<host>`.

## Test plan

- [ ] `make test-unit` passes (1160 tests, 15 new/modified in providers file)
- [ ] New unit tests cover: bare `::1`, bare full IPv6, already-bracketed `[::1]`, `host:port`, hostname, IPv4, CIDR, and mixed-list httpx client construction

Fixes: [OLS-3736](https://redhat.atlassian.net/browse/OLS-3736)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy bypass handling for IPv6 addresses, including loopback and fully qualified formats.
  * Prevented connection setup errors when IPv6 addresses are included in proxy exclusions.
  * Preserved correct behavior for hostnames, IPv4 addresses, CIDR ranges, and hosts with ports.

* **Tests**
  * Added coverage for IPv6 and other proxy exclusion formats.

### DIFF
--- a/ols/src/llms/providers/provider.py
+++ b/ols/src/llms/providers/provider.py
@@ -236,6 +236,30 @@ generic_to_llm_parameters: dict[str, dict[str, str]] = {
 }
 
 
+def _no_proxy_mount_key(host: str) -> str:
+    """Return the httpx mount-dict key for a no-proxy host entry.
+
+    Bare IPv6 addresses (e.g. ``::1``) are ambiguous in URL context: without
+    brackets the last colon-delimited segment is misread as a port number,
+    causing httpx to raise ``Invalid port: ':1'``.  IPv6 addresses are also
+    not domain names and have no subdomains, so they use an exact-match pattern
+    (no wildcard ``*`` prefix) with brackets as required by RFC 3986 §3.2.2.
+    All other entries (hostnames, IPv4 addresses, CIDR ranges) keep the
+    existing ``all://*<host>`` wildcard pattern.
+
+    Args:
+        host: A hostname, IP address, or CIDR range from the no-proxy list.
+
+    Returns:
+        An httpx-compatible URL pattern string for the mounts dict.
+    """
+    if host.count(":") >= 2:
+        # IPv6 address — bracket if bare, use exact-match (no wildcard)
+        bracketed = host if host.startswith("[") else f"[{host}]"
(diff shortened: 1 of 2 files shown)


sriroopar (approved):
Clean, well-scoped IPv6 fix with thorough test coverage — CIDR-before-IPv6 ordering is correct and no regressions.

sriroopar (changes requested):
### Root cause
PR #3033 (fix/OLS-3736-ipv6-no-proxy) branched off main at commit 1430287. Since then, a different PR merged into main:

PR #3033's branch was never rebased onto that change, so:

The pre-existing test at line 164 still calls the old 2-arg form:
 ```
_construct_httpx_client(False, False).
The new regression test this PR adds (test_construct_httpx_client_with_ipv6_in_no_proxy_hosts) 
```

copied that same old 2-arg call pattern. When CI runs the PR's branch merged against current main, both calls now pass 3 positional args (self + 2) into a method that only accepts 2 (self + use_async), producing exactly the TypeError you're seeing. This is also why the PR's own test-plan checkbox for make test-unit is unchecked — the author likely knows it's not green yet.

### Fix : 
Rebase the branch onto latest main, then update both call sites to drop the removed argument:

thoraxe:
Rebased onto current main and fixed the stale `_construct_httpx_client` call site that sriroopar flagged — the new IPv6 regression test was still calling it with the old 2-arg signature (`False, False`) after main dropped the second argument. Updated to `_construct_httpx_client(False)`. All 16 tests in `test_providers.py` pass locally.

sriroopar:
evals failure not caused by changes in this branch. Overriding with the same context.

sriroopar (approved):

PR #3044, opened 2026-08-05
OLS-3797: Simplify TLS cert handling — leverage operator-set SSL_CERT_FILE

blublinsky:
## Description

**Summary**
Remove the certificate generation chain and per-provider certificate store management. The operator already sets SSL_CERT_FILE pointing to a merged CA bundle (ols.pem) before OLS starts — Python's ssl.create_default_context(), httpx, and gRPC all pick it up automatically.

**Changes**

_Deleted:_

ols/utils/certificates.py — generate_certificates_file() is redundant
ols_config.certificate_directory config field
ols_config.extra_ca config field (operator handles CA merging)
provider_config.certificates_store config field
OtelTlsMode enum and audit.otel.tls_mode config field
tests/unit/extra_certs/test_extra_certs.py

_Simplified:_

_construct_httpx_client() — removed use_custom_certificate_store parameter; CA trust via SSL_CERT_FILE
init_tracer() — removed insecure/certificate_file parameters; gRPC CA trust via GRPC_DEFAULT_SSL_ROOTS_FILE_PATH set from SSL_CERT_FILE at startup
build_mcp_config() — removed manual ssl.create_default_context() / httpx_factory; MCP connections use SSL_CERT_FILE automatically
runner.py — removed generate_certificates_file() call, simplified OTEL init

_Bugfix (pre-existing):_

_construct_httpx_client() — mounts (for no_proxy_hosts) was not passed when a TLS security profile was configured, silently ignoring proxy bypass rules
Updated: 12 test files, 2 YAML test configs, 7 spec files, docs/config.puml, README.md

**Result**

-570+ lines removed, zero new dependencies. Backward compatible — existing configs with removed fields are silently ignored (no extra="forbid" on OLSConfig/ProviderConfig).
## Type of change

- [x ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
https://redhat.atlassian.net/browse/OLS-3797
- Closes #
https://redhat.atlassian.net/browse/OLS-3797

## Testing

## Summary by CodeRabbit

* **Configuration**
  * Simplified TLS certificate trust through the `SSL_CERT_FILE` environment variable.
  * Removed legacy certificate-directory, extra CA, provider certificate-store, and OTEL TLS mode settings.
  * OTLP gRPC connections now inherit CA trust automatically at startup.

* **Bug Fixes**
  * Improved consistent proxy handling for outbound HTTP connections.

* **Documentation**
  * Updated setup, security, provider, RAG, audit logging, and configuration guidance for the new certificate trust workflow.


ols/src/llms/providers/provider.py
```diff
@@ -415,28 +417,12 @@ def _construct_httpx_client(  # noqa: C901
...
-                verify = custom_context
-            logger.info(
-                "No security profiles. creating httpx.Client with verify %s", verify
-            )
+            logger.info("No security profiles. creating httpx.Client with verify=True")
             if use_async:
-                return httpx.AsyncClient(verify=verify, proxy=proxy, mounts=mounts)
-            return httpx.Client(verify=verify, proxy=proxy, mounts=mounts)
+                return httpx.AsyncClient(verify=True, proxy=proxy, mounts=mounts)
+            return httpx.Client(verify=True, proxy=proxy, mounts=mounts)
```
vimalk78, line 424: **Undocumented behavior change: hostname verification now enabled.**

The old code explicitly disabled hostname checking when using a custom certificate store:
```python
custom_context.check_hostname = False
```

The new `verify=True` uses Python's default SSL context, which has `check_hostname=True`. This is more secure, but it's a silent behavioral change that could break existing deployments where:
- Self-hosted LLMs are accessed via IP address
- Server certificates are issued to a different hostname than the connection URL
- Dev/test environments use self-signed certs that don't match the hostname

Recommend: at minimum document this in the PR description. Consider whether a release note is warranted for users who may have relied on the old behavior.

runner.py
```diff
@@ -79,25 +72,16 @@ def load_index():
     # init loading of query redactor
     config.query_redactor  # pylint: disable=W0104
 
+    # Let gRPC pick up the same CA bundle as Python ssl via SSL_CERT_FILE
+    os.environ.setdefault(
+        "GRPC_DEFAULT_SSL_ROOTS_FILE_PATH", os.environ.get("SSL_CERT_FILE", "")
```
vimalk78: **Empty-string env var when `SSL_CERT_FILE` is unset.**

When `SSL_CERT_FILE` is not set (e.g., local dev without the operator), this sets `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH` to `""`. gRPC's behavior with an empty-string path is undocumented, and gRPC has prior bugs with empty certs ([grpc#12146](https://github.com/grpc/grpc/issues/12146) — empty root certs silently bypassed verification).

A conditional guard is safer and more explicit:
```python
ssl_cert_file = os.environ.get("SSL_CERT_FILE")
if ssl_cert_file:
    os.environ.setdefault("GRPC_DEFAULT_SSL_ROOTS_FILE_PATH", ssl_cert_file)
```

vimalk78 (changes requested):
## Adversarial Review — OLS-3797 (Score: 77/100)

**4/6 Jira ACs met, 0 failed, 2 need human review.**

Well-executed simplification that removes ~570 lines of manual certificate management. The `mounts` bugfix in the TLS security profile path is a genuine improvement (no_proxy_hosts were silently ignored before). Spec and doc updates are thorough.

### Must-fix

**Undocumented hostname verification behavior change** — see inline comment on `provider.py`. The old code had `check_hostname = False`; the new `verify=True` enables hostname checking. This could break deployments connecting to LLMs via IP address or with hostname-mismatched certs. At minimum this needs to be called out in the PR description; it may also warrant a release note.

### Should-fix

**GRPC env var set to empty string** — see inline comment on `runner.py`. When `SSL_CERT_FILE` is unset, `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH` gets set to `""`. gRPC's behavior with an empty-string path is undocumented and it has prior bugs with empty certs ([grpc#12146](https://github.com/grpc/grpc/issues/12146)). A conditional guard is safer.

### AC note

AC #3 says *"Keep `generate_certificates_file()` for local dev"* but the PR deletes `certificates.py` entirely. The README update provides the alternative (`SSL_CERT_FILE`), so this is arguably the right call — but the Jira AC should be updated to reflect the decision.

blublinsky:
**Re: hostname verification change**

Good catch — the old code had `check_hostname = False` on the custom SSL context, which weakens TLS by allowing any valid CA-signed cert to pass regardless of hostname mismatch. This is a security hole, not an intentional feature.

The new `verify=True` uses Python's default SSL context with `check_hostname = True`, which is the correct secure behavior. The operator-managed cert bundle contains properly-issued certificates with correct hostnames/SANs.

This is an **intentional security improvement**. The only scenario where it could break is a misconfigured deployment connecting to an LLM via raw IP address with a certificate that lacks a SAN for that IP — the fix in that case is to use a proper hostname or add the IP as a SAN, not to disable hostname verification.

Will add a note to the PR description calling this out.

blublinsky:
**Re: empty-string GRPC env var**

Pushing back on this one. If `SSL_CERT_FILE` is not set, we have bigger problems — all HTTPS connections (LLM providers, MCP servers) would fail too, not just gRPC. The operator always sets it; local dev uses system/certifi defaults where `SSL_CERT_FILE` isn't needed at all.

The empty-string case only matters if someone explicitly sets `SSL_CERT_FILE=""`, which is a misconfiguration. Adding a guard here creates a false sense of safety for one protocol while everything else is already broken.

gRPC's C-core falls back to its built-in roots when `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH` is unset, which is the correct behavior for local dev. The `setdefault` ensures we don't override an explicitly set value.

vimalk78 (changes requested):
## Round 2 Adversarial Review — OLS-3797 (Score: 80/100)

**4/6 Jira ACs met, 0 failed, 2 need human review.**

### Round 1 Issues — Status

- **Hostname verification change** — **Resolved.** Author acknowledged as intentional security improvement. Spec updated to mandate hostname checking. No longer a finding.
- **Empty-string GRPC env var** — **Dropped.** Author's pushback is valid: if `SSL_CERT_FILE` is unset, all HTTPS connections fail too, not just gRPC OTEL.

### New Must-Fix

**`_ensure_cert_bundle()` self-inclusion bug** — see inline comment on `runner.py`. The output bundle matches the `.pem` input filter during `cert_dir.iterdir()`, causing linear growth on each pod restart.

### Still Open (human review)

- AC#3: `certificates.py` deleted vs Jira AC saying "Keep for local dev with SSL_CERT_FILE guard." Replaced with `_ensure_cert_bundle()` which serves a different purpose. Valid deviation but contradicts literal AC text — needs team alignment on whether to update the AC.
- AC#6: e2e test pass cannot be verified from code review. New `_ensure_cert_bundle()` has no unit tests.

### Positive

The `mounts=mounts` bugfix in the TLS security profile path is still present and correct — `no_proxy_hosts` were silently ignored before this PR.

runner.py
```diff
@@ -5,24 +5,45 @@
...
+    if not cert_file:
+        return
+    bundle_path = Path(cert_file).resolve()
+    cert_dir = bundle_path.parent
+    if not cert_dir.is_dir() or not str(cert_dir).startswith("/etc/certs"):
+        return
+    parts = [
+        f.read_bytes()
+        for f in sorted(cert_dir.iterdir())
+        if f.suffix in (".crt", ".pem") and f.is_file()
```
vimalk78: **Self-inclusion bug: bundle grows on every pod restart.**

`bundle_path` (e.g. `ols.pem`) has a `.pem` extension and lives in `cert_dir`. On first startup it doesn't exist, so only individual cert files are concatenated. On subsequent restarts, `ols.pem` already exists and matches this filter, so its contents (the previous bundle) are read as input and concatenated again.

Result: `N × original_size` after N restarts — eventually fills the writable volume.

Fix: exclude `bundle_path` from the input files:
```python
parts = [
    f.read_bytes()
    for f in sorted(cert_dir.iterdir())
    if f.suffix in (".crt", ".pem") and f.is_file() and f.resolve() != bundle_path
]
```

vimalk78 (commented):
## Code Review

*4 finding(s) posted as inline comments on the diff.*

### :rotating_light: Critical: AC3 violated: file deleted instead of kept with guard as the story required
[`ols/utils/certificates.py:1`](https://github.com/openshift/lightspeed-service/blob/13b9557ef762ee9a08cac83c350d18386380b0c2/ols/utils/certificates.py#L1)

<details><summary>Detail</summary>

OLS-3797 AC3 explicitly says 'Keep generate_certificates_file() for local dev. Add a guard: skip generation if SSL_CERT_FILE is already set and points to an existing file.' The PR deletes certificates.py entirely and removes extra_ca/certificate_directory from OLSConfig. The replacement (_ensure_cert_bundle in runner.py) has inverted logic — it runs only when SSL_CERT_FILE IS set, doing nothing in the local-dev case. Local-dev cert generation from certifi + extra_ca is completely lost, which is the opposite of what the AC specifies.
</details>

---
*Reviewed by adversarial code review system (5 findings survived verification)*

runner.py
```diff
@@ -75,25 +93,16 @@ def load_index():
     # init loading of query redactor
     config.query_redactor  # pylint: disable=W0104
 
+    # Let gRPC pick up the same CA bundle as Python ssl via SSL_CERT_FILE
+    os.environ.setdefault(
```
vimalk78: :warning: **Warning**: GRPC_DEFAULT_SSL_ROOTS_FILE_PATH set to empty string when SSL_CERT_FILE is absent

os.environ.setdefault("GRPC_DEFAULT_SSL_ROOTS_FILE_PATH", os.environ.get("SSL_CERT_FILE", "")) — when SSL_CERT_FILE is not set, os.environ.get returns "" and the variable is unconditionally written as an empty string. Setting it to "" is semantically different from leaving it unset. gRPC C-core currently tolerates it (checks strlen > 0 before opening the path), but the logic is contrary to intent and fragile against future gRPC changes. Fix: `if ssl_cert := os.environ.get("SSL_CERT_FILE"): os.environ.setdefault("GRPC_DEFAULT_SSL_ROOTS_FILE_PATH", ssl_cert)`

runner.py
```diff
@@ -5,24 +5,45 @@
...
+    in the same directory.
+    """
+    cert_file = os.environ.get("SSL_CERT_FILE")
+    if not cert_file:
+        return
+    cert_dir = Path(os.path.dirname(cert_file)).resolve()
+    if not cert_dir.is_dir():
+        return
+    bundle_name = os.path.basename(cert_file)
+    parts = [
```
vimalk78: :warning: **Warning**: _ensure_cert_bundle() reads its own output file on re-run, duplicating all certs

cert_dir.iterdir() collects all .crt/.pem files in cert_dir, including the bundle output file (e.g. ols.pem, which has .pem suffix) when it already exists. On a second invocation — in-place process restart, non-emptyDir volume, or any non-Kubernetes usage — the previous bundle is concatenated into the new one, doubling every certificate with each run. The docstring assumes an emptyDir (ephemeral on pod restart) but the code has no guard. Fix: add `and f.name != bundle_name` to the list comprehension filter.

ols/app/models/config.py
```diff
@@ -1153,18 +1141,10 @@ class AuditLoggingMode(StrEnum):
     DISABLED = "Disabled"
```
vimalk78, line 1141: :warning: **Warning**: Silent behavioral regression: removing OtelTlsMode silently switches insecure OTEL to TLS

OtelTlsMode enum and tls_mode field on OtelConfig were removed without an AC covering this change. OtelConfig uses Pydantic's default extra='ignore', so users with 'tls_mode: Insecure' in their YAML get no parse error — but insecure=True is no longer passed to OTLPSpanExporter. Their exporter silently switches to TLS mode, which will cause runtime export failures against endpoints that only accept plain-text gRPC. This is a silent breaking change for any deployment that relied on the insecure OTEL path.

runner.py
```diff
@@ -5,24 +5,45 @@
...
 from ols.src.auth.auth import use_k8s_auth
-from ols.utils.certificates import generate_certificates_file
 from ols.utils.environments import configure_gradio_ui_envs, configure_hugging_face_envs
 from ols.utils.logging_configurator import configure_logging
 from ols.utils.otel import init_tracer
 from ols.utils.pyroscope import start_with_pyroscope_enabled
 from ols.version import __version__
 
 
+def _ensure_cert_bundle() -> None:
```
vimalk78, line 27: :bulb: Suggestion: No unit tests for _ensure_cert_bundle(), which replaces fully-tested generate_certificates_file()

generate_certificates_file() had six dedicated tests in tests/unit/extra_certs/test_extra_certs.py (now deleted). The new _ensure_cert_bundle() — which contains the self-inclusion bug identified above — has zero test coverage. tests/unit/runners/ has no test file for runner.py at all. Tests for _ensure_cert_bundle() should cover: SSL_CERT_FILE unset (no-op), cert_dir not a directory (no-op), normal concatenation of .crt and .pem files, and exclusion of the bundle file itself from the iteration.

blublinsky:
test prow/e2e-ols-cluster
/test prow/ols-evaluation

ols/src/llms/providers/provider.py
blublinsky: Intentional. The old `check_hostname = False` was a workaround for our custom cert context that predated the operator-managed CA bundle. With the operator now mounting all required CAs and the bundle containing the full trust chain, TLS verification (including hostname check) should work for all deployments.

Self-hosted LLMs accessed via IP or with mismatched CNs already require a SAN/CN that matches the connection URL in any properly-issued cert — disabling hostname check was papering over mis-configuration.

Added a note to the PR description.

runner.py
blublinsky: Good catch. Fixed — now guarded with:
```python
if ssl_cert := os.environ.get("SSL_CERT_FILE"):
    os.environ.setdefault("GRPC_DEFAULT_SSL_ROOTS_FILE_PATH", ssl_cert)
```

runner.py
blublinsky: Fixed in the refactored implementation. The function now writes to `cert_path` (inside `bundle_dir`) and iterates only *sibling* directories — `bundle_dir` is explicitly skipped with `if d == bundle_dir`. No self-inclusion possible regardless of emptyDir semantics.

runner.py
blublinsky: Duplicate of earlier comment — same fix applied (conditional guard).

runner.py
blublinsky: Duplicate of earlier comment — fixed by the refactored implementation that iterates sibling dirs only and skips `bundle_dir`.

ols/app/models/config.py
blublinsky: The `OtelTlsMode` removal is intentional. The operator **always** configures a TLS endpoint for OTEL — the `Insecure` path was dead code in any operator-managed deployment. The exporter now relies on `GRPC_DEFAULT_SSL_ROOTS_FILE_PATH` (set from `SSL_CERT_FILE`) for CA trust, which is the standard gRPC mechanism.

For non-operator (dev) usage: if no `SSL_CERT_FILE` is set the gRPC env var is simply not set either, so gRPC falls back to its compiled-in roots — which is correct for connecting to public OTEL collectors.

No user ever had `tls_mode: Insecure` in a production CRD (the operator never generated it), and `extra="ignore"` in Pydantic means old configs with the field still parse cleanly.

runner.py
blublinsky: Added `tests/unit/runners/test_cert_bundle.py` with 5 tests covering:
- `SSL_CERT_FILE` unset → `RuntimeError`
- `SSL_CERT_FILE` outside allowed prefix → `RuntimeError`
- Normal concatenation from sibling directories
- Bundle dir excluded from iteration (no self-inclusion)
- Non-cert extensions (`.txt`, `.json`) ignored

vimalk78 (commented):
## AI Review — OLS-3797 (Round 2)

**Score: 90/100** | **Verdict: needs_review** | 5/6 ACs pass, 1 needs human review

### Adherence

**AC#3 — `certificates.py` deleted vs "Keep for local dev"** (NEEDS_REVIEW)

The Jira AC says: *"Keep `generate_certificates_file()` for local dev. Add a guard: skip generation if `SSL_CERT_FILE` is already set and points to an existing file."*

The PR deletes `ols/utils/certificates.py` entirely. The replacement `_ensure_cert_bundle()` in `runner.py` is operator-only — it requires `SSL_CERT_FILE` under `/etc/certs/` and reads from the RHEL system CA bundle at `/etc/pki/tls/certs/ca-bundle.crt`. For local dev, developers must now set `SSL_CERT_FILE` manually (README updated with instructions).

This is arguably the right architectural call (eliminates ~70 lines + `cryptography` dependency usage), but contradicts the literal AC text. **Needs team alignment:** update the AC to match, or restore a local-dev fallback.

All other ACs (mcp_utils cleanup, provider.py cleanup, GRPC env var, otel.py simplification, test updates) are fully met.

### Code Quality

**Positive — bugfix:** `_construct_httpx_client()` now passes `mounts` (no_proxy host bypass) in the TLS security-profile path. Previously, `no_proxy_hosts` were silently ignored when a TLS security profile was configured. Good catch.

**Positive — prior-round fix:** The self-inclusion bug (bundle file re-read on pod restart) is fixed — `bundle_dir` is excluded from cert iteration, confirmed by `test_excludes_bundle_dir_from_iteration`.

No code quality issues found.

---
🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #3055, opened 2026-08-11
OLS-3782 Adding additional logs to capture Azure Entra ID e2e test failures.

sriroopar:
## Summary

- When Azure Entra ID token acquisition fails, `retrieve_access_token()` now raises `LLMConfigurationError` instead of silently returning `None`
- Previously, the `None` token propagated into the `AzureChatOpenAI` client, causing `httpx.RemoteProtocolError: Server disconnected without sending a response` at invocation time
- `raise_missing_attribute_error()` also updated to raise `LLMConfigurationError` for consistency across all config validation paths
- `logger.error()` retained before raising so operators can detect credential failures in log aggregation
- The readiness probe's generic exception handler catches `LLMConfigurationError` and reports the provider as unhealthy (HTTP 503)

## Test plan

- [x] All 19 existing Azure OpenAI unit tests pass
- [x] `test_retrieve_access_token_on_error` asserts `LLMConfigurationError` is raised
- [x] `test_missing_tenant_id`, `test_missing_client_id`, `test_missing_client_secret` updated to assert `LLMConfigurationError`
- [x] `black` and `ruff` pass
- [ ] CI: `test_azure_entra_id` e2e test passes with valid Entra ID credentials

Fixes: https://redhat.atlassian.net/browse/OLS-3782

### DIFF
--- a/ols/src/llms/providers/azure_openai.py
+++ b/ols/src/llms/providers/azure_openai.py
@@ -12,6 +12,7 @@ from langchain_openai import AzureChatOpenAI
 
 from ols import constants
 from ols.app.models.config import AzureOpenAIConfig
+from ols.src.llms.llm_loader import LLMConfigurationError
 from ols.src.llms.providers.provider import LLMProvider
 from ols.src.llms.providers.registry import register_llm_provider_as
 
@@ -111,27 +112,22 @@ class AzureOpenAI(LLMProvider):
         """Load LLM."""
         return AzureChatOpenAI(**self.params)
 
-    def resolve_access_token(self, azure_config: AzureOpenAIConfig) -> Optional[str]:
+    def resolve_access_token(self, azure_config: AzureOpenAIConfig) -> str:
         """Retrieve and cache Azure OpenAI access token."""
         if TOKEN_CACHE.is_expired():
             logger.info(
                 "Cached AD token has expired (or missing) - generating a new one"
             )
             access_token = self.retrieve_access_token(azure_config)
-            if access_token:
-                TOKEN_CACHE.update_token(access_token.token, access_token.expires_on)
-            else:
-                return None
+            TOKEN_CACHE.update_token(access_token.token, access_token.expires_on)
         return TOKEN_CACHE.access_token
 
-    def retrieve_access_token(
-        self, azure_config: AzureOpenAIConfig
-    ) -> Optional[AccessToken]:
+    def retrieve_access_token(self, azure_config: AzureOpenAIConfig) -> AccessToken:
         """Retrieve access token to call Azure OpenAI."""
         if azure_config is None:
-            raise ValueError(
+            raise LLMConfigurationError(
                 "Credentials for API token is not set and "
-                "Azure-specific parameters are not provided."
+                "Azure-specific parameters are not provided. "
                 "It is not possible to retrieve access token."
             )
         if azure_config.tenant_id is None:
@@ -141,7 +137,6 @@ class AzureOpenAI(LLMProvider):
         if azure_config.client_secret is None:
             raise_missing_attribute_error("client_secret")
 
-        # everything is there, try to retrieve credential
         try:
             credential = ClientSecretCredential(
                 azure_config.tenant_id,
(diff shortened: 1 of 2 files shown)


sriroopar:
## Adversarial Review

### What's good

1. The core fix is correct — failing fast with `LLMConfigurationError` instead of silently producing a `None` token eliminates the root cause of the `httpx.RemoteProtocolError`.
2. Using `raise ... from e` preserves the exception chain for debugging.
3. The `resolve_access_token` return type narrowing (`Optional[str]` → `str`) makes the contract honest.

### Issues to address

**1. Inconsistent exception types for missing config attributes (Bug)**

`raise_missing_attribute_error()` at line 155–159 still raises `ValueError`, not `LLMConfigurationError`. This means missing `tenant_id` / `client_id` / `client_secret` raises `ValueError`, while missing `azure_config` raises `LLMConfigurationError`. Both are configuration errors and should raise the same type. The downstream handler in `ols.py:597` only catches `LLMConfigurationError`, so a `ValueError` from missing `tenant_id` would bypass that handler and propagate as an unhandled 500.

The three existing tests (`test_missing_tenant_id`, `test_missing_client_id`, `test_missing_client_secret`) still assert `ValueError` — this works today only because the error is raised during `__init__` before any HTTP endpoint is involved, but it's a latent inconsistency.

**Recommendation:** Change `raise_missing_attribute_error` to raise `LLMConfigurationError` (or at minimum a `ValueError` subclass of it, since `LLMConfigurationError` extends `Exception` not `ValueError`). Update the three tests accordingly.

**2. Lost `logger.error()` call**

The old code had `logger.error("Error retrieving access token: %s", e)` before returning `None`. The new code raises immediately without logging. While the exception message is descriptive, operators monitoring log aggregation (e.g. Splunk, ELK) for `ERROR`-level entries to detect credential rotation issues will no longer see this signal at the provider layer. The error will only appear wherever the exception is ultimately caught and logged (which may be at a different level or with different context).

**Recommendation:** Add `logger.error(...)` back before the `raise`, or confirm that all upstream catchers log at ERROR level.

**3. Startup crash risk — behavioral change not called out**

Previously, if Entra ID credentials were invalid at startup, the provider would silently create a client with `azure_ad_token=None` and fail later at invocation time. Now, the `LLMConfigurationError` is raised during `__init__` (via `default_params` → `resolve_access_token`), which means `load_llm()` will raise during:
- Server startup if this is the default provider (readiness check calls `load_llm`)
- Query handling when a user requests this provider

This is arguably *better* behavior (fail fast), but it's a behavioral change that could cause the **pod to crashloop** if the Entra ID secret has temporarily expired credentials. The old behavior was degraded-but-running; the new behavior is crash-on-load.

**Question:** Is this acceptable for the deployment model? Should this be caught at `load_llm` level and converted to a health check failure rather than crashing the worker?

**4. `TOKEN_CACHE.access_token` return type is still `Optional[str]`**

`resolve_access_token` now declares it returns `str`, but `TOKEN_CACHE.access_token` is typed as `Optional[str]` (line 29). On the happy path this works because `update_token` always sets it to a non-`None` value. But the type system doesn't enforce this — `mypy` may flag this as returning `Optional[str]` from a `str`-typed function.

**Recommendation:** Verify `make check-types` passes, or add an assertion after `TOKEN_CACHE.access_token` return.

**5. No test for the `azure_config is None` path with new exception type**

`test_missing_credentials_check` covers the `azure_config is None` path, but the fixture `provider_config_without_credentials` has no `azure_openai_config` key at all. This exercises the path where credentials are `None` *and* `azure_config` is `None`, which actually goes through `default_params` line 93 where `azure_config` is already `None` from line 59. The test works but it's testing an edge case (no azure_config at all), not the more realistic case where azure_config exists but token retrieval fails.

### Minor nits

- The removed comment `# everything is there, try to retrieve credential` was mildly useful for code scanning; removal is fine but not required by the fix.
- The PR description says "returns HTTP 422" for `LLMConfigurationError`, but 422 (Unprocessable Entity) is only returned by the `validate_provider_and_model` pre-check. When the error happens during actual query processing (in `docs_summarizer._prepare_llm`), it propagates through `except Exception as request_error:` at `ols.py:228` which re-raises, eventually hitting FastAPI's default 500 handler. The PR description should clarify this distinction.

PR #3066, opened 2026-08-24
OLS-3221 Reconnect Postgres cache after DB restart

xrajesh:
## Description

Fixes [OLS-3221](https://redhat.atlassian.net/browse/OLS-3221): app-server pods stay permanently unready after a PostgreSQL restart (node drain, eviction, OOM, rolling upgrade) and require a manual `oc rollout restart`.

### Root cause
`PostgresCache.ready()` detected a dead connection via `connection.closed` / `connection.poll()`. Neither reliably notices a **server-side** PostgreSQL restart — `closed` can stay `0` and `poll()` can return `POLL_OK` on a stale socket — so the pod kept a dead connection and never reconnected. Liveness always returned `alive=true` (never checks the DB), so the kubelet never restarted the pod either. Result: pod Running-but-NotReady forever.

### Fix (minimal, root-cause)
- `ready()` now verifies the connection with a real `SELECT 1` round-trip via the existing `connected()` helper, and reconnects when it fails. A server-side restart is reliably detected and the pod returns to Ready automatically — no restart, no manual intervention.
- Added `connect_timeout` (5s) and `statement_timeout` (5000ms), both configurable via `PostgresConfig`, so a probe against an unavailable database cannot hang the connection indefinitely.

### Why not couple liveness to the DB
Putting a shared dependency in the liveness probe is a known Kubernetes anti-pattern: a single DB outage would fail liveness on **every** app-server pod simultaneously and trigger a fleet-wide restart storm that does nothing to fix the DB. Recovery belongs in readiness + reconnect. This is an intentionally smaller alternative to #2964.

## Type of change
- [x] Bug fix

## Testing
- `make test-unit` — 1186 passed.
- Updated `ready()` unit tests to cover: live connection, dead-connection-then-reconnect ([OLS-3221](https://redhat.atlassian.net/browse/OLS-3221) regression), reconnect-fails, and `None` connection.
- Added tests asserting `connect_timeout` and `statement_timeout` are applied.
- Manual verification suggested: deploy with Postgres cache, delete the Postgres pod, restore it, and confirm app-server pods return to Ready without `oc rollout restart`.

## Summary by CodeRabbit

- **New Features**
  - Added configurable PostgreSQL connection and statement timeouts, with sensible defaults.
  - PostgreSQL sessions now apply the configured statement timeout automatically.

- **Bug Fixes**
  - Improved cache readiness checks by actively verifying connection health.
  - Automatically reconnects when an existing database connection is no longer usable.

- **Documentation**
  - Updated the PostgreSQL configuration reference to include the new timeout settings.

### DIFF
--- a/docs/config.puml
+++ b/docs/config.puml
@@ -119,6 +119,7 @@ class "OpenAIConfig" as ols.app.models.config.OpenAIConfig {
 }
 class "PostgresConfig" as ols.app.models.config.PostgresConfig {
   ca_cert_path : Optional[FilePath]
+  connect_timeout : Annotated
   dbname : str
   gss_encmode : str
   host : str
@@ -127,6 +128,7 @@ class "PostgresConfig" as ols.app.models.config.PostgresConfig {
   password_path : Optional[FilePath]
   port : Annotated
   ssl_mode : str
+  statement_timeout : Annotated
   user : str
   validate_yaml() -> Self
 }
--- a/ols/app/models/config.py
+++ b/ols/app/models/config.py
@@ -807,6 +807,8 @@ class PostgresConfig(BaseModel):
     gss_encmode: str = constants.POSTGRES_CACHE_GSSENCMODE
     ca_cert_path: Optional[FilePath] = None
     max_entries: PositiveInt = constants.POSTGRES_CACHE_MAX_ENTRIES
+    connect_timeout: PositiveInt = constants.POSTGRES_CACHE_CONNECT_TIMEOUT
+    statement_timeout: PositiveInt = constants.POSTGRES_CACHE_STATEMENT_TIMEOUT
     tls_security_profile: Optional["TLSSecurityProfile"] = None
 
     def __init__(self, **data: Any) -> None:
--- a/ols/constants.py
+++ b/ols/constants.py
@@ -138,6 +138,14 @@ POSTGRES_CACHE_DBNAME = "cache"
 POSTGRES_CACHE_USER = "postgres"
 POSTGRES_CACHE_MAX_ENTRIES = 1000
 
+# bound how long a (re)connect attempt may block, in seconds, so a readiness
+# check against an unavailable database cannot hang the probe indefinitely
+POSTGRES_CACHE_CONNECT_TIMEOUT = 5
+
+# server-side cap on how long a single statement may run, in milliseconds, so a
+# query against a wedged connection cannot block forever
+POSTGRES_CACHE_STATEMENT_TIMEOUT = 5000
+
 # look at https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE
 # for all possible options
 POSTGRES_CACHE_SSL_MODE = "require"
--- a/ols/src/cache/postgres_cache.py
+++ b/ols/src/cache/postgres_cache.py
@@ -388,32 +388,25 @@ class PostgresCache(Cache, PostgresBase):
     def ready(self) -> bool:
         """Check if the cache is ready.
 
-        Postgres cache checks if the connection is alive. When a dead
-        connection is detected and the database is available again, the
-        connection is automatically re-established.
+        The connection is verified with a real ``SELECT 1`` round-trip so that
+        a server-side PostgreSQL restart is reliably detected. When the
+        connection is found to be dead and the database is available again, the
+        connection is automatically re-established, restoring readiness without
+        administrator intervention.
 
         Returns:
             True if the cache is ready, False otherwise.
         """
         with self._tx_lock:
-            if not self.connection or self.connection.closed == 1:
-                try:
-                    logger.info("Detected dead connection, attempting reconnect")
-                    self.connect()
-                    return True
-                except Exception as e:
-                    logger.warning("Reconnect attempt failed: %s", e)
-                    return False
+            if self.connected():
+                return True
             try:
-                return self.connection.poll() == psycopg2.extensions.POLL_OK
-            except (psycopg2.OperationalError, psycopg2.InterfaceError):
-                try:
-                    logger.info("Connection error during poll, attempting reconnect")
-                    self.connect()
-                    return True
-                except Exception as e:
-                    logger.warning("Reconnect attempt failed: %s", e)
-                    return False
+                logger.info("Detected dead connection, attempting reconnect")
+                self.connect()
+                return True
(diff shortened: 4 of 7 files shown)


sriroopar:
## Adversarial Review

### 1. `SET statement_timeout` is session-level but applied before schema initialization — DDL can be killed

**File:** `ols/utils/postgres.py`

The `SET statement_timeout = 5000` (5 seconds) is applied **before** the schema initialization DDL runs (advisory lock acquisition, table creation). On a cold start or schema migration, the DDL may take longer than 5 seconds — especially if there's lock contention. This means the `SET LOCAL lock_timeout = '60s'` on the next line only protects the lock wait, but the 5-second statement timeout still applies to the DDL itself.

`SET statement_timeout` without `LOCAL` applies to the **entire session**, so every subsequent `cursor.execute` — including the `CREATE TABLE IF NOT EXISTS` and `CREATE INDEX` — is capped at 5 seconds. Under load this is a real concern.

**Recommendation:** Use `SET LOCAL statement_timeout` during schema init to override, or apply the session-level timeout *after* schema initialization completes.

---

### 2. `ready()` now takes the `_tx_lock` and calls `connect()` — potential deadlock with concurrent cache operations

**File:** `ols/src/cache/postgres_cache.py`, line ~398

The `ready()` method holds `self._tx_lock` while calling `self.connect()`. If `connect()` blocks (e.g., TCP handshake hanging for up to `connect_timeout` seconds), then **all other cache operations that need `_tx_lock`** are blocked for the entire reconnect duration. Since readiness probes run on a timer, a slow/failing Postgres could cause:
- Cache reads/writes to queue up behind the lock
- Request latency spikes
- Cascading timeouts in the application

**Recommendation:** Either release the lock before reconnecting and re-acquire after, or use a separate lock/flag for the reconnect path.

---

### 3. `connected()` swallows all exceptions — the new `ready()` can't distinguish transient vs permanent failures

The `connected()` helper (not modified in this PR) catches broad exceptions during `SELECT 1`. The new `ready()` calls `connected()` and if it returns `False`, immediately tries to reconnect. There's no distinction between:
- A transient network blip (worth retrying)
- An authentication failure (will never succeed)
- A configuration error (will never succeed)

With the broad `except Exception` in the reconnect path, permanent failures are silently logged as warnings and the pod stays NotReady — which is correct behavior — but there's no escalation or metrics emission to help operators distinguish "DB is down temporarily" from "credentials rotated and we'll never reconnect."

---

### 4. The `except Exception` catch in reconnect is too broad

```python
except Exception as e:
    logger.warning("Reconnect attempt failed: %s", e)
    return False
```

This catches `KeyboardInterrupt`, `SystemExit`, and programming errors like `TypeError` or `AttributeError`. If there's a bug in `connect()` itself (e.g., a schema migration error, a missing config field), it will be swallowed as a warning and the pod will report not-ready — making the real error invisible.

**Recommendation:** Catch `(psycopg2.Error, OSError)` to cover database and network errors. Let programming errors propagate.

---

### 5. `SET statement_timeout` uses a plain `%s` parameter — potential type mismatch

```python
cursor.execute("SET statement_timeout = %s", (str(config.statement_timeout),))
```

While this works because psycopg2 will quote the string, PostgreSQL expects `statement_timeout` to be set as an integer (milliseconds) or a string with units (e.g., `'5s'`). Passing a string `'5000'` without units works but is fragile — if someone changes the config to include units (e.g., `"5s"`), psycopg2 would quote it as `'5s'` which also works, but the Pydantic model declares it as `PositiveInt` so a unit suffix would fail validation. This is fine today but the coupling between the type annotation and the SQL format is implicit.

---

### 6. Race condition: `connected()` can succeed, but connection dies before the method returns `True`

```python
with self._tx_lock:
    if self.connected():
        return True
```

The `connected()` call issues `SELECT 1` and returns `True`. Between that return and the caller using the connection for an actual operation, the connection could die again. This is an inherent TOCTOU issue. While not fixable in the readiness check itself, it means the readiness probe can report `True` but the very next cache operation fails. The old code had the same issue, so this isn't a regression — but worth noting.

---

### 7. Tests mock `connected()` rather than simulating the actual failure path

```python
with patch.object(cache, "connected", return_value=False):
    assert cache.ready()
```

By mocking `connected()`, the tests don't verify that a **real** stale connection (server-side restart) is actually detected by the `SELECT 1` in `connected()`. The old tests at least simulated `connection.closed = 1` and `poll()` raising errors. The new tests only verify the *orchestration* of `ready()` (if `connected()` is False, call `connect()`), not the *detection* mechanism itself.

**Recommendation:** Add an integration-style test (or a more detailed unit test) that verifies `connected()` returns `False` when the cursor's `execute("SELECT 1")` raises `OperationalError`.

---

### 8. Deleted test coverage: `test_ready_reconnects_on_poll_error` and `test_ready_returns_false_when_poll_reconnect_fails` removed

The PR removes 4 test scenarios (~32 lines of tests) and adds 2 simpler replacements. The removed tests covered:
- Poll raising `OperationalError` → reconnect succeeds
- Poll raising `InterfaceError` → reconnect fails

These are replaced with a single "connected returns False" mock. While the new implementation doesn't use `poll()`, the test reduction means there's less coverage of the actual failure modes. If someone introduces a regression in `connected()`, the test suite won't catch it via the `ready()` tests.

---

### 9. `connect_timeout` doesn't protect against TCP black-hole scenarios

`connect_timeout` in libpq only covers the TCP connection phase. If the TCP handshake succeeds but the server never responds to the startup packet (e.g., a firewall silently dropping packets after SYN/ACK), the connection can hang indefinitely. The `statement_timeout` won't help because no statement has been issued yet — the connection is in the authentication/startup phase.

**Recommendation:** Consider adding `keepalives_idle` + `keepalives_interval` + `keepalives_count` configuration to detect dead connections at the TCP level.

---

### Summary

| Severity | Issue |
|----------|-------|
| **High** | Statement timeout applied before DDL — schema init can be killed |
| **High** | Lock held during reconnect — latency spike / starvation risk |
| **Medium** | Broad `except Exception` swallows programming errors |
| **Medium** | Tests mock the detection mechanism rather than testing it |
| **Low** | TOCTOU between `connected()` and actual use |
| **Low** | No TCP keepalive configuration |

The core idea is sound — replacing `connection.closed` / `poll()` with a real `SELECT 1` round-trip is the right approach. But the implementation has execution-order and concurrency concerns that could cause production issues under exactly the conditions this PR is trying to fix (database restarts under load).

PR #3071, opened 2026-08-26
OLS-4019: drop unsupported params for LiteLLM

dlom:
xref: [OLS-4019](https://redhat.atlassian.net/browse/OLS-4019)

The periodic for this repo failed this morning (see card/slack link in card):

The evaluation pipeline uses LiteLLM to call openai/gpt-5-mini. LiteLLM is sending logprobs and top_logprobs parameters that gpt-5-mini does not support. This error occurred repeatedly across multiple conversations during the evaluation scoring phase, but only 1 evaluation was counted as an ERROR (the others apparently recovered or were scored through a different code path).

```
litellm.UnsupportedParamsError: openai does not support parameters:
['logprobs', 'top_logprobs'], for model=gpt-5-mini
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved evaluation compatibility by automatically ignoring unsupported parameters when communicating with judge models.

### DIFF
--- a/eval/system.yaml
+++ b/eval/system.yaml
@@ -121,6 +121,7 @@ environment:
   DEEPEVAL_DISABLE_PROGRESS_BAR: "YES"     # Disable DeepEval progress bars
 
   LITELLM_LOG: ERROR                       # Suppress LiteLLM verbose logging
+  LITELLM_DROP_PARAMS: "True"              # Drop params unsupported by the judge model
 
 # Logging Configuration
 logging:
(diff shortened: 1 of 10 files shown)


sriroopar:
@dlom Thank you very much for this contribution, would you mind extending the fix across all the system_eval files within the directory. I am not sure which specific eval is of context here but this change across all the eval files would be helpful. Thanks! :)

sriroopar:
@dlom Recent push, removes the fix from system.eval, which requires the fix. All the 10 files would use the fix, if you could update the PR, that would be great! Thanks again!

sriroopar (approved):
Thank you for the fix!

PR #3077, opened 2026-08-31
OLS-3599: Relax OKP tool mandate when BYOK content is configured

blublinsky:
## Summary

When BYOK (Bring Your Own Knowledge) indexes are configured alongside OKP (Solr),
the system prompt forced the LLM to "ALWAYS call search_openshift_documentation"
before answering. This caused OKP results to overshadow BYOK context that already
answered the question, making BYOK effectively useless.

- Add `SOLR_DOCS_TOOL_SUPPLEMENT_WITH_BYOK` prompt variant that tells the LLM to
  use provided context first and only call the OKP tool if context is insufficient
- Wire `byok_active` flag through `DocsSummarizer` → `GeneratePrompt` based on
  whether `reference_content.indexes` is configured
- Non-BYOK deployments are unaffected (original mandatory supplement is unchanged)

## Type of change

- [ ] Refactor
- [ ] New feature
- [ x] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
https://redhat.atlassian.net/browse/OLS-3599
- Closes #
https://redhat.atlassian.net/browse/OLS-3599

## Testing

## Summary by CodeRabbit

* **New Features**
  * Improved document-based answers when reference content is available.
  * Relevant documents already in context can now be used directly, with searches performed only when additional information is needed.
  * Existing citation and concise-response guidance is preserved.

* **Bug Fixes**
  * Document search guidance now adapts automatically based on whether reference content is configured.

* **Tests**
  * Added coverage for reference-content detection and adaptive search guidance.

### DIFF
--- a/ols/src/prompts/prompts.py
+++ b/ols/src/prompts/prompts.py
@@ -97,6 +97,18 @@ Grounded answers (passages from ``search_openshift_documentation``):
 * If the answer is not based on such passages, stay concise per the general style rules above.
 """
 
+SOLR_DOCS_TOOL_SUPPLEMENT_WITH_BYOK = """
+Solr docs tool:
+* ``search_openshift_documentation`` searches published Red Hat product documentation (not live cluster resources).
+* The provided context contains domain-specific knowledge. If it answers the question, use it directly. Only call ``search_openshift_documentation`` to fill in general OpenShift details that the domain knowledge does not cover. Never contradict or override relevant domain knowledge.
+* When you use a passage from the tool, cite its ``title`` and ``docs_url`` from the tool JSON. Never invent documentation URLs.
+
+If the tool is used, grounded answers (passages from ``search_openshift_documentation``):
+* Bullets or numbered steps are allowed so passage detail is not dropped for brevity.
+* Keep concrete details from passages (manifests, ``oc`` commands, console paths).
+* If the answer is not based on such passages, stay concise per the general style rules above.
+"""
+
 TROUBLESHOOTING_SYSTEM_INSTRUCTION = """# ROLE
 You are "OpenShift Lightspeed", an AI assistant specializing in OpenShift troubleshooting and diagnostics.
 
--- a/ols/src/query_helpers/docs_summarizer.py
+++ b/ols/src/query_helpers/docs_summarizer.py
@@ -98,6 +98,8 @@ class DocsSummarizer(QueryHelper):
             self._solr_hybrid is not None and self._solr_client is not None
         )
         self._solr_docs_tool_prompt_guidance = solr_docs_tool_active
+        ref = config.config.ols_config.reference_content
+        self._byok_active = ref is not None and bool(ref.indexes)
         self._tool_calling_enabled = bool(self.mcp_servers) or solr_docs_tool_active
         if self.mcp_servers:
             logger.info("MCP servers provided: %s", list(self.mcp_servers.keys()))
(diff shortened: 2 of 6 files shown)


ols/src/prompts/prompts.py
```diff
@@ -97,6 +97,18 @@
 * If the answer is not based on such passages, stay concise per the general style rules above.
 """
```
sriroopar, line 99: Spec mentions that changes are required on the BYOK context instruction — replace USE_CONTEXT_INSTRUCTION ("Use the retrieved document") with a BYOK-aware variant that labels content as domain-specific

Also another missing spec change. Chunk labeling — format BYOK chunks with "Domain Knowledge:" prefix instead of "Document:"

ols/src/query_helpers/docs_summarizer.py
```diff
@@ -98,6 +98,8 @@ def __init__(
             self._solr_hybrid is not None and self._solr_client is not None
         )
         self._solr_docs_tool_prompt_guidance = solr_docs_tool_active
+        ref = config.config.ols_config.reference_content
```
sriroopar, line 101: the non empty check should rather be towards the final _rag_context_  after the BYOK retrieval — that's already available in _build_final_prompt via the rag_chunks parameter. To avoid cases where context is missing in the final result.

ols/src/prompts/prompts.py
```diff
@@ -97,6 +97,18 @@
...
 
+SOLR_DOCS_TOOL_SUPPLEMENT_WITH_BYOK = """
+Solr docs tool:
+* ``search_openshift_documentation`` searches published Red Hat product documentation (not live cluster resources).
+* The provided context may already contain relevant documents. If they answer the question, use them directly. Only call ``search_openshift_documentation`` if the context is insufficient.
```
sriroopar: stronger prompt clausing suggested by spec 

`The LLM is instructed to use search_openshift_documentation only to fill in general OpenShift details that the domain knowledge does not cover, and never to contradict or override relevant domain knowledge.`

sriroopar (changes requested):
Left a few comments based on the spec @blublinsky ! Thanks!

blublinsky:
**Re: BYOK context instruction + chunk labeling**

Addressed by rewording the BYOK supplement to label context as "domain-specific knowledge" and adding the "never contradict or override" clause from the spec. This keeps the change minimal — no separate `USE_CONTEXT_INSTRUCTION` variant or chunk prefix needed. The LLM gets the semantic signal through the Solr tool guidance itself.

Regarding chunk labeling ("Domain Knowledge:" prefix) — this would require plumbing BYOK awareness into the chunk formatter, which is a deeper change. The current rewording already tells the LLM to treat context as domain knowledge. Happy to add chunk labeling as a follow-up if testing shows the LLM still doesn't prioritize BYOK content sufficiently.

blublinsky:
**Re: checking rag_context instead of config**

The prompt already handles this case: "If it answers the question, use it directly. Only call `search_openshift_documentation` if the context is insufficient." When BYOK retrieval returns nothing, the context is empty, so the LLM naturally falls through to calling the tool — no additional code changes needed in token processing.

blublinsky:
**Re: stronger prompt wording**

Already incorporated — the updated BYOK supplement now reads: "Only call `search_openshift_documentation` to fill in general OpenShift details that the domain knowledge does not cover. Never contradict or override relevant domain knowledge."

blublinsky:
**Overall design note**

Splitting prompts into two variants and picking the appropriate one based on configuration makes the code simpler and more maintainable than threading BYOK awareness through the context instruction, chunk formatter, and token processing layers.

sriroopar (approved):
Thank you for the clarifying comments @blublinsky .

PR #3081, opened 2026-09-03
OLS-2190: pass username and version for on-prem watsonx

DeokarT:
[OLS-2190](https://redhat.atlassian.net/browse/OLS-2190). Also OLS-2849, because that is just the next error after username.

IBM Cloud watsonx is fine. Leave it alone. Secret with `apitoken` only, URL on `*.ml.cloud.ibm.com`. That path does not change.

On-prem Cloud Pak for Data is not IBM Cloud. The provider still built `ChatWatsonx` with apikey only, so you get `Did not find username` / `WATSONX_USERNAME`. Stick username in as an env var and you hit missing `version` (`5.0`–`5.3`). Scale the operator to 0 and edit the Deployment and it gets overwritten. That is not a fix.

What this does:

- IBM Cloud (or empty URL, we default to us-south): still apikey only.
- Any other URL: read `username` and `version` from the same credentials directory we already mount. Pass them into `ChatWatsonx`. `instance_id` is optional and defaults to `openshift`.
- If those files are missing, raise a real error. Not the Pydantic one.

The secret already lands as a directory under `/etc/apikeys/<secret>/`. Extra keys just show up as files. No new volume mounts.

Operator PR: https://github.com/openshift/lightspeed-operator/pull/2013

### DIFF
--- a/.ai/spec/how/llm-providers.md
+++ b/.ai/spec/how/llm-providers.md
@@ -164,6 +164,8 @@ WatsonX uses IBM's `GenTextParamsMetaNames` constants for parameter keys instead
 
 WatsonX also passes `params=self.params` to `ChatWatsonx` (as a nested dict), unlike OpenAI-compatible providers that spread params as `**self.params`.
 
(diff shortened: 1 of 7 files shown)


onmete:
WantsonX on-prem is deferred - the team has no capacity to proceed here.
/close

PR #3083, opened 2026-09-03
OLS-3135 block non-streaming tool approval bypass

xrajesh:
## Summary

Fix for R-002: Non-Streaming Query Path Completely Bypasses Tool Approval Safety Gate.
https://redhat.atlassian.net/browse/OLS-3135 . Proposing a fix to avoid /v1/query making changes 
to cluster with no approval. 

- remove the streaming gate from approval validation policy
- block approval-required tool calls on `/v1/query` with guidance to use `/v1/streaming_query`
- add unit coverage for approval policy and non-streaming fail-closed behavior

## Testing
- `uv run pytest tests/unit/tools/test_approval.py -q`
- `uv run python - <<'PY'\nfrom ols import config\nconfig.ols_config.authentication_config.module = "k8s"\nimport pytest\nraise SystemExit(pytest.main(["tests/unit/tools/test_tools.py", "-q"]))\nPY`\n- `uv run ruff check ols/src/tools/approval.py ols/src/tools/tools.py tests/unit/tools/test_approval.py tests/unit/tools/test_tools.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Approval-required tools in non-streaming requests now fail safely without executing or waiting for approval.
  * Error responses identify the affected tool call and direct callers to the streaming query endpoint for approval workflows.
  * Approval settings are now consistently applied to non-streaming requests, including requests using an “always approve” policy.
  * Approval behavior for streaming requests remains unchanged.


blublinsky:
I think there is a bit cleaner way

In approvals.py:

def is_approval_enabled(
    approval_type: ApprovalType | str,
) -> bool:
    # Normalize enum/string config value before strategy checks.
    approval_value = _approval_type_value(approval_type)
    # Approval flow is active only for explicit approval strategies.
    return approval_value in {
        ApprovalType.ALWAYS.value,
        ApprovalType.TOOL_ANNOTATIONS.value,
    }

....
def need_validation(
    streaming: bool,
    approval_type: ApprovalType | str,
    tool_annotation: dict[str, object] | None = None,
) -> bool:
    """Return true when a tool call must go through approval validation."""
    # Fast exit when approval flow is disabled for this request.
    if not is_approval_enabled(approval_type=approval_type):
        return False
    // for streaming base validation requirements on annotations only    
    if  !streaming:
        annotation_payload = normalize_tool_annotation(tool_annotation)
         if not annotation_payload:
              return True
           value = annotation_payload.get("readOnlyHint")
           return not (isinstance(value, bool) and value) 
    # Normalize enum/string config value before per-strategy decision.
    approval_value = _approval_type_value(approval_type)
    match approval_value:
        case ApprovalType.TOOL_ANNOTATIONS.value:
            # Annotation strategy: require approval by default unless the tool
            # explicitly declares readOnlyHint=true.
            annotation_payload = normalize_tool_annotation(tool_annotation)
            if not annotation_payload:
                return True
            value = annotation_payload.get("readOnlyHint")
            return not (isinstance(value, bool) and value)
        case _:
            # ALWAYS strategy (and any unknown fallback) requires approval.
            return True

In tools.py

    need_approval = need_validation(
        streaming=streaming,
        approval_type=config.tools_approval.approval_type,
        tool_annotation=tool_annotation,
    )
    if not need_approval:
        return
    if  streaming:
        approval_id = str(uuid4())
        if audit_ctx is None:
           logger.warning(
             "Tool approval requested without audit context; "
              "approval will be unresolvable for tool=%s",
               tool_name,
           )
        user_id = audit_ctx.user_id if audit_ctx else ""
        register_pending_approval(approval_id=approval_id, user_id=user_id)

         if audit_ctx:
            audit_ctx.logger.tool_approval_requested(
               tool_name=tool_name,
                approval_id=approval_id,
           )

       yield _approval_required_event(
           approval_id=approval_id,
            tool_name=tool_name,
            tool_description=tool.description,
             tool_args=tool_args,
             tool_annotation=tool_annotation,
        )
        outcome = await get_approval_decision(
            approval_id=approval_id,
             timeout_seconds=config.tools_approval.approval_timeout,
        )
     else:  
       // Always reject for query
        outcome = "rejected"

xrajesh:
@blublinsky
   With the proposed approach, I think there is inconsistency: if the admin configures `ApprovalType.ALWAYS`, but the MCP tool advertises `readOnlyHint: true`, the non-streaming (/query) would allow the tool to execute
 without approval. That is inconsistent with the behaviour of `ALWAYS`, where every tool call should require human approval.

blublinsky:
> @blublinsky With the proposed approach, I think there is inconsistency: if the admin configures `ApprovalType.ALWAYS`, but the MCP tool advertises `readOnlyHint: true`, the non-streaming (/query) would allow the tool to execute without approval. That is inconsistent with the behaviour of `ALWAYS`, where every tool call should require human approval.

There is inconsistency no matter what. If the set is always, then we will have to reject everything - not good.
It can be ok to support never, but always will probably kill it.

This said, If you want to do this, its even less change. remove
if !streaming:
annotation_payload = normalize_tool_annotation(tool_annotation)
if not annotation_payload:
return True

and thats it

Your call

Caveat. OKP is a tool and requires the same confirmation. So "Always" will break OKP

blublinsky:
You are still over complicating it

ols/src/tools/tools.py
```diff
@@ -413,6 +414,24 @@ async def _evaluate_and_emit_approval_event(
...
+            audit_ctx.logger.tool_approval_decision(
+                approval_id="non-streaming",
+                decision=ApprovalOutcome.REJECTED,
+                tool_name=tool_name,
+            )
+        yield _approval_rejection_event(
+            tool_name=tool_name,
+            tool_call_id=tool_id,
+            outcome=ApprovalOutcome.REJECTED,
+        )
```
blublinsky: This is way too complex. KISS, please

PR #3084, opened 2026-09-04
OLS-3454 OLS-3455 OLS-3456 OLS-3457 OLS-3458 OLS-3459: Standardize reasoning configuration, migrate providers to it

blublinsky:
## Description

Introduced a new freeform `reasoning_config` model parameter and
migrated to it all of the relevant providers.

## Changes

- Replace legacy reasoning model-name detection with explicit `reasoning_config`.
- Remove deprecated `reasoning_effort`, `reasoning_summary`, and top-level
  `verbosity` model fields.
- Pass reasoning settings through LangChain for all current providers
- Skip deployment CA bundle construction in explicit local mode.
- Add a local OpenAI reasoning configuration for manual testing.
- Update specifications and testing documentation.
- Update provider and configuration unit tests.
- Synchronize the service version and OpenAPI schema to `1.1.4`.

## Testing

- OpenAI provider tests: passed
- Azure OpenAI provider tests: passed
- Integration tests: passed
- Ruff checks: passed
- Manual local OpenAI test with `gpt-5.2`: passed
- Streaming reasoning validation produced locally

## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] CVE fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Configuration Update
- [ ] CI configuration change

- Related Issue #
- Closes #

## Testing

## Behavior Changes

* Reasoning settings now use a flexible `reasoning_config` dictionary across supported providers.
* Sampling defaults are no longer inferred from model names and are omitted when reasoning settings are not configured.
* Reasoning parameters are supported for additional hosted and cloud providers.
* Localhost runs now skip unnecessary deployment certificate setup.
* Bedrock requests support reasoning configuration and updated token handling.
* Credential hot reload is now supported through configuration.

## Documentation

* Updated provider, configuration, and testing documentation for these behaviors.


ols/src/llms/providers/bedrock.py
```diff
@@ -126,6 +156,12 @@ def load(self) -> BaseChatModel:
 
             params["model_id"] = model_id
             params["region_name"] = region
+
+            max_tokens = params.pop("max_tokens", 4096)
+            params["max_completion_tokens"] = max_tokens
```
vimalk78: The Anthropic Bedrock path removes `max_tokens` and replaces it with `max_completion_tokens`. `ChatBedrockConverse` expects the Bedrock-compatible `max_tokens` parameter; `max_completion_tokens` is for the OpenAI-compatible `ChatOpenAI` path. Please keep `max_tokens` here so Anthropic Bedrock initialization and the configured output-token limit continue to work.

ols/src/llms/providers/provider.py
```diff
@@ -153,6 +161,8 @@ class ProviderParameter:
     ProviderParameter("temperature", float),
     ProviderParameter("top_p", float),
     ProviderParameter("top_k", int),
+    ProviderParameter("thinking_level", str),
+    ProviderParameter("thinking_budget", int),
```
vimalk78, line 169: `include_thoughts` is listed as a supported Gemini reasoning key in the specification, but it is absent from `GoogleVertexParameters`. `_validate_parameters()` will therefore silently discard it before `ChatGoogleGenerativeAI` is constructed. Please add the boolean parameter to the allowlist and forward it from `reasoning_config` in `google_vertex.py`, with a test covering retention.

vimalk78 (changes requested):
## Review summary

I found two must-fix provider integration issues and one documentation issue.

- Anthropic Bedrock passes `max_completion_tokens` to `ChatBedrockConverse`, which expects `max_tokens`.
- Gemini `include_thoughts` is documented but is silently discarded by provider validation and is not forwarded to `ChatGoogleGenerativeAI`.
- `examples/olsconfig.yaml` still documents the removed reasoning fields and should be updated to `reasoning_config`.

The detailed review is saved locally in `.ols/reviews/pr-3084.md`. `git diff --check` also reports trailing whitespace in `.ai/spec/how/llm-providers.md`.

ols/src/llms/providers/bedrock.py
blublinsky: fixed

ols/src/llms/providers/provider.py
blublinsky: Its not about what Geminin can do. Its about what ChatGoogleGenerativeAI class supports

ols/src/llms/providers/google_vertex.py
```diff
@@ -50,8 +51,19 @@ def default_params(self) -> dict[str, Any]:
...
         if self.provider_config.url is not None:
-            params["base_url"] = str(self.provider_config.url)
-        return params
+            default_params["base_url"] = str(self.provider_config.url)
+
+        model_config = self.provider_config.models.get(self.model)
+        model_params = getattr(model_config, "parameters", None) or ModelParameters()
+        reasoning_config = model_params.reasoning_config or {}
+        if reasoning_config:
+            default_params["include_thoughts"] = True
```
vimalk78: `include_thoughts` is always set to `True` whenever `reasoning_config` is non-empty. This ignores the configured value and makes `include_thoughts: false` impossible. Please preserve `reasoning_config["include_thoughts"]` when that key is present, and add coverage for both `true` and `false`. The concern is not whether `ChatGoogleGenerativeAI` supports the parameter; it is whether the service honors the user's configured value.

ols/src/llms/providers/google_vertex.py
```diff
@@ -89,6 +101,20 @@ def default_params(self) -> dict[str, Any]:
             "credentials": self.credentials,
         }
 
+        model_config = self.provider_config.models.get(self.model)
+        model_params = getattr(model_config, "parameters", None) or ModelParameters()
+        reasoning_config = model_params.reasoning_config or {}
+        if reasoning_config:
```
vimalk78: This implementation only handles `thinking_enabled` and `budget_tokens`, so it drops documented Anthropic thinking keys such as `type` and `display`. The ticket example `{type: adaptive, display: summarized}` therefore cannot work. Please forward the configured thinking dictionary to `ChatAnthropicVertex` instead of reconstructing it from only these two keys.

ols/src/llms/providers/rhelai_vllm.py
```diff
@@ -42,12 +44,17 @@ def default_params(self) -> dict[str, Any]:
...
             "cache": None,
             "temperature": 0.01,
-            "max_completion_tokens": 512,
+            "max_completion_tokens": 4096,
             "verbose": False,
             "http_client": self._construct_httpx_client(False),
             "http_async_client": self._construct_httpx_client(True),
         }
 
+        model_config = self.provider_config.models.get(self.model)
```
vimalk78, line 54: OLS-3459 requires a `ChatVLLMReasoning` implementation that preserves `reasoning_content`/`reasoning` in streaming and non-streaming responses and emits reasoning chunks from the execution agent. This change only adds the generic OpenAI `reasoning` payload to `ChatOpenAI`; it does not implement the required vLLM response handling or switch RHOAI/RHELAI to the new subclass.

vimalk78 (changes requested):
## Re-review summary

The previous Bedrock token-name issue and Gemini parameter allowlist issue were addressed. I found three remaining provider/requirement issues.

- Gemini `include_thoughts` is hardcoded to `true` instead of honoring the configured value.
- Vertex Anthropic drops documented thinking keys such as `type` and `display`.
- OLS-3459's vLLM reasoning integration is still not implemented.

Additionally, `pyproject.toml` still allows `langchain-openai>=0.2.9`, although the new `reasoning` and `verbosity` fields require a newer minimum (at least `0.3.28`). `git diff --check` still reports trailing whitespace in `.ai/spec/how/llm-providers.md`.

ols/src/llms/providers/google_vertex.py
blublinsky: fixed

ols/src/llms/providers/google_vertex.py
blublinsky: This is extremely bad. Different models for different providers support different options. The most generic subset, that works for both bedrock and vertex for majority of models is:
````
reasoningConfig:
  effort: "high"           # Generic: low | medium | high | xhigh | max
  budgetTokens: 10000      # Generic: int, for fine-grained control
  enabled: true            # Generic: bool (best-effort)
````
  This is what I am doing here. Anything else is going to be too complex

File: ols/src/llms/providers/rhelai_vllm.py (hunk 5 above)
Comment (blublinsky) [reply ols/src/llms/providers/rhelai_vllm.py:54]: done

File: ols/src/llms/providers/utils.py [hunk 6]
```diff
@@ -51,3 +54,16 @@ def load_vertex_credentials(credentials_json: str) -> GoogleCredentials:
...
+    params: ModelParameters, default_parameters: dict[str, Any]
+) -> None:
+    """Populate reasoning configuration from model parameters."""
+    reasoning_config = params.reasoning_config or {}
+
+    if reasoning_config:
+        reasoning_parameters = dict(reasoning_config)
+        if "verbosity" in reasoning_parameters:
+            default_parameters["verbosity"] = reasoning_parameters.pop("verbosity")
+        default_parameters["reasoning"] = reasoning_parameters
```
vimalk78, line 69: This helper is also used by RHOAI/RHELAI vLLM. Passing the configuration as `reasoning` causes the current `ChatOpenAI` implementation to select the Responses API; `reasoning` is a Responses-only field. vLLM exposes the OpenAI-compatible Chat Completions API. Please keep the vLLM path on Chat Completions (for example, map the supported effort setting to `reasoning_effort` and/or explicitly force `use_responses_api=False`) while retaining the custom response-field extraction.

ols/src/llms/providers/google_vertex.py
```diff
@@ -89,6 +104,24 @@ def default_params(self) -> dict[str, Any]:
...
+            thinking_config = {}
+            if reasoning_config.get("thinking_enabled"):
+                if reasoning_config.get("thinking_effort") in ["high", "xhigh", "max"]:
+                    thinking_config["type"] = "adaptive"
+                    thinking_config["display"] = "summarized"
+                else:
+                    thinking_config["type"] = "enabled"
+                    budget = reasoning_config.get("budget_tokens", 0)
+                    if budget > 0:
```
vimalk78: When `reasoning_config` is non-empty but `thinking_enabled` is false or absent, this still adds `model_kwargs={"thinking": {}}`. The provider specification says disabled or absent thinking must send no thinking configuration. Only add `model_kwargs["thinking"]` when thinking is enabled.

ols/src/llms/providers/bedrock.py
```diff
@@ -105,13 +107,49 @@ def _build_sigv4_auth(self, region: str) -> AwsSigV4Auth:
...
+
+        else:
+            # Fallback to additional_model_request_fields for older models
+            if "type" in reasoning_config:
+                params["additional_model_request_fields"]["thinking"]["type"] = (
+                    reasoning_config["type"]
+                )
+            if "budget_tokens" in reasoning_config:
+                tokens = reasoning_config["budget_tokens"]
+                # for Anthropic Bedrock min tokens is 1,024, max below model's max_tokens
```
vimalk78, line 147: This clamp can return a budget greater than or equal to `max_tokens`: for example, `max_tokens=512` produces `budget_tokens=1024`. Bedrock requires the thinking budget to be at least 1024 and below the output-token limit. Please reject incompatible values or ensure the output limit is increased before sending the request.

vimalk78 (changes requested):
## Re-review

The Gemini `include_thoughts` issue is resolved, and the vLLM subclass has been added. I found three remaining functional issues.

1. The vLLM providers still receive the Responses-only `reasoning` field, which routes `ChatOpenAI` to the Responses API instead of vLLM Chat Completions.
2. Vertex Anthropic sends an empty thinking configuration when thinking is disabled or absent.
3. Bedrock budget clamping can produce `budget_tokens >= max_tokens`.

The declared `langchain-openai>=0.2.9` lower bound also does not support the new `reasoning` and `verbosity` fields; it should be raised to at least `0.3.28`. `git diff --check` still reports trailing whitespace in `.ai/spec/how/llm-providers.md:185`.

ols/src/llms/providers/utils.py
blublinsky: done

ols/src/llms/providers/google_vertex.py
blublinsky: fixed

ols/src/llms/providers/bedrock.py
blublinsky: fixed

vimalk78 (approved):

PR #3088, opened 2026-09-10
OLS-3261 Retry e2e app readiness timeouts

sriroopar:
## Summary
- classify an app-server container readiness timeout as `RuntimeError`
- allow the existing one-time `adapt_ols_config` retry to recover from a transient provider-reconfiguration rollout timeout
- cover the timeout type with a unit test

## Testing
- `uv run pytest tests/unit/e2e/test_cluster.py -q`
- `uv run ruff check tests/e2e/utils/cluster.py tests/unit/e2e/test_cluster.py`
- `uv run black --check tests/e2e/utils/cluster.py tests/unit/e2e/test_cluster.py`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error reporting when server containers fail to become ready by raising a more specific runtime error.

- **Tests**
  - Added coverage verifying that a timeout while waiting for running containers produces the expected error.

### DIFF
--- a/tests/e2e/utils/cluster.py
+++ b/tests/e2e/utils/cluster.py
@@ -475,7 +475,7 @@ def wait_for_running_pod(
         "Waiting for containers in the server pod to become ready",
     )
     if not r:
-        raise Exception("Timed out waiting for containers to become ready")
+        raise RuntimeError("Timed out waiting for containers to become ready")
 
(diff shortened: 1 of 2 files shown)


blublinsky:
I am not convinced. This will make the error more explicit, but will not fix the root cause

